PR #85, opened 2017-08-16
Use default name

ajtritt:
Adding the following change to schema:
https://github.com/NeurodataWithoutBorders/nwb-schema/issues/16
https://github.com/NeurodataWithoutBorders/nwb-schema/pull/42

And changes to form.build.map and form.spec to use default_name


oruebel:
@ajtritt It looks like there is an error in the spec for BehavioralEpochs. The key ```default_name``` is missing a ```t```, i.e.,  ```defaul_name: BehavioralEpochs``` should be ```default_name: BehavioralEpochs```

PR #86, opened 2017-08-16
Remove unused imports in setup.py

arokem:
Certainly not as important as running the tests, but you can consider adding a run of [flake8](https://pypi.python.org/pypi/flake8) to the CI. Something like this: https://github.com/uwescience/shablona/blob/master/.travis.yml#L40


oruebel (approved):
Looks good to me.

Issue #117, opened 2017-09-16
saving with "to_X()" methods on NWBFile


neuromusic:
idea for discussion:

I really like the way that pandas handles saving & converting data to various formats. Basically, any target format, whether that's saving to disk or converting to another python data format, uses the same syntax: methods on the object of the format "to_X" where X is the target format.

```
nwb_file.to_hdf('my_data.nwb')
```

this simplifies file saving for the user... they don't need to understand anything about `form` at all. further, they can readily discover supported target formats with simple tab completion.

obviously, behind the scenes this would be a simple wrapper for form's HDF5IO

similarly, one could imagine e.g. converting to an xarray with syntax like

```
xr_dataset = nwb_file.to_xarray()
```

oruebel:
I like the idea, but I would probably make it part of a separate I/O package in PyNWB rather than adding I/O-specific functions directly to NWBFile, but that's just implementation details.

neuromusic:
To clarify, the proposal here is quite specifically to add these as method calls directly on the NWBFile instance.

ajtritt:
Adding these method calls would require the NWBFile class to have a dependency on a backend implementation and the object-relational mapping layer. I'd prefer to keep these classes agnostic to these parts of form.

I added another scenario to the Gist demonstrating different I/O scenarios. This demonstrates what @oruebel suggested.
https://gist.github.com/ajtritt/11f2bda4846a96eb67cd9e9a7e1b2bdc#file-io_scenarios-py-L76

Users could still discover by tab-completion, but it would have to be on package. Would that be okay?

neuromusic:
>  I'd prefer to keep these classes agnostic to these parts of form.

Um... why?
 
1. pynwb already has a dependency on form
1. form has a handful of backends that it explicitly supports

I don't see any technical problem with adding backend-aware methods on the class, whereas the usability benefits of saving from a method directly on the data object are substantial.

ajtritt:
Um, because, apart from some basic utilities in form, NWBFile and other Container classes actually have almost no dependency on form. These are the classes that users interface with

ajtritt:
[...]

PR #121, opened 2017-09-21
adds save "methods" w/ to_X syntax #117

neuromusic:
no unit tests or documentation, but I manually tested that the implementation works in 2.7 & 3.6

this makes the assumption that new backends will conform to the module structure of the HDF5IO backend. specifically: a module with a single class

this specifically addresses @ajtritt's concern in #117 that "For every backend, we have to edit NWBFile to add some new to_<backend> method to it."

### DIFF
--- a/src/pynwb/file.py
+++ b/src/pynwb/file.py
@@ -1,9 +1,12 @@
 from datetime import datetime
 from dateutil.parser import parse as parse_date
 from collections import Iterable
+from inspect import getmembers, ismodule
 
 from .form.utils import docval, getargs, fmt_docval_args, call_docval_func, get_docval
 from .form import Container
+from .form import backends
+from .form.backends.io import FORMIO
 
 from . import register_class, CORE_NAMESPACE
 from .base import TimeSeries, ProcessingModule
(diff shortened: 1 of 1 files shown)


neuromusic:
example use:

``` Python
from datetime import datetime
from pynwb import NWBFile

nwb = NWBFile('the PyNWB tutorial', 'my first synthetic recording', 'EXAMPLE_ID', datetime.now(),
            experimenter='Dr. Bilbo Baggins',
            lab='Bag End Labatory',
            institution='University of Middle Earth at the Shire',
            experiment_description='I went on an adventure with thirteen dwarves to reclaim vast treasures.',
            session_id='LONELYMTN')

nwb.to_hdf5('example.nwb')
```

jcfr (changes requested):
Please, rebase and squash the commit together. 

I also suggest you install and run `flake8` locally to fix the style.

```
$ pip install flake8
$ flake8
```

neuromusic:
thanks @jcfr but I don't think @ajtritt wants this PR merged in (see the convo in #117)

oruebel:
Yes, I believe this pull request is superseded by #171

neuromusic:
to be clear, #171 does not address #117 

perhaps this should stay open until #117 is resolved?

jcfr:
Commit have been squashed, style fixed and topic updated.

src/pynwb/file.py
```diff
@@ -146,6 +149,31 @@ def __init__(self, **kwargs):
         for attr in recommended:
             setattr(self, attr, kwargs.get(attr, None))
 
+        io_to_wrap = []
+        for name, obj in getmembers(backends):
+            # find the backend modules
+            if ismodule(obj) & hasattr(obj, '__path__'):
```
jcfr, line 155: `&` -> `and`

jcfr: Could you add a `is_backend` function ? This seems to be a fragile way of checking if an object is  a backend

src/pynwb/file.py
```diff
@@ -146,6 +149,31 @@ def __init__(self, **kwargs):
...
+                # find the backend IO classes
+                for name2, obj2 in getmembers(obj):
+                    try:
+                        if issubclass(obj2, FORMIO):
```
jcfr, line 159: Adding a `is_backend_io` function may be relevant too.

Also since there is a test on subclass, not sure why catching `TypeError` is needed

jcfr (changes requested):
Improve robustness and adding tests would be ideal

Issue #120, opened 2017-09-20
Create appveyor team to setup continuous integration


jcfr:
In order to have a Appveyor organization named "NeurodataWithoutBorders", 

Here are the steps to follow:

1. Create the following gmail account like this one:

```
  email: NeurodataWithoutBorders.appveyor@gmail.com
```

2. Create an organizational Appveyor account associated with this email.

3. Add the "pynwb" project to it. Then the following URL will be valid:

   https://ci.appveyor.com/project/NeurodataWithoutBorders/pynwb

4. Add @dorukozturk and Myself (@jcfr) as collaborators (with administrative privileges) to the NeurodataWithoutBorders user

5. When completed, following this steps will allow developer signing up to their appveyor account to choose the "NeurodataWithoutBorders" team/user.

References:

* https://www.appveyor.com/docs/team-setup/#setting-up-appveyor-account-for-github-organization
* https://github.com/scikit-build/scikit-build/issues/94#issuecomment-234357835

jcfr:
cc: @mgrauer

mgrauer:
Thanks @jcfr .

It is a bit unclear to me what is the next step here, is there something that someone else needs to continue with, or is this simply a record of what you have done, and the task is now complete?

jcfr:
Apologize for the confusion, I copied the text from a previous communication and omitted to update the tense. Changing to present tense should make things clearer. 

> someone else needs to continue 

Indeed, one of the core dev should go ahead and create the email and corresponding AppVeyor account.  The idea, is to share the credential to that email account with few others  ...

oruebel:
@dorukozturk @jcfr I've create the email and AppVeyor account. Can you send me your email addresses that I should use so that I can add to the AppVeyor

jcfr:
Thanks @oruebel , the email associated with my appveyor account email is my kitware one.

oruebel:
@jcfr you are now an admin on the AppVeyor. For some reason I'm having trouble adding PyNWB to account. Can you take it from here?

jcfr:
@oruebel  I confirm I can now log to AppVeyor selecting `NeurodataWithoutBorders`. 

That said, I can not  add the project. Which is strange because the following is mentioned:

```
AppVeyor was authorized 32 minutes ago for admin:repo_hook,read:org,repo:status access to GitHub repositories as nwb-bot account
```


For @nwb-bot **GitHub** account, could you check that:
[...]

PR #122, opened 2017-09-22
CI Enhancements

dorukozturk:
This PR enhances the CI aspect of pynwb.
Here are the keys component of this PR:

1) Added tox to reduce the complexity of running tests on multiple python versions.
2) Added Appveyor as the primary Windows test runner.
3) Added Circleci as the primary Linux test runner.
4) Configured Travis to be the primary Osx test runner.
5) Added placeholder badges for different CI providers. (We will need to replace the urls with the real ones. Currently they are pointing to a fork that I created)
6) Fixed some minor issues on closing test/temporary files.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,30 @@
+version: 2
+jobs:
+   build:
+     docker:
+       - image: ubuntu:xenial
+     steps:
+       - checkout
+       - run:
+           name: Install system requirements
+           command: |
+             apt-get update
+             apt-get install -y software-properties-common python-software-properties
+       - run:
+           name: Install python3.6 and pip
+           command: |
+             add-apt-repository ppa:jonathonf/python-3.6
+             apt-get update
+             apt-get install -y python-pip python3.6
+       - run:
+           name: Install tox
+           command: |
+             pip install tox
+       - run:
+           name: Run the tests for Python 2.7
+           command: |
+             tox -e py27
+       - run:
+           name: Run the tests for Python 3.6
+           command: |
+             tox -e py36
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ docs/tmpl.index.rst
 # coverage output
 tests/coverage/htmlcov
 .coverage
+
+# tox
+.tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,25 @@
-language: python
+os: osx
+
+language: generic
+
+branches:
+ only:
+  - dev
+
 sudo: false
 
 python:
 - '2.7'
 - '3.6'
 
+before_install:
+  - brew outdated pyenv || brew upgrade pyenv
+  - pyenv install 3.6.2
+  - pyenv global 3.6.2
+
 install:
-- travis_retry pip install -r requirements.txt
-- travis_retry pip install -e .
+- pip install tox-travis tox-pyenv
 
 script:
-- make test
+- tox -e py27
+- tox -e py36
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include license.txt
--- a/README.rst
+++ b/README.rst
@@ -4,7 +4,8 @@
 .. image:: https://circleci.com/gh/NeurodataWithoutBorders/pynwb.svg?style=shield
     :target: https://circleci.com/gh/NeurodataWithoutBorders/pynwb
 
-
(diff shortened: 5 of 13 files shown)


dorukozturk:
@ajtritt or @oruebel Can one of you activate Circleci and Appveyor for this repo?

jcfr:
xref #120

ajtritt (dismissed):
Looks good to me, thanks for doing this.

oruebel:
I'm working on creating the AppVeyor accounts, i.e.,  #120 right now

nicain (dismissed):
Looks all good to me!

oruebel:
@dorukozturk  quick question regarding the following change:

```
     def setUp(self):
-        self.test_temp_file = tempfile.NamedTemporaryFile()
-        self.f = h5py.File(self.test_temp_file.name, 'w')
+        self.test_temp_file = tempfile.NamedTemporaryFile(delete=True)
+        name = self.test_temp_file.name
+        self.test_temp_file.close()
+        self.f = h5py.File(name, 'w')

```

If I remember the behavior of NamedTemporaryFile correctly the file will be deleted automatically once ```self.test_temp_file.close() ``` gets called. This means that now ```self.f = h5py.File(name, 'w')``` should actually create a new file with the same name as the temporary file, rather than actually using the file that was created by NamedTemporaryFile. If this is the case, then I believe the file that is being created may not get deleted properly once the tests are done. I have not tested this, so I might be wrong, but I think removing ```self.test_temp_file.close()``` or adding an explicit rm of the file to the tearDown would fix this.

jcfr:
>  the behavior of NamedTemporaryFile correctly the file will be deleted

That is correct, I wonder with the issue was related to python 14243 (see [here](https://github.com/scikit-build/scikit-ci/blob/9cae36620eddafec41d611c52a490e098cf7b7fa/ci/driver.py#L165-L166) for details)

Either way, if this workarounds a particular problem, a comment should be added.

oruebel:
@jcfr I'm not sure. The link seems to suggest to set```delete=False``` on NamedTemoraryFile. However in the code here sets```delete=True```. Either way,  since the temp file is being closed in the setUp function, I believe we need to explicitly delete the created file in the tearDown function. It's fine if this is needed to work around another problem, however, we should make sure that the file is being deleted when the tests are done. If it is a workaround, then it would be also be good to add a quick comment to the code as well  just so we don't accidentally unfix the fix later.

jcfr:
> we need to explicitly delete the created file in the tearDown function.

:+1:

dorukozturk:
@oruebel According to h5py.File documentation [here](http://docs.h5py.org/en/latest/high/file.html), if the file exists, in the write mode it will try to truncate. On Windows though if that file is open, operating system won't let h5py to delete and recreate that file. It was raising an error on my builds. It was not executing the tests, thus tearDown method was not getting called at all. I had to solve that problem on the setUp phase. Because once we created the file and it was open by the Python process, the OS was blocking anything else to delete that file. My workaround was using NamedTempFile as a temporary path generator, get the flie path it creates, dump the file and let h5py recreate a fresh file with the path generated by NamedTempFile. I must have forget to delete that kwarg (delete=True) which is unnecessary. It is the default behavior anyways. I will delete it. I hope that clarifies. 

Also @oruebel you are right, we should delete the files in the tearDown. I will add that when I have the chance.

jcfr:
@dorukozturk  AppVeyor is now available: https://ci.appveyor.com/project/NeurodataWithoutBorders/pynwb

jcfr:
@oruebel  To setup CircleCI, look like you need to either add the project or grant one of us with admin access to do so.

circleci-pywnb-admin-required-to-add

oruebel:
@dorukozturk thanks for the clarification on the reason for the change. That makes sense. Can you update the tearDown function to delete the file. Essentially what I believe is missing is:

```
import os
os.remove(f.filename)
```
Also, could you add a quick comment to the setUp() function with the reason why we need to close the NamedTemporaryFile.

dorukozturk:
@oruebel I activated Circleci, thus this is ready for the final review.

jcfr:
Well done 👍 

@dorukozturk Look like the CircleCI and Appveyor build associated haven't been submitted 
* https://ci.appveyor.com/project/NeurodataWithoutBorders/pynwb/history
* https://circleci.com/gh/NeurodataWithoutBorders/pynwb

May be you could update the last commit of the PR and force push to re-trigger them ...

Here are the status from PR #122 I could retrieve:
* TravisCI: https://travis-ci.org/NeurodataWithoutBorders/pynwb/builds/280139880

Issue #159, opened 2017-10-20
CI/coverage: Associate form, pynwb and integration tests with a different flag


jcfr:
From @ajtritt 

> I was wondering if there was a way to figure out which new code needs unit tests by running just the unit tests.

It is indeed possible. See https://docs.codecov.io/docs/flags

I suggest we run the test driver three times:

1.  with `--form` , associating the coverage flag `form`

2.  with `--pynwb`, associating the coverage flag `pynwb`

3.  with `--integration`, associating the coverage flag `integration`

oruebel:
Closing. Was resolved in #160.

PR #160, opened 2017-10-20
ci: Run each test-suite independently and label associated coverage report

jcfr:
See #159

Suggested-by: Andrew Tritt <ajtritt@lbl.gov>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,8 +34,12 @@ jobs:
              pip3 install -r requirements.txt
              pip3 install -r requirements-dev.txt
              pip3 install -e .
-             coverage run --source src test.py
-             codecov
+             coverage run --source src test.py --pynwb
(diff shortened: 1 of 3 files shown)


jcfr:
Look like the python implementation of `codecov` CLI does not support the `-C` parameter. See https://docs.codecov.io/docs/flags

jcfr:
For each tile, there is now a list of toggle button associated with each flags:

ci-coverage-toggle

jcfr:
That said, I suspect we need to clean up the previous coverage report between each run of test suite.

This is available in the bash implementation of the codecov client but not in the python one

dorukozturk (approved):

jcfr:
Also, it look like I am not authorized to merge pull request.

Would it be possible to grant me the right or proceed with the merge .. thanks

PR #163, opened 2017-10-23
Pin specific package versions in requirements

### LINKED ISSUE #100: specify required dependencies in setup.py
currently, setup.py does not have `install_requires` & the user must `$ pip install -r requirements.txt` to install dependencies.

- `setup.py` should specify minimal dependencies so that `pip` can install pynwb and its dependencies
- `requirements.txt` should specify the precise environment that the package is tested against

dorukozturk:
Fixes #100


setup.py
```diff
@@ -21,6 +22,11 @@
 print('found these packages:', pkgs)
 
 schema_dir = 'data'
+
+# HACK: numpy needs to be installed before scipy
+pip.main(['install', 'numpy'])
+pip.main(['install', 'scipy'])
```
jcfr: This should be done extending the Install command and not directly here. 

https://cbuelter.wordpress.com/2015/10/25/extend-the-setuptools-install-command/
https://github.com/scikit-build/scikit-build/tree/master/skbuild/command

jcfr:
What is the rational for pinning ? pynwb being a library, pinning should be left to application using it.

The CI should always use `-U` when installing.

http://nvie.com/posts/pin-your-packages/

oruebel:
Specifying exact versions of packages seems like it might be overly restrictive for users that want to build applications on top of PyNWB. I can see that we may need to specify for some libraries minimum version numbers but I'm not sure about requiring exact combinations of package versions.

dorukozturk:
I was reading [this](https://packaging.python.org/discussions/install-requires-vs-requirements/) and as far as I understand there are 2 ways of describing dependencies of a project. The "abstract" way and the "concrete" way. install_requires section in the setup.py should be abstract without any version information pinned and it should not have sub-dependencies and so on. Since the preferred way of installing pynwb is with:
```
python setup.py build
python setup.py install
```
according to http://pynwb.readthedocs.io/en/latest/getting_started.html#installation, we won't be forcing versions for people following the instructions.

Requirements.txt on the other hand should describe an entire Python environment with specific versions so that a virtualenv or something else can be created by using the requirements.txt file. I think the correct way to create the requirements.txt is to do the following (for pynwb):
```
mkvirtualenv pynwb-env --> Create Fresh environment
python setup.py build
python setup.py install --> Install all pynwb dependencies
pip freeze > requirements.txt --> Snapshot the environment so it is reproducable
```

Does that make sense?

jcfr:
> Abstract vs concrete

As you pointed out, it makes sense to have "abstract" dependencies in `setup.py`.

Regarding the concrete dependencies, here are few comments:

* By pinning dependencies, the CI currently doesn't test for the case of a user installing the package with one of the dependency being more recent that what we test with.

* To mitigate this,  few approaches:

  1) Setup service like https://requires.io/ or https://pyup.io/ to notify the project automatically (and even create a pull request automatically)

  2) specify package in requirements using the for `foo >= X`, and ensure a build+test of pynwb is triggered every day


The advantage of (1) is that we will more easily understand which update of version cause things to break.


> Since the preferred way of installing pynwb

Generally speaking and moving forward. pip should be used to install python packages.

As soon as we have package published on pypi, the recommended way to install will definitively be `pip install pynwb`

oruebel:
@jcfr that sounds good to me. I'm fine with either approach for the requirements checking.

dorukozturk:
@jcfr That makes perfect sense. I will setup requires.io. I will also change the installation instructions in another PR so that people install pynwb with pip. Does that sound good?

jcfr:
👍 

>  that people install pynwb with pip

It would be nice to describe both "mode":

* regular:

```
cd pynwb-src
pip install .  #preferred

or 

python setup install
```

* developer install:

```
pip install -e . # preferred

or 

python setup.py develop .
```

neuromusic:
I ❤️ `pip install -e .`

requirements.txt
```diff
@@ -1,7 +1,11 @@
...
+chardet==3.0.4
+h5py==2.7.1
+idna==2.6
+numpy==1.13.3
+python-dateutil==2.6.1
+requests==2.18.4
+ruamel.yaml==0.15.34
+scipy==0.19.1
+six==1.11.0
+urllib3==1.22
```
jcfr, line 11: Waiting we have a maintainer documentation section in readthedocs, may be worth adding a comment in the requirements.txt itself explaining how to obtain the list of pinned requirements. (Using pip freeze ..)

jcfr (approved):
Merging now. Documentation writing will be part of an other PR

Issue #137, opened 2017-09-29
CI: Support automatic versioning using versioneer


jcfr:
Currently the version is hard-coded in setup.py, the idea use [python-versioneer](https://github.com/warner/python-versioneer)

It allows to:

* have a unique version number based on the latest tag (if no tag is present, version starts with `0`)

* automatically integrate version number in source distribution

* properly label development (or nightly) release 

* simplify the release process reducing the number of actions

[Here](https://github.com/scikit-build/scikit-ci/blob/8fb1d361c36082608636311d0ed115f7b202ac25/setup.cfg#L7-L12) is example of configuration associated with `setup.cfg`

jcfr:
@oruebel  After dev is merged to master and a release is created, we will have to tag the repository. Options are to use strings like `0.2.0` or `v0.2.0`. Which one do you prefer ?

oruebel:
I'm ok with either format. My vote is for the variant with the `v` but either way is fine. The bigger question for me is, what version number we should use for the SfN release. For the NWB format it will be something like 2.0.0beta but I'm not sure which version we should use for PyNWB. We should discuss this during our next telecon. @ajtritt any thoughts?

PR #164, opened 2017-10-24
Publish sources and wheels on github

### LINKED ISSUE #136: CI/Circle: Upload source and binary distributions as "latest" release on Github
* [ ] Use [publish-github-release-py add-on](http://scikit-ci-addons.readthedocs.io/en/latest/addons.html#publish-github-release-py) to upload the packages.
* [ ] Also add `How to make a release` section to the documentation. For example, see [here](https://github.com/scikit-build/scikit-build/blob/master/docs/make_a_release.rst). Note that the example is missing the submission of PR to the conda-forge recipe

dorukozturk:
Fixes #136

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,7 @@ jobs:
            command: |
              apt-get update
              apt-get install -y software-properties-common python-software-properties
+             apt-get install -y libblas-dev liblapack-dev
        - run:
            name: Install python3.6 and pip
            command: |
@@ -48,3 +49,9 @@ jobs:
            name: Test installation from a wheel for Python 3.6
            command: |
              tox -e py36 recreate --installpkg dist/pynwb-0.1-py3-none-any.whl
+       - run:
+           name: Publish wheel and source distribution on github
+           command: |
+             pip install scikit-ci-addons
+             ci_addons publish_github_release NeurodataWithoutBorders/pynwb \
+                 --prerelease-packages "dist/*" --exit-success-if-missing-token
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,10 +25,14 @@ jobs:
            name: Run the tests for Python 2.7
            command: |
              tox -e py27
+           # Install is expected to be quick. Increase timeout in case there are some network issues.
+           # While pip installing tox does not output by default. Circle thinks task is dead after 10 min.
+           no_output_timeout: 30m
        - run:
            name: Run the tests for Python 3.6
            command: |
              tox -e py36
+           no_output_timeout: 30m
        - run:
            name: Report coverage
            command: |
--- a/tox.ini
+++ b/tox.ini
@@ -16,4 +16,5 @@ deps =
 
 commands =
     python test.py
+    python setup.py sdist
     python setup.py bdist_wheel


jcfr:
@dorukozturk  The rational behind using http://scikit-ci-addons.readthedocs.io/en/latest/addons.html#publish-github-release-py is that it will:
* update the reference associated with the latest tag
* remove previous packages and upload the new one

Under the hood, it makes uses of https://github.com/j0057/github-release#readme

jcfr:
In a nutshell, the command would be:

```
pip install scikit-ci-addons
ci_addons publish_github_release NeurodataWithoutBorders/pynwb \
           --prerelease-packages "dist/*"
```

Specifying `--exit-success-if-missing-token` would ensure that the build would not fail if triggered for PR not able to "see" the GITHUB_TOKEN env. variable (needed to upload the package)

The GITHUB_TOKEN should be created by the @nwb-bot user and added as a secure env variable.

jcfr:
Well done.

Since you are admin on the repo, I would expect the github token to be set.

It currently report:
```
skipping: A token is expected.

Specify the --token parameter or set GITHUB_TOKEN environment variable.
See https://help.github.com/articles/creating-an-access-token-for-command-line-use/

--------------------------------------------------------------------------------
usage: publish_github_release.py [-h]
                                 [--release-packages [PATTERN [PATTERN ...]]]
                                 [--prerelease-packages [PATTERN [PATTERN ...]]]
                                 [--prerelease-packages-clear-pattern PATTERN]
                                 [--prerelease-packages-keep-pattern PATTERN]
                                 [--prerelease-tag PRERELEASE_TAG]
                                 [--prerelease-name PRERELEASE_NAME]
                                 [--prerelease-sha PRERELEASE_SHA]
                                 [--token GITHUB_TOKEN]
                                 [--exit-success-if-missing-token]
                                 [--re-upload]
                                 [--display-python-wheel-platform] [--dry-run]
                                 ORG/PROJECT

```

oruebel:
@jcfr how does versioning of those releases work. Also for major releases, is there a good way to include release notes with an overview of major changes?

dorukozturk:
@jcfr I think this is ready for the last review. PTAL.

jcfr:
>  how does versioning of those releases work. Also for major releases, is there a good way to include release notes with an overview of major changes?

@oruebel This corresponds to the latest release buildable of the `dev` branch and installable using `pip install pynwb -f https://github.com/NeurodataWithoutBorders/pynwb/releases/tag/latest`, 

when a stable release (e,g vX.Y.Z) will have to be done:
* the release will be tagged, then CI will upload package with the proper naming convention (see #137)
* "release" ready packages will be uploaded to the latest release 
* the developer in charge of the release will then:
  * download the package from the "latest" staging area (it can be done with a one liner)
  * signed and uploaded to pypi.test using `twine` python cli
  * check that it looks good
  * signed and uploaded to pypi using `twine` python cli
  * then we will have instruction to also easily create the github release associated

.circleci/config.yml
```diff
@@ -25,10 +25,12 @@ jobs:
            name: Run the tests for Python 2.7
            command: |
              tox -e py27
+           no_output_timeout: 30m
```
jcfr: Why installing these takes so long ? Since there is already wheels available it should be fairly quick.

dorukozturk: It started to happen this morning. Maybe circle invalidated their cache for the xenial docker images? For some reason pip installing scipy takes forever.

jcfr: I suggest to do a verbose install to find out more details. Even without cache, it should be quick.

I suspect the wheels end up being build from source ...

dorukozturk: I think you are right. Any ideas why that would happen all of a sudden?

jcfr: Could you point me to the verbose log ?

dorukozturk: Funny thing is it started to use the wheel again. I just rebuild without cache and it just got significantly faster. https://circleci.com/gh/NeurodataWithoutBorders/pynwb/168. I suggest we keep the timout flags there though in case it tries to build wheel again since we don't know what caused that. Does that makes sense?

jcfr: > we keep the timout flags

Sure. In that case, add a comment indicating why the timeout is there.

Ideally, timeout should be added to all CI.

>

jcfr: May be a comment like this:

```
no_output_timeout: 30m  # Install is expected to be quick. Timeout early in case there are some network issues.
```

dorukozturk: Can we do it on other CIs if this occurs? This was happening specifically because circle thinks if the task waits more than 10 minutes without stdouting anything the task is a zombie. While installation tox does not print out anything by default so it was waiting more than 10 minutes. It did not cause problems with others.

jcfr: > Can we do it on other CIs if this occurs?

Sounds good

> While installation tox does not print out anything by default

That is the type of info to specify in the comment.

jcfr (approved):
LGTM

In some future iteration, we may have a `latest-dev` and `latest` release.

PR #169, opened 2017-10-26
Remove scipy from dependencies

### LINKED ISSUE #168: Reconsider usage of scipy
From what I can tell it is only used here https://github.com/NeurodataWithoutBorders/pynwb/blob/17699d46b6bc3902b8fa47fb24f4f376b1b7ddc0/tests/build_fake_data.py#L58
Can we replace that line without bringing scipy and all of it's dependencies?

dorukozturk:
Fixes #168.
Thanks to @neuromusic for the one-liner.
Makes the builds faster and it is easier for the end user not to install scipy especially on Windows.

### DIFF
--- a/tests/build_fake_data.py
+++ b/tests/build_fake_data.py
@@ -5,7 +5,6 @@ from pynwb import NWBFile, get_build_manager
 from form.backends.hdf5 import HDF5IO
 
 import numpy as np
-import scipy.stats as sps
 import os
 from datetime import datetime
 
@@ -55,7 +54,7 @@ np.random.seed(1234)
 ephys_data = np.random.rand(data_len)
 ephys_timestamps = np.arange(data_len) / rate
 spatial_timestamps = ephys_timestamps[::10]
-spatial_data = np.cumsum(sps.norm.rvs(size=(2,len(spatial_timestamps))), axis=-1).T
+spatial_data = np.cumsum(np.random.normal(size=(2,len(spatial_timestamps))), axis=-1).T
 
 ephys_ts = ElectricalSeries('test_ephys_data',
                             'test_source',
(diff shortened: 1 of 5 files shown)


tests/build_fake_data.py
```diff
@@ -76,7 +75,7 @@
                            timestamps=spatial_timestamps,
                            resolution=0.1,
                            comments="This data was generated with numpy, using 1234 as the seed",
```
neuromusic, line 77: FWIW, this comment is incorrect. 1234 is used as the seed before the fake electrical series gets generated. This data uses the next seed.

ajtritt (approved):

tests/build_fake_data.py
dorukozturk: @ajtritt Should we address the random number issues that @neuromusic and @nicain raised in another PR?

ajtritt: There is no issue. The random number generator is already seeded with a constant seed.

neuromusic: The seed isn't reset between 

https://github.com/NeurodataWithoutBorders/pynwb/blob/43dfb293b79a4303d4ea1629706132c19b95c0da/tests/build_fake_data.py#L54

and

https://github.com/NeurodataWithoutBorders/pynwb/blob/43dfb293b79a4303d4ea1629706132c19b95c0da/tests/build_fake_data.py#L57

so the comment in 

https://github.com/NeurodataWithoutBorders/pynwb/blob/43dfb293b79a4303d4ea1629706132c19b95c0da/tests/build_fake_data.py#L77

is false.

If the parameters used to generate the size of the first array change, this will change the second array.

image

PR #170, opened 2017-10-26
Add versioneer for dynamic versioning

### LINKED ISSUE #137: CI: Support automatic versioning using versioneer

dorukozturk:
Fixes #137.

Sorry for the gigantic PR. Most of it is auto-generated (reviewer can ignore src/pynwb/_version.py and versioneer.py) code which enables dynamic versioning using [versioneer](https://github.com/warner/python-versioneer). Couple of things changed in the CI yaml files to make the wheel installation dynamic. Before wheel names were hardcoded. Now there is some regex action going on to install the wheels.


jcfr:
<s>Do we really need to include the source of versionner ? </s>

Answering my own question: yes

jcfr (approved):
Well done. :+1

Issue #180, opened 2017-11-01
Potential typo


dorukozturk:
https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/io/core.py#L14 Should this line be def carg_data?

nicain:
Yes! Nice catch

ajtritt:
Yes, it should be, but it doesn't matter, the name of the methods decorated with `ObjectMapper.constructor_arg` and `ObjectMapper.obj_attr` does not matter.

PR #181, opened 2017-11-01
Style according to pep8

### LINKED ISSUE #178: Style pynwb according to pep8
For standardized styling projects should follow [pep8](https://www.python.org/dev/peps/pep-0008/).

dorukozturk:
Fixes #178 #180
Applies flake8 styling to pynwb. 
@jcfr PTAL.

### DIFF
--- a/src/pynwb/form/spec/namespace.py
+++ b/src/pynwb/form/spec/namespace.py
@@ -17,12 +17,16 @@ _namespace_args = [
     {'name': 'schema', 'type': list, 'doc': 'location of schema specification files or other Namespaces'},
     {'name': 'full_name', 'type': str, 'doc': 'extended full name of this namespace', 'default': None},
     {'name': 'version', 'type': (str, tuple, list), 'doc': 'Version number of the namespace', 'default': None},
-    {'name': 'date', 'type': (datetime, str), 'doc': "Date last modified or released. Formatting is %Y-%m-%d %H:%M:%S, e.g, 2017-04-25 17:14:13",
+    {'name': 'date', 'type': (datetime, str),
+     'doc': "Date last modified or released. Formatting is %Y-%m-%d %H:%M:%S, e.g, 2017-04-25 17:14:13",
      'default': None},
     {'name': 'author', 'type': (str, list), 'doc': 'Author or list of authors.', 'default': None},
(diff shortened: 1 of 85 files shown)


src/pynwb/io/__init__.py
```diff
@@ -1,3 +1,4 @@
+# flake8: noqa
```
jcfr: I suggest to exclude specific error instead. E,g `#  noqa: EXXX`

jcfr (dismissed):

oruebel:
@dorukozturk is there a way to customize the flake8 checking? E.g., it would be nice to relax the 80 characters per line limit to something more reasonable.

jcfr:
> it would be nice to relax the 80 characters per line limit to something more reasonable.

Indeed, length has already been  increased to 100. I suggest to bump it to 120

See https://github.com/NeurodataWithoutBorders/pynwb/pull/181/files#diff-380c6a8ebbbce17d55d50ef17d3cf906

ajtritt:
I don't use punchcards and I got rid of my CRT monitor, so I can do 120 characters.

oruebel:
I agree, 120 characters is much more reasonable.

dorukozturk:
This is still work in progress, I push time to time to see if I broke something. I set the limit to 120 characters.

src/pynwb/form/spec/namespace.py
```diff
@@ -256,15 +264,17 @@ def __get_spec_path(cls, ns_path, spec_path):
...
+            {'name': 'resolve', 'type': bool,
+             'doc': 'whether or not to include objects from included/parent spec objects', 'default': True},
             returns='a dictionary describing the dependencies of loaded namespaces', rtype=dict)
     def load_namespaces(self, **kwargs):
         """Load the namespaces in the given file"""
         namespace_path, resolve = getargs('namespace_path', 'resolve', kwargs)
         # load namespace definition from file
         if not os.path.exists(namespace_path):
-            raise FileNotFoundError("namespace file '%s' not found" % namespace_path)
+            raise OSError("namespace file '%s' not found" % namespace_path)
```
dorukozturk: See https://github.com/nimiq/promptastic/issues/17. FileNotFoundError is only a Python 3 Exception.

jcfr (dismissed):

PR #198, opened 2017-11-07
Fix broken release process by removing the old release candidate

dorukozturk:
Fixes circleci tests for dev

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,5 +63,5 @@ jobs:
              if [ "${CIRCLE_BRANCH}" == "dev" ]; then
                pip install scikit-ci-addons
                ci_addons publish_github_release NeurodataWithoutBorders/pynwb \
-                 --prerelease-packages "dist/*" --exit-success-if-missing-token
+                 --prerelease-packages "dist/*" --exit-success-if-missing-token --re-upload
              fi


.circleci/config.yml
```diff
@@ -58,5 +58,5 @@ jobs:
              if [ "${CIRCLE_BRANCH}" == "dev" ]; then
                pip install scikit-ci-addons
                ci_addons publish_github_release NeurodataWithoutBorders/pynwb \
-                 --prerelease-packages "dist/*" --exit-success-if-missing-token
+                 --prerelease-packages "dist/*" --exit-success-if-missing-token --re-upload
```
jcfr, line 66: An other approach would be:

```
--prerelease-packages "dist/*" \
--prerelease-packages-clear-pattern "*" \
--prerelease-packages-keep-pattern "*.g<COMMIT_DATE>*" \
--exit-success-if-missing-token
```

that said, this wouldn't allow us to collect the package generated after tagging the repo.

If using `--re-upload` works, let's go with it.

jcfr (approved):

jcfr (changes requested):
Thinking more about it, simply re-uploading is not enough to clear up the old package

jcfr (approved):
Ooop. Having a look at the code, we should be good. I will update the documentation of the add-on to clarify things.

jcfr:
See https://github.com/scikit-build/scikit-ci-addons/blob/9ce5c15363c57c22549e0eb16f11a445372709cf/anyci/publish_github_release.py#L315-L318

Issue #203, opened 2017-11-07
remove get_build_manager dependency from validator


ajtritt:
pynwb/validate.py still calls get_build_manager. This code needs to be updated to call pynwb.get_manager

oruebel:
Pull request #187 I believe fixes that part

jcfr:
Closing, This is fixed in https://github.com/NeurodataWithoutBorders/pynwb/commit/f2d12e6bd155cb52331c1f11b58e4737317d4adc

PR #204, opened 2017-11-07
pynwb: Update validator to use get_manager() function. See #203

jcfr:


jcfr:
CircleCI and Appveyor are green. Integrating without waiting for slower Travis.

PR #187, opened 2017-11-02
Validator function call bug fixes

oruebel:
Updated validator to use get_manager() instead of get_build_manager() and write_out() instead of write()

PR #205, opened 2017-11-07
pynwb/validate: Simplify message reporting

jcfr:


src/pynwb/validate.py
```diff
@@ -31,22 +31,22 @@ def print_errors(errors):
     args = parser.parse_args()
 
     if not os.path.exists(args.path):
-        write('%s not found' % path, file=sys.stderr)  # noqa: F821
+        write_out('%s not found' % args.path)
```
jcfr: @oruebel  This new PR cleans up the call to `write_out` removing extra unused parameters like `file=sys.stderr` and `end=''` that were present in #187 

Is expected to write error on `stdout` instead of `stderr` ? 

Also `sys.out` is invalid, it should be `sys.stdout`

oruebel: Probably stderr.

jcfr:
@oruebel  Do you have an example of input for the validate CLI ? I would like to add simple test to ensure  the cli can run.

oruebel:
@jcfr I don't think we have any test for the validator yet. You could probably use the file that @nicain added (or was planing to add) for testing the legacy data read also to run the validator on. Ultimately, I think we need to add unit tests for the validator.

jcfr:
Merging. Appveyor and circleci are green.

And trying to manually run the "validate" CLI with invalid path display an error message:

```
$ python ./src/pynwb/validate.py /path/to/invalid
/path/to/invalid not found

 $ echo $?
1
```

jcfr:
Integrated in https://github.com/NeurodataWithoutBorders/pynwb/commit/bf182d196922311d95af6248dc91da4b33639be6

PR #226, opened 2017-11-09
Change the convention that versioneer uses

### LINKED ISSUE #216: ci: Update versioneer to use not including a "+" sign. Not support by GitHub
Since GitHub will change the "+" into a "." when package are uploaded, we can not use the "pep440-post" style.

Available options are documented here: https://github.com/warner/python-versioneer/blob/master/details.md#how-do-i-select-a-version-style

May be the `pep440`

dorukozturk:
## Motivation

Fixes #216.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,8 +2,8 @@
 VCS = git
 versionfile_source = src/pynwb/_version.py
 versionfile_build = pynwb/_version.py
-style = pep440-post
-tag_prefix = '*.*.*'
+style = pep440-pre
+tag_prefix = *.*.*
 
 [flake8]
 max-line-length = 120
--- a/src/pynwb/_version.py
+++ b/src/pynwb/_version.py
@@ -40,8 +40,8 @@ def get_config():
     # _version.py
     cfg = VersioneerConfig()
     cfg.VCS = "git"
-    cfg.style = "pep440-post"
-    cfg.tag_prefix = ""
+    cfg.style = "pep440-pre"
+    cfg.tag_prefix = "*.*.*"
     cfg.parentdir_prefix = "None"
     cfg.versionfile_source = "src/pynwb/_version.py"
     cfg.verbose = False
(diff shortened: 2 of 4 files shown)


src/pynwb/_version.py
```diff
@@ -40,8 +40,8 @@ def get_config():
     # _version.py
     cfg = VersioneerConfig()
     cfg.VCS = "git"
-    cfg.style = "pep440-post"
-    cfg.tag_prefix = ""
+    cfg.style = "pep440-pre"
+    cfg.tag_prefix = "'*.*.*'"
```
jcfr: Is it expected to have single quote within double quote ?

dorukozturk: Nope my bad, fixing it right now.

setup.cfg
```diff
@@ -2,7 +2,7 @@
 VCS = git
```
jcfr, line 2: Suggested commit message:

```
packaging: Support download from GitHub release

This commit updates the convention used by versioneer
to pep440-pre [1]. Doing so allows to install the development packages
from the GitHub release. It ensures that no plus character
is included in the original package name. It was 
switched by a dot by GitHub and prevented successful install
of the package using:

  pip install pynwb --find-links https://github.com/NeurodataWithoutBorders/pynwb/releases/tag/latest  --no-index

[1] https://github.com/warner/python-versioneer/blob/master/details.md#how-do-i-select-a-version-style
```

jcfr:
Let's not this hold up the release, we need to also update the `keep-pattern` here:

https://github.com/NeurodataWithoutBorders/pynwb/blob/80c26fac6150725f0542617b6e80a0c0d60bf6fc/.circleci/config.yml#L69

with:

```
--prerelease-packages-keep-pattern "*dev<DISTANCE>*" \
```

But before doing so, I need to update the mini-language of the publish script to support `<DISTANCE>`.

See http://scikit-ci-addons.readthedocs.io/en/latest/addons.html#mini-language-for-packages-selection

src/pynwb/_version.py
dorukozturk: I take it back. It is autogenerated and expected.

PR #229, opened 2017-11-09
Add requirements doc

### LINKED ISSUE #218: ci: Add requirements-doc.txt
Requirement to build the documentation should be listed in a dedicated requirements file, this will facilitate building of the documentation locally.

For example: https://github.com/scikit-build/cmake-python-distributions/blob/master/docs/requirements-docs.txt

dorukozturk:
## Motivation

Fixes #218.


requirements-dev.txt
```diff
@@ -1,4 +1,15 @@
-codecov
-coverage
-tox
-unittest2
+certifi==2017.11.5
```
jcfr, line 1: Could we document somewhere the top level requirements  ? 

as a side note, for devel and doc, I would suggest to not pin. But I could be convinced otherwise ... for example for the doc, I am not sure if it is worth the maintenance hassle ..

Or if we keep them pinned, let's make that the require badge look at the files

dorukozturk: I think we should keep the dev requirements pinned since for testing we do 
```
pip install -r requirements.txt -r requirements-dev.txt
```
We want to reproduce the same environment over and over again. I agree with the requirements-doc.txt though. I think there is no reason to pin the doc requirements. Requires.io detects all the requirements files already(https://requires.io/github/NeurodataWithoutBorders/pynwb/requirements/?branch=add-requirements-doc).

jcfr: For now, let's move forward with this. We can revisit if it cause problem.

jcfr (approved):

jcfr:
Closing. Integrated in https://github.com/NeurodataWithoutBorders/pynwb/commit/87a7598304248f0c7da93cac5af12de45329cb67

PR #245, opened 2017-11-11
Move coverage report task to tox

### LINKED ISSUE #240: Fix CodeCov environment to match one of the test environments
## 1) Bug

When CircleCI runs codecov, it seems to be running with a different environment. Unit tests for both Py2 and Py3 pass, but codecov fails when running the same code.

See [this build](https://circleci.com/gh/NeurodataWithoutBorders/pynwb/414?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

## 2) Feature Request
Ensure that codecov runs the in the same environment

### Problem/Use Case
Deployment gets held up, even though all tests pass.

## Checklist

- [ ] Have you ensured the feature or change was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues/) ?
- [ ] Have you included a brief and descriptive title?
- [ ] Have you included a clear description of the problem you are trying to solve?
- [ ] Have you included a minimal code snippet that reproduces the issue you are encountering?

dorukozturk:
## Motivation

Coverage was ran against the default Python3 in circleci (which was Python 3.5). Now it is a step in the tox file and reported by py36 in tox. Fixes #240.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,8 @@ develop: build
 	$(PYTHON) setup.py develop
 
 test:
-	$(PYTHON) test.py
+	pip install -r requirements-dev.txt
+	tox
 
 test_docker:
 	docker build --quiet --no-cache --tag neurodatawithoutborders/pynwb:python27_test -f ./docker/python27_test/Dockerfile .
@@ -38,6 +39,7 @@ test_docker:
 	docker run --rm -it neurodatawithoutborders/pynwb:python36_test bash -c 'python test.py'
 
 apidoc:
+	pip install -r requirements-doc.txt
 	cd docs && $(MAKE) apidoc
 
 htmldoc-only: apidoc
@@ -59,9 +61,8 @@ pdfdoc:
 	@echo "To view the PDF documentation open: docs/_build/latex/PyNWB.pdf"
 
 coverage:
-	$(COVERAGE) run --source=. test.py
+	tox -e coverage
 
-coverage_html: coverage
-	$(COVERAGE) html -d tests/coverage/htmlcov
-	@echo ""
-	@echo "To view coverage data open: tests/coverage/htmlcov/index.html"
+coverage-open:
+	@echo "To view coverage data open: ./tests/coverage/htmlcov/index.html"
+	open ./tests/coverage/htmlcov/index.html || xdg-open ./tests/coverage/htmlcov/index.html
--- a/tox.ini
+++ b/tox.ini
@@ -18,3 +18,35 @@ commands =
     python test.py
     python setup.py sdist
     python setup.py bdist_wheel
+
+[testenv:coverage]
+usedevelop = True
+basepython=python3.6
+
+commands =
+    coverage run --source src test.py
+    coverage html -d tests/coverage/htmlcov
+
+[testenv:CI]
+commands =
+    coverage run --source src test.py --pynwb
+    codecov -F pynwb
+    coverage run --source src test.py --integration
+    codecov -F integration
+    coverage run --source src test.py --form
+    codecov -F form
+
+[testenv:CI-py27]
+basepython=python2.7
+passenv = *
+commands = {[testenv:CI]commands}
+
+[testenv:CI-py35]
+basepython=python3.5
+passenv = *
+commands = {[testenv:CI]commands}
+
+[testenv:CI-py36]
+basepython=python3.6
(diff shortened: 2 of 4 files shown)


tox.ini
```diff
@@ -18,3 +18,12 @@ commands =
     python test.py
     python setup.py sdist
     python setup.py bdist_wheel
+
```
jcfr, line 21: Let's do it on both py27 and py36, that way we will cover more.

dorukozturk: You think this is a good idea? I realized for local development when you do "tox" you get bunch of warnings because it will try to do codecov when you develop locally too. Another approach is to explicitly use the coverage in the .tox/py36/bin environment.

jcfr: > will try to do codecov when you develop locally too. 

This should be conditioned by an env variable.

Here is how you can pass env var: https://github.com/j0057/github-release/blob/master/tox.ini

jcfr:
I just rebased and published

dorukozturk:
@jcfr I think this is ready if all CIs pass.

jcfr:
@dorukozturk Well done. Could you now look into updating these commands: https://github.com/NeurodataWithoutBorders/pynwb/blob/d2a53149cef95ce0f6f70552301cfc6c647f30c7/Makefile#L61-L67

dorukozturk:
@jcfr I will update the Makefile. There is a problem with Appveyor coverage though. See [this](https://ci.appveyor.com/project/NeurodataWithoutBorders/pynwb/build/1.0.435/job/s97s08hqy3kjr86a)

Makefile
```diff
@@ -38,9 +39,11 @@ test_docker:
 	docker run --rm -it neurodatawithoutborders/pynwb:python36_test bash -c 'python test.py'
 
 apidoc:
+	pip install -r requirements-doc.txt
 	cd docs && $(MAKE) apidoc
 
 htmldoc-only: apidoc
+	pip install -r requirements-doc.txt
```
jcfr: This is not needed. Already done in `apidoc` target

tox.ini
```diff
@@ -18,3 +18,35 @@ commands =
...
+[testenv:coverage]
+commands =
+    coverage run --source src test.py
```
jcfr: Which python env will this be using ?

dorukozturk: Good point. Which one should it use? 3.6 maybe since 2.7 will be deprecated on 2020 and 3.6 is newer?

jcfr: 3.6

dorukozturk: Ok, I think this is ready one more time. :)

tox.ini
```diff
@@ -18,3 +18,39 @@ commands =
...
+basepython=python3.6
+
+commands =
+    coverage run --source src test.py
+
+[testenv:coveragehtml]
+basepython=python3.6
+
+commands =
+    coverage html -d test/coverage/htmlcov
```
jcfr: `test/coverage/htmlcov` -> `tests/coverage/htmlcov`

jcfr:
I suggest to simplify doing something like this:

```patch
diff --git a/Makefile b/Makefile
index 7656475..0a1ff3b 100644
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ pdfdoc:
 coverage:
        tox -e coverage
 
-coverage_html: coverage
-       tox -e coveragehtml
+coverage-open:
        @echo ""
-       @echo "To view coverage data open: tests/coverage/htmlcov/index.html"
+       @echo "To view coverage data open: ./tests/coverage/htmlcov/index.html"
+       open ./tests/coverage/htmlcov/index.html || xdg-open ./tests/coverage/htmlcov/index.html
diff --git a/tox.ini b/tox.ini
index e89f051..0981647 100644
--- a/tox.ini
+++ b/tox.ini
@@ -24,12 +24,7 @@ basepython=python3.6
 
 commands =
     coverage run --source src test.py
-
-[testenv:coveragehtml]
-basepython=python3.6
-
-commands =
-    coverage html -d test/coverage/htmlcov
+    coverage html -d tests/coverage/htmlcov
```

jcfr:
Closing. Integrated in https://github.com/NeurodataWithoutBorders/pynwb/commit/0889756ee2a3e772ab5334b39d0348ca0a9df4d8

PR #246, opened 2017-11-11
requirements: Per recommendation, constrain version to <0.15

jcfr:
> This is the new (0.15+) interface for ``ruamel.yaml``, it is still in
> the process of being fleshed out. Please pin your dependency to
> ruamel.yaml<0.15 for production software.

Source: http://yaml.readthedocs.io/en/latest/basicuse.html

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ setup_args = {
     [
         'numpy',
         'h5py',
-        'ruamel.yaml',
+        'ruamel.yaml<0.15',
         'python-dateutil',
         'six',
         'requests'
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -27,7 +27,7 @@
     [
         'numpy',
         'h5py',
-        'ruamel.yaml',
+        'ruamel.yaml<15',
```
nicain: Just double checking that this should be 15 and not .15

nicain (commented):
I really like using -e and -U before pip freeze

setup.py
jcfr: Good catch.

jcfr:
Since all test are currently passing,  and this addresses no particular problem. I am closing it.

I originally thought this was related to #248, but it is not.

@nicain @ajtritt @oruebel I think differently, we can still re-open and integrate.

PR #252, opened 2017-11-13
Update requirements-doc.txt

nicain:
## Motivation

@choldgraf mentions here that allensdk is a dependency for that notebook that I checked in.  @jcfr will adding this to the requirements file totally screw up how our dependencies run?  Please reject this PR it this is a bad idea...
https://twitter.com/choldgraf/status/929846273825562627

## Checklist

### DIFF
--- a/requirements-doc.txt
+++ b/requirements-doc.txt
@@ -1,2 +1,3 @@
+allensdk
 sphinx
 sphinx_rtd_theme


jcfr:
For notebook, it is usually recommended to install the package in the first cell, that way the notebook is independent and it can distributed as is.

In this case it would be:

```
! pip install pynwb
! pip install allensdk
```

nicain:
OK, got it.  Ill add that change in when addressing https://github.com/NeurodataWithoutBorders/pynwb/issues/253

choldgraf:
FWIW, adding a dependency install in a notebook is not very good practice IMO. It blurs the line between the computational environment needed to run code, and the code itself. There's a reason these things are separated out into requirements files! 

E.g., what happens if a dependency changes? You'd need to go into each notebook and change the first cell. Also, if people run into errors when pip-installing allensdk (as I did), it'll be even more confusing if the output is in a cell at the top of a jupyter notebook. This is especially true if the SDK depends on non-standard things, like matplotlib < 2.0.

Just my 2 cents!

nicain:
Thanks @choldgraf, good points. I’ll turn that cell into a markdown and call in installation instructions, but have it not executed, so I can have it both ways. @jcfr does this sound like a good compromise?

choldgraf:
if you wanna chat about ways to present the notebooks etc I'm happy to come by the booth and chat @ sfn!

jcfr:
> Thanks @choldgraf, good points. I’ll turn that cell into a markdown and call in installation instructions, but > have it not executed, so I can have it both ways. @jcfr does this sound like a good compromise?

Yes, that would be great.


Hi @choldgraf , 

Thanks for your input. 

Makes sense. We shouldn't explicitly install dependencies in the notebook. 

Regarding that specific PR, the requirements should not be added to `requirements-doc.txt`, the notebook are not part of the documentation uploaded on readthedocs.


### State of the art

Looking at the [A-gallery-of-interesting-Jupyter-Notebooks](https://github.com/jupyter/jupyter/wiki/A-gallery-of-interesting-Jupyter-Notebooks) available on the Jupyter wiki, there are two patterns:
* single notebook
* collection of notebooks

From I can observe, (and as you pointed out), dependencies are note installed in the notebook.

Here are some of the approach to help the user:

* it assumes user has the environment setup with the dependencies, it can also provides hint regarding the version used in the notebook . For example, see [here](http://nbviewer.jupyter.org/github/changhiskhan/talks/blob/master/pydata2012/pandas_timeseries.ipynb)

* it includes a  `0 - Getting Started.pynb` or `0 - Setup.ipynb` notebook explaining how to setup the environment, what to install and also doing some  version check. For example, 
  * https://github.com/InsightSoftwareConsortium/SimpleITK-Notebooks/blob/master/Python/00_Setup.ipynb
  * https://github.com/jakevdp/PythonDataScienceHandbook/blob/master/notebooks/00.00-Preface.ipynb


Also in case the notebook are in a dedicated repo, there is its own`requirements.txt`:
* https://github.com/jakevdp/PythonDataScienceHandbook

It also possible to have maintenance script to automate the update of the notebooks:
* https://github.com/jakevdp/PythonDataScienceHandbook/tree/master/tools


### Should we have a dedicated repo ? 

My understanding is that the pynwb project is useful to store the source code of pynwb python package and associated tests, and also manage associated script and tooling to drive its continuous integration and documentation generation.

I then wonder if we should instead have a repository named:

`nwb-notebooks `  or `pynwb-notebooks` ? 

where we would document, maintain and test these notebooks.


### What about a requirements-notebooks.txt ? 

> There's a reason these things are separated out into requirements files!

If we keep the notebooks in the pynwb repository, I think we could have a `requirements-notebooks.txt ` file.

### Notebooks and Continuous integration

>  what happens if a dependency changes? 
Generally speaking, the continuous integration testing would execute each notebook and report a error.


> You'd need to go into each notebook and change the first cell.
if we would end up specifying notebook dependencies that way, yes.

Note that dependencies can be "soft" and do not have to be pinned. In that case, it would be a convenience for the user of the notebook.

> , if people run into errors when pip-installing allensdk (as I did), it'll be even more confusing if the output is 
> in a cell at the top of a jupyter notebook. This is especially true if the SDK depends on non-standard 
> things, like matplotlib < 2.0.

Ideally, notebooks should be tested.

choldgraf:
These are all great points - here's a quick thought: have you guys looked into `sphinx-gallery` for generating documentation? It lets you embed rST into python files, then it'll render them into a gallery and create a jupyter notebook for the output content. That way you can run the examples / tutorials with your continuous integration, have rendered HTML of your analyses online, and have notebooks that people can download. E.g., [matplotlib ](http://matplotlib.org/gallery/index.html) and [scikit-learn](http://scikit-learn.org/stable/auto_examples/index.html) both use sphinx-gallery for their documentation.

jcfr:
> have you guys looked into sphinx-gallery for generating documentation?

Not yet. That is definitively something to consider.

We have a team meeting coming up, and I will make sure to bring that up.

Cc: @mgrauer

choldgraf:
Let me know if I can be helpful in any of those conversations or in setting something up!

oruebel:
@jcfr @nicain @choldgraf I generally agree with the points that have been raised here and these are things that @ajtritt and I have been discussing as well. From my point-of-view the main reason that this issue has not been addressed yet is simply that other items were more critical to make the release happen. I'll make sure that we discuss this topic during our next developer meeting.

PR #257, opened 2017-11-13
Add requirements for the notebooks

jcfr:
See #252


jcfr:
@nicain @choldgraf  Waiting we improve the testing/generation/presentation of notebooks, this should help user understand the requirements to run notebooks.

neuromusic:
related to strategies for notebook presentation, here's how seaborn does it:

these notebooks get built into "tutorials"

https://github.com/mwaskom/seaborn/tree/master/doc/tutorial

each notebook has a "raw" cell at the top with 

```
.. _aesthetics_tutorial:

.. currentmodule:: seaborn
```

so they can later be referenced with sphinx syntax.

This is the script that gets called by make to execute the notebooks, convert them to rst, then clear their outputs. These are just subprocess calls that could get baked directly into make or travis or whatever pynwb is using to build the documentation.

https://github.com/mwaskom/seaborn/blob/master/doc/tutorial/tools/nb_to_doc.py

choldgraf:
Another option worth looking into is sphinx-gallery. This would let you write your example files as python scripts that'd get built with the rest of your code (e.g., so that you know they run the way you'd expect). You can embed rST into the files, automatically generate a gallery page from the script outputs, and it'll also create jupyter notebooks that users can download (and there will soon be an auto-binder link feature).

IME I still haven't seen a good documentation-via-notebook solution, but I've seen sphinx gallery work pretty well (e.g., matplotlib and scikit-learn both use it)

ajtritt:
Do we need this now that we have #411?

oruebel:
I think the answer is that the two notebooks we have, i.e., (1) the MeisterLab example and (2) the Allen Legacy example should be converted to Sphinx Gallery. (1) should be an issue for me and (2) I think would be an issue for @nicain . I'll create the issues so that we could then remove the notebooks folder and just do Sphinx Galleries. 

As it stands, at least the MeisterLab example notebook can't be easily tested because it (a) requires the download of data from CRCNS and (b) runs fairly long. I think it would be Ok to include it in the separate, long-running tests that we've talked about before but it should not be part of the testing that is done with every pull request.

oruebel:
Created #452 and #453 to port the notebooks to Sphinx Gallery. @ajtritt  how does it work with rendering long-running scripts as part of docs? Can they have be re-rendered every time or can we just run them offline and update them as needed?

t-b:
How about we merge this PR until we moved everything to sphinx-gallery?

rly:
This is incorporated in `requirements-doc.txt`

PR #264, opened 2017-11-15
Test doc example

jcfr:
## Motivation

Add automatic testing for examples to ensure users will not face errors when reading the documentation.

## How to test the behavior?

Try to reproduce the examples found in the documentation or run the example tests using `./test.py --example`.

### DIFF
--- a/docs/code/creating-and-writing-nwbfile-2.py
+++ b/docs/code/creating-and-writing-nwbfile-2.py
@@ -1,27 +1,31 @@
 
 def main():
+    import os.path
+
     # example: start
     from datetime import datetime
 
-    from pynwb import NWBFile, TimeSeries
+    from pynwb import NWBFile, TimeSeries, get_manager
     from pynwb.form.backends.hdf5 import HDF5IO
 
     start_time = datetime(1970, 1, 1, 12, 0, 0)
     create_date = datetime(2017, 4, 15, 12, 0, 0)
 
-    nwbfile = NWBFile('the PyNWB tutorial', 'test.nwb', 'a test NWB File', 'TEST123', start_time,
+    nwbfile = NWBFile('the PyNWB tutorial', 'a test NWB File', 'TEST123', start_time,
                       file_create_date=create_date)
 
     ts = TimeSeries('test_timeseries', 'example_source', list(range(100, 200, 10)), 'SIunit',
                     timestamps=list(range(10)),
                     resolution=0.1)
 
-    nwbfile.add_raw_timeseries(ts)
+    nwbfile.add_acquisition(ts)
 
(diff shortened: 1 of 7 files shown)


docs/code/creating-and-writing-nwbfile-2.py
```diff
@@ -0,0 +1,32 @@
+
+def main():
+    import os.path
+
+    # example: start
+    from datetime import datetime
+
+    from pynwb import NWBFile, TimeSeries, get_manager
+    from pynwb.form.backends.hdf5 import HDF5IO
```
dorukozturk, line 9: nitpick: Should the imports be at the top of the file?

jcfr: Since we want have them showing up in the example, they have been purposely added in the middle .. just before each comment of the form `# example: start` / `# example: end` that are used in the sphinx directive `litteralinclude`.

test.py
```diff
@@ -1,70 +1,59 @@
...
-# Run integration tests
-if flags['integration'] in args.suites:
-    logging.info('running integration tests')
-    integration_test_result = unittest.TextTestRunner(verbosity=args.verbosity).run(
-        unittest.TestLoader().discover("tests/integration"))
-    TOTAL += integration_test_result.testsRun
-    FAILURES += len(integration_test_result.failures)
-    ERRORS += len(integration_test_result.errors)
 
+def _import_from_file(script):
```
dorukozturk, line 28: I think it is better to:
1) Change the test file names so that they are importable
2) Moving creating-and-writing-nwbfile-2.py  and creating-and-writing-nwbfile.py under tests/integration
3) Importing them when needed rather than "load_source"ing them.
4) Adding some assertions as opposed to just creating files and removing them? It might be better to make sure we check the integrity of the files(without testing the h5py library).

jcfr: Good point. I just created an issue to capture these: https://github.com/NeurodataWithoutBorders/pynwb/issues/270

In the mean time, I suggest we integrate so that we have:
* a "not prefect" but working baseline for improving example testing
* an example documentation that is not broken

Also, since we may move toward ipython notebook for example and tutorials, I would prefer not to spend too much time yet.

Let me know what you think ?

dorukozturk: Great, I think this is perfectly fine. We can work on #270 later. Otherwise it LGTM.

dorukozturk (dismissed):

jcfr:
@dorukozturk  Thanks for the review. Topic rebased, I will integrate as soon CI returns green

jcfr:
Integrating. AppVeyor and CircleCI are green, and TravisCI for python 3.6.3 is also green.

PR #271, opened 2017-11-15
Add usedevelop flag for coverage-tox issue

### LINKED ISSUE #269: coverage: Investigate why coverage is reported to be 0%
### Steps to Reproduce

See See https://codecov.io/gh/NeurodataWithoutBorders/pynwb

dorukozturk:
## Motivation

Fixes #269 

Coverage was reporting 0% even if the coverage was ran. Running tox environment in development mode and correctly referencing pynwb path fixed this issue. 

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,15 +36,25 @@ ci-steps: &ci-steps
         command: |
           . ../venv/bin/activate
           pip install tox
-          tox -e $TOX_ENV
+          tox -e $TEST_TOX_ENV
         # Install is expected to be quick. Increase timeout in case there are some network issues.
         # While pip installing tox does not output by default. Circle thinks task is dead after 10 min.
         no_output_timeout: 30m
     - run:
+        name: Run coverage and submit to codecov-io
+        command: |
+          . ../venv/bin/activate
+          tox -e $COVERAGE_TOX_ENV
+    - run:
+        name: Build wheel and source distribution
+        command: |
+          . ../venv/bin/activate
+          tox -e $BUILD_TOX_ENV
+    - run:
         name: Test installation from a wheel
         command: |
           . ../venv/bin/activate
-          tox -e $TOX_ENV recreate --installpkg dist/*-none-any.whl
+          tox -e $TEST_TOX_ENV recreate --installpkg dist/*-none-any.whl
     - run:
         <<: *publish-prerelease-on-github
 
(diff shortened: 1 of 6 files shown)


jcfr:
Hi,

Thanks. Look like the _version.py script wasn't we excluded from the
report. Is the branch on the latest dev ?

May the config setting is ignored ?

(Can't easily check, on my phone)

Jc

On Nov 15, 2017 3:59 PM, "Doruk Ozturk" <notifications@github.com> wrote:

@dorukozturk <https://github.com/dorukozturk> requested your review on:
NeurodataWithoutBorders/pynwb#271
<https://github.com/NeurodataWithoutBorders/pynwb/pull/271> Add usedevelop
flag for coverage-tox issue.

—
You are receiving this because your review was requested.

Reply to this email directly, view it on GitHub
<https://github.com/NeurodataWithoutBorders/pynwb/pull/271#event-1343865534>,
or mute the thread

jcfr:
Not sure why is has to be done in developer mode.

Here is what I used to do to make sure all codecov variables are passed: https://github.com/scikit-build/cmake-python-distributions/blob/a6ed3f1c89a88c3968792d167313716dbb96b33b/.circleci-matrix.yml#L17

May be related is #272

.circleci/config.yml
```diff
@@ -63,21 +63,24 @@ jobs:
...
+     - COVERAGE_TOX_ENV: "py27"
     <<: *ci-steps
 
   python35:
     docker:
       - image: circleci/python:3.5.4-jessie
     environment:
-     - TOX_ENV: "py35,CI-py35"
+     - TEST_TOX_ENV: "py35,CI-py35"
+     - COVERAGE_TOX_ENV: "py35"
```
jcfr: Are the variable switched ? Look like `COVERAGE_TOX_ENV` should be associated with `CI-py35`

jcfr: Also not clear why it is `py35,CI-py35` instead of `CI-py35`

dorukozturk: I think you are right. I think all envs inherit from testenv so CI-py35 should be enough.

dorukozturk: I think I will refactor tox.ini. We should not need comments there.

dorukozturk:
@jcfr I think this is ready. I fixed all the coverage related issues and made appveyor also run tox for consistency. Also fixed the running coverage twice issue and added some comments too.

appveyor.yml
```diff
@@ -9,21 +9,44 @@ environment:
     - PYTHON_VERSION: 2.7
       PYTHON: C:\Miniconda-x64
       ENV: miniconda
+      TEST_TOX_ENV: py27
+      COVERAGE_TOX_ENV: coverage-py27
+      BUILD_TOX_ENV: build-py27
```
jcfr, line 14: How do you ensure that tox will run into the miniconda env ? 

References:
* https://stackoverflow.com/questions/30555943/is-it-possible-to-use-tox-with-conda-based-python-installations
* https://github.com/tox-dev/tox/tree/master/doc/drafts and also `https://github.com/tox-dev/tox/issues/273`

jcfr:
Nice, much cleaner. Also nice to rename the tox env.

The issue with this approach is that tox is called  multiple times (once for build, once for test wheel install, once for coverage)

On windows, this kills the CI speed (which was not already great). Now at 20 minutes

I still think have one tox env with conditional execution of command would be the way to go.

jcfr:
> The issue with this approach is that tox is called multiple times (once for build, once for test wheel install, 
> once for coverage)

> On windows, this kills the CI speed (which was not already great). Now at 20 minutes

> I still think have one tox env with conditional execution of command would be the way to go.

Having an other look. I suggest we move forward with this and improve in follow up PR.

We may have to revisit the miniconda testing

dorukozturk:
@jcfr I agree. I can work on conda-tox stuff on another PR.

jcfr (approved):

PR #276, opened 2017-11-17
Update .gitignore to ignore vscode launch.json

nicain:
## Motivation

@nicain is using vscode, and has a custom launch.json.  This way his config is not forced on anyone else
## Checklist


jcfr:
👍 

For future reference, the file is usually locally modified but the corresponding changes are specific and it would not make sens to generalize them.

Issue #267, opened 2017-11-15
docs: Fix ReadTheDocs so that it includes API documentation


jcfr:
### Steps to Reproduce

See https://pynwb.readthedocs.io/en/latest/api_docs.html

### Suggested changes

Two approaches:

* ensure `make apidocs` is called on the readthedocs worked

* commit generated API rst files  (look if we could be added to and include from `docs/source/api`)

nicain:
I am trying to implement some backwards compatibility for a different type of file (xref https://github.com/NeurodataWithoutBorders/pynwb/issues/275) and having the module documentation is really helpful

PR #279, opened 2017-11-17
doc: Add module API documentation. See #267

jcfr:
## Motivation

API documentation missing from readthedocs. See #267 

Look like an alternative would be to write our builder on readthedocs, not sure if it is worth the effort. I think it is reasonable to expect developer to locally generated the documentation and checked in the update.

See https://docs.readthedocs.io/en/latest/builds.html#understanding-what-s-going-on

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,10 +9,6 @@
 
 docs/notebooks/boc/
 
-# Auto-generated apidocs RST files
-docs/source/pynwb*.rst
-docs/source/modules.rst
-
 ### Python ###
 # Byte-compiled / optimized / DLL files
 __pycache__/
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -158,6 +158,6 @@ doctest:
 	      "results in $(BUILDDIR)/doctest/output.txt."
 
 apidoc:
-	$(SPHINXAPIDOC) -f -e -o $(RSTDIR)  $(SRCDIR)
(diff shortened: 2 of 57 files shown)


oruebel:
@jcfr the apidoc should be run on read-the-docs directly. I believe that is how @ajtritt had set the PyNWB docs up originally and that is also how the schema-docs work. If the docs are not being generated then that should be a bug in ``docs/source/conf.py`` I can take a look later.

nicain (approved):

oruebel:
@jcfr @nicain looking at the  ``conf.py`` it already has the ``run_apidoc(_)`` function to generate the apidocs as part of the build. I'll check why it is not being run.

oruebel:
@jcfr @nicain @dorukozturk Ok, I found the problem. This issue is that in https://github.com/NeurodataWithoutBorders/pynwb/commit/39c9a711cefeaef84e76d4adcfaa754b84155f09  @dorukozturk added a ``setup`` function at the end of conf.py but the file already contained a ``setup`` function (line 124) to do the autogenerate of the souces and the new function simply overwrote the previous setup.py function. I'll fix the bug and submit a pull request soon. @jcfr as far as I can tell, this pull request only adds the autogenerated sphinx sources. As such I suggest we do not merge this pull request and I'll add the fix in the upcoming request.

oruebel:
@jcfr @nicain @dorukozturk @ajtritt I just added #280 which should fix this problem.

nicain:
Closed because #280 has resolved the original concern.

Issue #309, opened 2017-12-20
Docs about updating or adding data


matham:
I'm looking at pynwb and how I can use it in our pipeline, but it's not exactly clear to me some of the data manipulation aspects as a user:

* Can I make use of the fact that we're dealing with hdf5 files and directly manipulate the file, or do I always have to go through the pywnb wrapper.
* Specifically, the docs show how to create a file and add data, but how do I stream live data to the file? E.g. say I create a TimeSeries representing ephys data, how can keep appending data (efficiently) to the TimeSeries? E.g. h5py lets you do it with http://docs.h5py.org/en/latest/faq.html#appending-data-to-a-dataset.
* Say I have the file and then call `io.write(f)`. Then I add more data and call it again. Would it re-write the whole file again? What about huge files? I guess it isn't clear to me how thin the pynwb wrapper is on top of h5py or if it's completely seperate.
* It doesn't let me e.g. change the description of `ElectricalSeries` etc. once created. In a typical experiment I'd creates a bunch of data, write in a notepad the conditions for each run and then later fill in all the info to the file. But if it's write once that wouldn't work. I suppose I can create the data live and then later import it to a new file and add all this info then. But that may create data duplication, which is probably unwanted. Although I suppose it may be best to not edit original data at all once created, but rather link to it.

Thanks in advance!

ajtritt:
In the order that you asked your questions:

- Yes. PyNWB is just writing HDF5 files... there is nothing about these files that prevents you from opening them up with h5py or any other HDF5 bindings.
- There is some support for writing streams. Unfortunately the documentation is a bit lacking. The key data structure is the [DataChunkIterator](http://pynwb.readthedocs.io/en/latest/pynwb.form.data_utils.html). You would wrap up your stream object with a DataChunkIterator and pass it in place of a list/ndarray/DataFrame/etc The stream will not be read from until `io.write(f)` is called on the parent file. Also, these streams will be pulled from in serial. As a disclaimer, this is an advanced feature that we haven't had the chance to test in the real world outside of converting datasets that are too large to fit into memory. I assume in your case you would want to write data from multiple streams concurrently?
 We would appreciate any input you have on the is. @oruebel may also have more to say
[...]

PR #310, opened 2017-12-22
Proof of concept for #309

matham:
This implements a streaming data iterator as described in https://github.com/NeurodataWithoutBorders/pynwb/issues/309. I think it fully implements it, although it still has a few rough edges and I didn't run `flake8` on it. Just want to get some feedback first.

The name `plumb_data` is random and can be changed if you like.

You can test with
```py
import numpy as np
from datetime import datetime
from pynwb import NWBFile
from pynwb import get_manager
from pynwb.form.backends.hdf5 import HDF5IO
from pynwb.form.build import Builder, BuildManager
from pynwb.form.data_utils import DataChunkIterator, DataChunkStream
from pynwb.ecephys import ElectricalSeries


rate = 10.0
np.random.seed(1234)
data_len = 1000
ephys_data = np.random.rand(data_len)
ephys_timestamps = np.arange(data_len) / rate
spatial_timestamps = ephys_timestamps[::10]
spatial_data = np.cumsum(np.random.normal(size=(2, len(spatial_timestamps))), axis=-1).T

f = NWBFile('the PyNWB tutorial', 'my first synthetic recording', 'EXAMPLE_ID', datetime.now(),
            experimenter='Dr. Bilbo Baggins',
            lab='Bag End Laboratory',
            institution='University of Middle Earth at the Shire',
            experiment_description='I went on an adventure with thirteen dwarves to reclaim vast treasures.',
            session_id='LONELYMTN')

filename = r"G:\Python\example.h5"
io = HDF5IO(filename, manager=get_manager(), mode='w')

epoch_tags = ('example_epoch',)
ep1 = f.create_epoch(source='an hypothetical source', name='epoch1', start=0.0, stop=1.0,
                     tags=epoch_tags,
                     description="the first test epoch")
ep2 = f.create_epoch(source='an hypothetical source', name='epoch2', start=0.0, stop=1.0,
                     tags=epoch_tags,
                     description="the second test epoch")
device = f.create_device(name='trodes_rig123', source="a source")
electrode_name = 'tetrode1'
source = "an hypothetical source"
description = "an example tetrode"
location = "somewhere in the hippocampus"
electrode_group = f.create_electrode_group(electrode_name,
                                           source=source,
                                           description=description,
                                           location=location,
                                           device=device)
for idx in [1, 2, 3, 4]:
    f.add_electrode(idx,
                    x=1.0, y=2.0, z=3.0,
                    imp=float(-idx),
                    location='CA1', filtering='none',
                    description='channel %s' % idx, group=electrode_group)
electrode_table_region = f.create_electrode_table_region([0, 2], 'the first and third electrodes')

stream = DataChunkStream(min_buffer_size=3000, dtype=ephys_data.dtype)
ephys_ts = ElectricalSeries('test_ephys_data',
                            'an hypothetical source',
                            stream,
                            electrode_table_region,
                            timestamps=ephys_timestamps,
                            # Alternatively, could specify starting_time and rate as follows
                            # starting_time=ephys_timestamps[0],
                            # rate=rate,
                            resolution=0.001,
                            comments="This data was randomly generated with numpy, using 1234 as the seed",
                            description="Random numbers generated with numpy.random.rand")
f.add_acquisition(ephys_ts, [ep1, ep2])

spatial_ts = SpatialSeries('test_spatial_timeseries',
                           'a stumbling rat',
                           DataChunkIterator(spatial_data),
                           'origin on x,y-plane',
                           timestamps=spatial_timestamps,
                           resolution=0.1,
                           comments="This data was generated with numpy, using 1234 as the seed",
                           description="This 2D Brownian process generated with "
                                       "np.cumsum(np.random.normal(size=(2, len(spatial_timestamps))), axis=-1).T")
f.add_acquisition(spatial_ts, [ep1, ep2])

io.write(f)


def read():
    io2 = HDF5IO(filename, get_manager(), mode='r')
    f2 = io2.read()
    d = f2.get_acquisition('test_ephys_data')
    print(d.num_samples, d.data.shape)
    io2.close()


read()
for _ in range(10):
    stream.append_data(ephys_data)
    io.plumb_data(f)
    read()

stream.flush_data()
io.plumb_data(f)
read()
io.close()
```
Notice that it buffers with 3000. When run this prints

```
0 (0,)
0 (0,)
0 (0,)
3000 (3000,)
3000 (3000,)
3000 (3000,)
6000 (6000,)
6000 (6000,)
6000 (6000,)
9000 (9000,)
9000 (9000,)
10000 (10000,)
```

### DIFF
--- a/src/pynwb/base.py
+++ b/src/pynwb/base.py
@@ -1,7 +1,7 @@
 from collections import Iterable
 
 from .form.utils import docval, getargs, popargs, fmt_docval_args
-from .form.data_utils import DataChunkIterator
+from .form.data_utils import AbstractDataChunkIterator
 
 from . import register_class, CORE_NAMESPACE
 from .core import NWBContainer
@@ -90,7 +90,7 @@ class TimeSeries(NWBContainer):
              'doc': ('Name of TimeSeries or Modules that serve as the source for the data '
                      'contained here. It can also be the name of a device, for stimulus or '
                      'acquisition data')},
-            {'name': 'data', 'type': (Iterable, 'TimeSeries', DataChunkIterator),
+            {'name': 'data', 'type': (Iterable, 'TimeSeries', AbstractDataChunkIterator),
              'doc': 'The data this TimeSeries dataset stores. Can also store binary data e.g. image frames'},
             {'name': 'unit', 'type': str, 'doc': 'The base unit of measurement (should be SI unit)'},
             {'name': 'resolution', 'type': (str, float),
@@ -102,7 +102,7 @@ class TimeSeries(NWBContainer):
              'default': _default_conversion},
 
             # time related data is optional, but one is required -- this will have to be enforced in the constructor
-            {'name': 'timestamps', 'type': (Iterable, 'TimeSeries', DataChunkIterator),
+            {'name': 'timestamps', 'type': (Iterable, 'TimeSeries', AbstractDataChunkIterator),
              'doc': 'Timestamps for samples stored in data', 'default': None},
             {'name': 'starting_time', 'type': float, 'doc': 'The timestamp of the first sample', 'default': None},
             {'name': 'rate', 'type': float, 'doc': 'Sampling rate in Hz', 'default': None},
@@ -139,7 +139,7 @@ class TimeSeries(NWBContainer):
         if isinstance(data, TimeSeries):
             data.fields['data_link'].append(self)
             self.fields['num_samples'] = data.num_samples
-        elif isinstance(data, DataChunkIterator):
+        elif isinstance(data, AbstractDataChunkIterator):
             self.fields['num_samples'] = -1
         else:
             self.fields['num_samples'] = len(data)
--- a/src/pynwb/ecephys.py
+++ b/src/pynwb/ecephys.py
@@ -3,7 +3,7 @@ from h5py import RegionReference
 from collections import Iterable
 
 from .form.utils import docval, getargs, popargs, call_docval_func
-from .form.data_utils import DataChunkIterator, ShapeValidator
+from .form.data_utils import AbstractDataChunkIterator, ShapeValidator
 
 from . import register_class, CORE_NAMESPACE
 from .base import TimeSeries, _default_resolution, _default_conversion
@@ -115,7 +115,7 @@ class ElectricalSeries(TimeSeries):
              'doc': ('Name of TimeSeries or Modules that serve as the source for the data '
                      'contained here. It can also be the name of a device, for stimulus or '
                      'acquisition data')},
-            {'name': 'data', 'type': (list, np.ndarray, DataChunkIterator, TimeSeries, Iterable),
+            {'name': 'data', 'type': (list, np.ndarray, AbstractDataChunkIterator, TimeSeries, Iterable),
              'doc': 'The data this TimeSeries dataset stores. Can also store binary data e.g. image frames'},
 
             {'name': 'electrodes', 'type': ElectrodeTableRegion,
@@ -126,7 +126,7 @@ class ElectricalSeries(TimeSeries):
             {'name': 'conversion', 'type': float,
              'doc': 'Scalar to multiply each element by to conver to volts', 'default': _default_conversion},
 
-            {'name': 'timestamps', 'type': (list, np.ndarray, DataChunkIterator, TimeSeries, Iterable),
+            {'name': 'timestamps', 'type': (list, np.ndarray, AbstractDataChunkIterator, TimeSeries, Iterable),
              'doc': 'Timestamps for samples stored in data', 'default': None},
             {'name': 'starting_time', 'type': float, 'doc': 'The timestamp of the first sample', 'default': None},
             {'name': 'rate', 'type': float, 'doc': 'Sampling rate in Hz', 'default': None},
@@ -169,9 +169,9 @@ class SpikeEventSeries(ElectricalSeries):
              'doc': ('Name of TimeSeries or Modules that serve as the source for the data '
                      'contained here. It can also be the name of a device, for stimulus or '
                      'acquisition data')},
-            {'name': 'data', 'type': (list, np.ndarray, DataChunkIterator, TimeSeries, Iterable),
+            {'name': 'data', 'type': (list, np.ndarray, AbstractDataChunkIterator, TimeSeries, Iterable),
              'doc': 'The data this TimeSeries dataset stores. Can also store binary data e.g. image frames'},
-            {'name': 'timestamps', 'type': (list, np.ndarray, DataChunkIterator, TimeSeries, Iterable),
+            {'name': 'timestamps', 'type': (list, np.ndarray, AbstractDataChunkIterator, TimeSeries, Iterable),
              'doc': 'Timestamps for samples stored in data'},
             {'name': 'electrodes', 'type': ElectrodeTableRegion,
              'doc': 'the table region corresponding to the electrodes from which this series was recorded'},
@@ -194,11 +194,11 @@ class SpikeEventSeries(ElectricalSeries):
         name, source, data, electrodes = popargs('name', 'source', 'data', 'electrodes', kwargs)
         timestamps = getargs('timestamps', kwargs)
         if not (isinstance(data, TimeSeries) and isinstance(timestamps, TimeSeries)):
-            if not (isinstance(data, DataChunkIterator) and isinstance(timestamps, DataChunkIterator)):
+            if not (isinstance(data, AbstractDataChunkIterator) and isinstance(timestamps, AbstractDataChunkIterator)):
                 if len(data) != len(timestamps):
                     raise Exception('Must provide the same number of timestamps and spike events')
             else:
-                # TODO: add check when we have DataChunkIterators
+                # TODO: add check when we have AbstractDataChunkIterators
                 pass
         super(SpikeEventSeries, self).__init__(name, source, data, electrodes, **kwargs)
 
@@ -400,11 +400,11 @@ class FeatureExtraction(NWBContainer):
     @docval({'name': 'source', 'type': str, 'doc': 'The source of the data'},
             {'name': 'electrodes', 'type': ElectrodeTableRegion,
              'doc': 'the table region corresponding to the electrodes from which this series was recorded'},
-            {'name': 'description', 'type': (list, tuple, np.ndarray, DataChunkIterator),
+            {'name': 'description', 'type': (list, tuple, np.ndarray, AbstractDataChunkIterator),
              'doc': 'A description for each feature extracted', 'ndim': 1},
-            {'name': 'times', 'type': (list, tuple, np.ndarray, DataChunkIterator),
+            {'name': 'times', 'type': (list, tuple, np.ndarray, AbstractDataChunkIterator),
              'doc': 'The times of events that features correspond to', 'ndim': 1},
-            {'name': 'features', 'type': (list, tuple, np.ndarray, DataChunkIterator),
+            {'name': 'features', 'type': (list, tuple, np.ndarray, AbstractDataChunkIterator),
              'doc': 'Features for each channel', 'ndim': 3},
             {'name': 'name', 'type': str, 'doc': 'the name of this container', 'default': 'FeatureExtraction'})
     def __init__(self, **kwargs):
--- a/src/pynwb/form/backends/hdf5/h5tools.py
+++ b/src/pynwb/form/backends/hdf5/h5tools.py
@@ -8,7 +8,7 @@ import warnings
 from ...container import Container
 
 from ...utils import docval, getargs, popargs, call_docval_func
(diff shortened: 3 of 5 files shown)


ajtritt:
Can you explain why you added the additional class DataChunkStream? Was it to have an object to append to? You should use DataChunkIterator by passing in a stream (i.e. Iterable) to the data argument. For example, you could create a class that reads data from an input stream, formats it, and returns it as a DataChunk when __next__ is called.

matham:
Yeah, it was partially to be able to append to because the explicit append was simpler for the typical case where I just need to keep adding data.

But the main reason for making a new class rather than passing my data as a iterator argument to DataChunkIterator is that after one of reading of all the chucks in the first write, that internal data iterator is used up (as a `StopIteration` would have been raised) so at the next write, additional calls would not be able to return new data. I'd need to create a new iterator for my data for every write, which I'd need to somehow pass in to DataChunkIterator. I could overwrite some of the methods of DataChunkIterator and pass it these new iterable, but it's just simpler to do it with a new class using a queue. In addition, there are a few significant issues with DataChunkIterator that I explained at https://github.com/NeurodataWithoutBorders/pynwb/issues/309#issuecomment-353509601.

`DataChunkStream` doesn't need to exist in pynwb, as it could be code that I just have in my project, but I do think this would probably be simpler to use for the typical case as it doesn't require iterables, just data appends as it comes in, and should also be more performant. But yeah, `DataChunkStream` could be left out from the PR.

matham:
Hey I'm wondering if you think the basic idea is good? I can remove the part you mentioned, but otherwise, would this be a good approach of (1) checking for `AbstractDataChunkIterator` everywhere rather than the subclass so that we can more easily implement our own iterators, and (2) the `plumb_data` approach?

oruebel:
@matham using ``AbstractDataChunkIterator`` in place of ``DataChunkIterator`` makes sense to me. ``AbstractDataChunkIterator`` was added as base class a bit later (i.e., after ``DataChunkIterator``) so that is probably an artifact from that. The only thing we need to be careful with is to make sure that, e.g., ``__chunked_iter_fill__`` does not inadvertently use any functions from ``DataChunkIterator`` that are missing on the abstract base class, although I think that shouldn't be the case. 

I'm not sure I fully understand the ``plumb`` data part. I think it's probably the name that's confusing me a bit, what does the term ``plumb`` refer to here? It would be nice to have a bit more comments on the functions to make the purpose of each function a bit more explicit in the code. Also, it looks like the code you added is not covered by any tests right now.

matham:
Ok, I'll fix things up and more (code) explanations and tests. The name `plumb` was just temporary, maybe you have better name suggestion?.

The basic conceptual idea is that when you call `io.write`, it writes all the data for normal data source. For iterator data sources it also writes all the data currently accumulated in the iterator. This is how it currently works and I didn't change it. Basically in `io.write` it does something like:

```py
for channel in channels:
    for data in channel.data:
        do something
```
The problem with this is that when you call `io.write`, it iterates through the `channel.data` iterator until it's exhausted and the function returns, at which point the iterator is dead and further calls to it will just keep raising `StopIteration` and not return new data. The problem is that I would like to keep adding data to the iterator as it's being generated and then and keep calling `write` say every 5 minutes to save, which is impossible once the iterator is exhausted. I also don't want to block in `write` of course to wait for the new data.

Instead, I created my `DataChunkStream` class, where if you look, it's `__iter__` method creates a new generator/iterator each time it's called. That means we can call the above code `for data in channel.data` many times until the current iterator it's exhausted and it raises a `StopIteration`, meaning all the data accumulated until now has been written. Now, as new data is generated and added to `DataChunkStream`, we'll call `plumb_data` each time, and each time it does `for data in channel.data` it'll generate a new iterator which will iterate over the existing data until done.

So basically, I can call `io.write` once to initialize everything and write the not iterable data. Then, I can call `plumb_data` periodically, which goes over all the channels that have iterator data and write them. This would only work of course for iterators which generate a new iterator every time it's called, hence why I added the `DataChunkStream` class. I realize I may not have been very clear about this, and maybe that's what @ajtritt was asking? Do you still think I should remove it? This kind of iterable class can be generated by users, but I included it because I figured it'd be more convenient as not all users would fully understand how to do that. It's also pretty easy to use as you can just append to the queue.

Does this make more sense?

oruebel:
@matham thanks for the explanation, code, and for being open to discussions. I think I understand now better what you are doing. I agree that the fact that currently ``DataChunkIterators`` are processed sequentially and until completion is a limitation for what you are trying to do. The problem you are trying to solve is an important one but also a tricky one, so please allow me to dig a bit deeper with you into this. Let me ask a couple more questions to see if we can iterate to a solution:

* Why does ``write_plumb_data`` need to generate the list of “unfinished” iterators (i.e., ``datas``) each time you call the function? In the context of data streaming where you would call ``plumb_data`` many times it seems like constructing ``datas`` each time may lead to quite a bit of overhead. I’m wondering, wouldn’t it make sense to break ``write_plumb_data`` into two functions, one that constructs the list of data streams and one that writes the data streams. In this way you could reuse the list of data streams rather than having to reconstruct it every time.
* From a user perspective, I think one concern is that in the current example a user needs to change how they call the backend to do the write, i.e., instead of just ``write`` one needs to add a loop to call ``write_plumb_data`` until all data streams are exhausted. I think it would make sense to integrate this process more tightly with the API by:
   1.  adding a function that does the iteration over ``plumb_data``  
   2. extending HDF5IO.write to (optionally) process the data streams, 
   3. adding any parameters needed to HDF5IO to configure how the data streams should be handled (e.g., it could be configurable whether HDF5IO.write processes all streams to completion or whether the user will need to do this)
* How do you know right now when you need to stop calling ``plumb_data``, i.e., how do you know when all streams are exhausted, i.e., is there a function that does this?
* When do you think I/O should be flushed? What I mean is, HDF5 buffers I/O operations and if I understand the code correctly then I think the HDF5 file may end up in an invalid state after calling ``plumb_data`` and is only guaranteed to be valid once you close the file. Maybe adding an option to ``plumb_data`` to allow the user to flush I/O might be the way to go.

One thing that concerns me a bit about the current design is that it (for lack of a better word) “encourages” having multiple streams writing to the same HDF5 file. This seems potentially problematic for a couple of reasons: 
*  **write performance** the I/O performance of each stream becomes directly dependent on what the other streams are doing
*  **data corruption** if something goes wrong on one data stream it is possible that the whole file becomes corrupted and you lose the data from all streams
* **read performance** Lets say we have two streams ``A`` and ``B`` each consisting of a bunch of chunks ``ai`` and ``bi``. If both streams write to the same file then on disk you may end up with a highly fragmented file where chunks for ``A`` and ``B`` are intermixed, e.g,. ``a1, b1, a2, a3, b2, a4, b3`` . This kind of fragmentation can be problematic for the performance of data read. 

To address these problems, I think in particular in the context of data streaming it seems desirable to have the option to have each data stream go to a separate HDF5 file. To do this, we could simply have an option on the ``AbstractDataChunkIterator`` that would tell the HDF5IO backend to create a new HDF5 file to store the corresponding dataset (and attributes) and then that dataset would be linked to via an external link from the main file. This would a) avoid having to update the main file while you are streaming data, b) would make the streams independent of each other (as each could go to its on file and, hence, also allow you to process streams independently, c) help alleviate at least in part the issue of data corruption and performance. 

Once we have finalized how we want to deal with the data streaming I think we’ll also need to create some examples for this and ideally a tutorial with best practices, because this gives folks a lot of power but it can also lead to a lot of problem if one is not careful.

matham:
`proposal to separate write_plumb_data into write_iterator_data and get_data_iterators`

Yes, that's a good idea. I wasn't sure if there would be an issue with someone creating a new data group between `write_iterator_data` and `get_data_iterators` (if that's even allowed). But, if it's allowed we can just make it clear in the docs that creating a new group requires calling `get_data_iterators` again, otherwise it won't be written. Either way we can have `def write_iterator_data(iterators=None):...`, so that if the user doesn't provide the iterators and it's None, we do `iterators = get_data_iterators()`.

`Won't the user have to call plumb_data in a loop?`

I'm not sure if I missed something here, or if I wasn't clear about the queue and plumbing. The idea, and what actually happens in `write_plumb_data`, is that when a user calls it, all the data that has been enqueued up to that moment is written then. So the user should know, if there's data that has been enqueued, but they haven't called `write_iterator_data`, then the data isn't written. But once they call `write_iterator_data`, **all** the data is written. When implementing `DataChunkStream`, I tried to make absolutely sure that no data will be unwritten once `write_plumb_data` is called. So the user will not need to call it in a loop, only once whenever they want to write the currently unwritten data. Does this resolve the issue? I also didn't understand what you meant about updating the `HDF5IO` API.

`How do you know right now when you need to stop calling plumb_data`

The same as the last, you only need to call it once, and it writes all the currently unwritten data. So if you appended data to the `DataChunkStream` you know you need to call `write_iterator_data` and it'll write everything. Specifically, every time `write_iterator_data` is called, `__iter__` of the `DataChunkStream` is called and it creates a new generator that iterates over **all** the data appended up to that moment when the generator was created and it writes it all.

`When do you think I/O should be flushed?`

I'm not sure I understood this. I assumed that as we edit the hdf5 data, hfpy will automatically flush the data to disk? Is that not the case? Will it just remain in RAM if we don't flush it explicitly? If so, I agree we should have a flush parameter.

`About hdf5 data corruption and fragmentation issues`.

I'm not too familiar with these issues as I haven't stress used hdf5 yet, although I'd admit I always worried how I can make sure that new data doesn't corrupt the file. In past circumstances I handled it by creating a backup every few minutes. But I'm not sure why you're worried here about corruption anymore than a normal `write`, which also writes many channels and groups? Are you worried about causing an exception during streaming? About fragmentation, aren't we passing it on from HDF5 to the OS/disk? Does HDF5 not deal well with fragmentation compared to OS? I guess that may explain the slow loading time in some of my hdf5 data with e.g. 60 electrodes.

Either way your idea about optionally creating a new file for each stream sounds great, although I'm not sure how to do this myself.

I can definitely add examples and a tutorial.

matham:
There's one thing I had forgotten about. I added a buffer option to the `DataChunkIterator`, so that if the data accumulated in the iterator is less than the buffer, calling `write_iterator_data` would do nothing. Instead, you'd need to flush the iterator by calling `iterator.flush_data()` and then call `write_iterator_data`. 

This does seem a bit complex to the user, so instead I can add a `flush_data` method to the abstract iterator class and then add a flush parameter to `write_iterator_data`, which when true would first flush the iterator using `flush_data` and then write it.

oruebel:
@matham I have been chatting with @ajtritt about this and I think we need to take a careful look at designing the different pieces needed to implement the support for multiple data streams. The proof of concept you have created here is very helpful and I really appreciate all the effort you have put into putting this together! 

The problem you are trying to solve is quite challenging because there are so many different moving parts involved, from how we handle data in the API to how we write data and how HDF5 handles the data. Since the problem of multiple data streams is fundamental to a number of use cases, I really want us to get this right (or as close to right as possible). Let me suggest the following course of action:

- It would be helpful if you could outline the use-case you are trying to solve. You mentioned #309 that you want to integrate your data pipeline. How does this pipeline look? How many data streams are involved? How large are the data streams? Do you want to write the data to one or multiple files (and why)? etc. Having this information would help us better understand the requirements when we try to define a design to solve the problem. 
- I will create a couple of issues to try to break down the problem into more manageable chunks. I will reference pull request #310 and #309 issues there to make sure we don't lose track of our discussions. 
- @ajtritt and I will get together to try to come up with a design that will address these issues as well as how to use it to address the larger issue. 
- Once I have created the issues I would close this pull request. We'll keep you in the loop as we go through the process to make this work. 

Does this course of action sound reasonable to you?

matham:
Hey @oruebel 

Sorry about the long delay in responding. I got busy with other things and am now getting back to this.

My use case is as follows: we are trying to stimulate a slice with light while its electrical activity is being recorded on a multi-electrode array. The electrode data (as well as associated digital and analog data) is recorded in a propriety file formats that we later convert to an HDF5 file.

During the experiment, we shine light of different patterns dynamically constructed, on the slice and want to observe the electrical response to these patterns. Consequently, continuously during the experiment, for every video frame shined on the slice we record a bunch of associated info to the hdf5 file. This is the streaming data I'm talking about.

* For example, for each video frame displayed, for every object on the screen, we save the intensity of this object at that frame.
* For each video frame we also output a n-bit digital signal on a device that is synced with the frame display and that n-bits of data is recorded by the electrophys recording system for post-hoc data alignment.
* There may be other data streams that we'd save to the file live, e.g. the start and stop time of a perfusion device.

So this data is continuously streamed. I'd like to be able to just keep streaming them to the pynwb hdf5 file as the data is created/acquired. Once done, I'd take our electrophys data and convert to pynwb from the propriety format and link to it from the streamed data file so all the data is in one place.

There's another aspect to this, I couldn't figure out how to save added data without saving the whole file again through calling e.g. io.write (but it's been a month since I tried last). E.g. imagine that I need to update some configuration data while the experiment is being prepared and want to autosave every couple of minutes. But this may be a separate issue.

Thanks for your patience!

PR #331, opened 2018-01-18
Architecture docs

oruebel:
## Motivation

This pull request primarily extends the documentation of the software architecture based on text and figures from talks by @ajtritt and @oruebel.

## How to test the behavior?

Build the pynwb docs and review the Overview section

## Checklist


oruebel:
@ajtritt this pull requests includes a lot of material from your latest slides. Can you please have a look to make sure what I added is correct.

jcfr:
I would recommend against adding the pptx (aka binary files) to the source tree.

After few revision of these, the size of the repo will grow ...

oruebel:
@jcfr I tend to agree about the pptx files. The problem I have is that we don't have a good place where people can easily identify which files belong to which figure. Having them in the source tree was easiest for that for now.  Do you have any suggestions where we can put these files and still be able to easily find them when we need to change a figure?

Issue #306, opened 2017-12-20
Where should pynwb validate schema relationships?


neuromusic:
I'm currently working on getting a bunch of calcium imaging data into pynwb. I've got additional behavior data streams and stimulus streams, but currently, I'm just trying to build out a minimal file with calcium traces.

My short term problem is that I have no clean way to check whether I'm writing files correctly.

But I feel like this problem highlights broader issues with when and where pynwb enforces the schema & I'm hoping this issue can open a discussion about how we might be able to improve pynwb for users.

To start, pynwb won't let me mock up a single object without having it's dependencies specified. Practically speaking, this means I have to build out a script to write data backwards... I want to save calcium traces, but I can't create the RoiResponseSeries until I've created an ImageSegmentation, so I try to make a ImageSegmentation, but pynwb won't let me until I've created a PlaneSegmentation, etc, etc, etc.

now I've got dozens of objects in memory (6+n, where 'n' is the number of neurons in my recording) and I want to save them which means I need to register them with the NWBFile container.

This happens by creating a module and running "add_container" over and over again.

``` Python
# now let's create the processing module to store this in
ophys_module = nwb_file.create_processing_module(
    name="ophys",
    description="calcium responses",
    source="my rig",
)

ophys_module.add_container(dFF_wrapper)
ophys_module.add_container(dFF_series)
ophys_module.add_container(seg_interface)
ophys_module.add_container(segmentation)
ophys_module.add_container(imaging_plane)
for roi in roi_list:
    ophys_module.add_container(roi)
ophys_module.add_container(max_proj)
ophys_module.add_container(optical_channel)
```

Note: at this point, pynwb does absolutely zero validation or inspection of the chain of dependencies. This means that if I have a bug and lose track of the set of required objects and they don't get passed into the `add_container` method on the processing module object, **pynwb lets me save an invalid file**

https://github.com/NeurodataWithoutBorders/pynwb/issues/146

It should be impossible to save an invalid file. I need to be able to trust that every file I save with pynwb is valid and can be read back out.

And while this is true, I have no quick way to check whether or not this is true because "validate.py" is broken:

https://github.com/NeurodataWithoutBorders/pynwb/issues/127

While fixing these bugs will go a long way toward giving users some peace of mind that they will be able to read their data back out, 

## Where should pynwb validate the data model? 

[...]

Issue #146, opened 2017-10-02
pynwb does not enforce class dependencies on file write


neuromusic:
I'm writing an NWB file using the script here:
https://gist.github.com/neuromusic/28742cbd858fa874a99864e24667401c

importantly for this issue, I only register ONE of the containers with the NWBfile object... the final RoiResponseSeries

The file saves with no problems.

I can't validate the created file because #127 

However, pynwb cannot load this file because it throws a generic Exception, 32 levels deep:

```
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-6-152dc3ba765a> in <module>()
      5 from form.backends.hdf5 import HDF5IO
      6 io = HDF5IO('ophys.nwb',manager=manager, mode='r')
----> 7 read_data = io.read()

/local1/Code/pynwb/src/form/utils.pyc in func_call(*args, **kwargs)
    273                     msg = ', '.join(parse_err)
    274                     raise_from(TypeError(msg), None)
--> 275                 return func(self, **parsed['args'])
    276         else:
    277             def func_call(*args, **kwargs):

/local1/Code/pynwb/src/form/backends/io.pyc in read(self, **kwargs)
     27     def read(self, **kwargs):
     28         f_builder = self.read_builder()
---> 29         container = self.__manager.construct(f_builder)
     30         return container
     31 

/local1/Code/pynwb/src/form/utils.pyc in func_call(*args, **kwargs)
    273                     msg = ', '.join(parse_err)
    274                     raise_from(TypeError(msg), None)
--> 275                 return func(self, **parsed['args'])
    276         else:
    277             def func_call(*args, **kwargs):

/local1/Code/pynwb/src/form/build/map.pyc in construct(self, **kwargs)
     57         result = self.__containers.get(builder_id)
     58         if result is None:
---> 59             result = self.__type_map.construct(builder, self)
     60             self.prebuilt(result, builder)
     61         return result

/local1/Code/pynwb/src/form/utils.pyc in func_call(*args, **kwargs)
    273                     msg = ', '.join(parse_err)
    274                     raise_from(TypeError(msg), None)
--> 275                 return func(self, **parsed['args'])
    276         else:
    277             def func_call(*args, **kwargs):

/local1/Code/pynwb/src/form/build/map.pyc in construct(self, **kwargs)
    912             raise ValueError('No ObjectMapper found for builder of type %s' % str(container.__class__.__name__))
    913         else:
[...]

PR #339, opened 2018-01-29
extension clarifying example

bendichter:
## Motivation

In the example case, the attributes, datasets, and groups are not "additional", since they are the only ones, so I changed to code to be a bit more clear and different from the case where you are truly extending a type

### DIFF
--- a/docs/source/example.rst
+++ b/docs/source/example.rst
@@ -324,16 +324,10 @@ Create a new type
 
     from pynwb.spec import NWBGroupSpec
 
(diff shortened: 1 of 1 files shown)


ajtritt (approved):

ajtritt:
@neuromusic I would like to take this type of approach for addressing the issues you've raised in #306 and #146

ajtritt:
@bendichter can you resolve this conflict?

ajtritt:
taken care of in #458

PR #344, opened 2018-01-31
add compressed timeseries example

bendichter:


bendichter:
ok, now it's running into the actual bug. I'll fix that too

PR #358, opened 2018-02-07
Enh/ref attribute

ajtritt:
## Motivation

Attributes can store references, but HDF5IO did not have support this. 

## How to test the behavior?
see test_write_attribute_reference_container and test_write_attribute_reference_builder in tests/unit/form_tests/test_io_hdf5.py

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 PYTHON = python
+FLAKE = flake8
 COVERAGE = coverage
 
 help:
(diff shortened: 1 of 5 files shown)


ajtritt:
@dorukozturk There's some problem with my code that is dependent on both implementation and version. Do you have any suggestions on how to recreate this locally so I can sort out the problem?

jcfr:
Which python env are you using to develop ?  (anaconda, official python, ... ?)

ajtritt:
@jcfr The problem wasn't with the environment. My code was behaving differently depending on order of iteration over a dict. The iteration order is dependent on the hash function used by the dict, which is controlled by the PYTHONHASHSEED. 

To recreate the exact environment, I had to use the same seed that tox used when it ran on CircleCI. We should probably document that somewhere. 

I just fixed the iteration so it wasn't influenced by iteration order.

bendichter:
Woah that sounds like a tricky bug. Nice work!

mgrauer:
@ajtritt Would using an ordereddict help your case?

PR #369, opened 2018-02-22
update file write demo

bendichter:
add documentation for writing using a context, and use NWBHDF5IO instead of HDF5IO


ajtritt:
made obsolete by #411

PR #377, opened 2018-02-23
Fix typo on method call to load given extensions

dorukozturk:
## Motivation

When extensions keyword argument is a string or a list, namespace tries to be load via the load_namespace method (See [here](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/__init__.py#L72) and [here](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/__init__.py#L79)). Though the TypeMap object has a "load_namespaces" method (see [here](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/form/build/map.py#L730)).

Also I realized this part of the code is not covered by the tests (see [here](https://codecov.io/gh/NeurodataWithoutBorders/pynwb/src/dev/src/pynwb/__init__.py#L68)).  

Should we add tests to ensure written files can be read back using the extension files? 

## How to test the behavior?
```
mynwb = NWBHDF5IO('somefile.nwb', extensions='somepath.yaml')
```

### DIFF
--- a/src/pynwb/__init__.py
+++ b/src/pynwb/__init__.py
@@ -69,14 +69,14 @@ def get_manager(**kwargs):
         if isinstance(extensions, list):
             for ext in extensions:
                 if isinstance(ext, str):
-                    type_map.load_namespace(ext)
+                    type_map.load_namespaces(ext)
                 elif isinstance(ext, TypeMap):
                     type_map.merge(ext)
                 else:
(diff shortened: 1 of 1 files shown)


ajtritt (approved):

ajtritt:
@dorukozturk yes, a round trip test for extensions would be a great thing to have.

PR #391, opened 2018-03-09
Enh/roi refactor

ajtritt:
## Motivation

ROIs were structured inefficiently and difficult to use.

## How to test the behavior?
See docs/code/write-read-ophys.py
(This also demonstrates some read functionality)
```python
# create your NWBFile object
nwbfile = NWBFile(...)

# create acquisition metadata
optical_channel = OpticalChannel('test_optical_channel', 'optical channel source',
                                 'optical channel description', '3.14')
imaging_plane = nwbfile.create_imaging_plane('test_imaging_plane',
                                             'ophys integration tests',
                                             optical_channel,
                                             'imaging plane description',
                                             'imaging_device_1',
                                             '6.28', '2.718', 'GFP', 'somewhere in the brain',
                                             (1, 2, 1, 2, 3), 4.0, 'manifold unit', 'A frame to refer to')

# create acquisition data
image_series = TwoPhotonSeries(name='test_iS', source='a hypothetical source', dimension=[2],
                               external_file=['images.tiff'], imaging_plane=imaging_plane,
                               starting_frame=[1, 2, 3], format='tiff', timestamps=list())
nwbfile.add_acquisition(image_series)


mod = nwbfile.create_processing_module('img_seg_example', 'ophys demo', 'an example of writing Ca2+ imaging data')
img_seg = ImageSegmentation('a toy image segmentation container')
mod.add_data_interface(img_seg)
ps = img_seg.create_plane_segmentation('integration test PlaneSegmentation', 'plane segmentation description',
                                       imaging_plane, 'test_plane_seg_name', image_series)
# add two ROIs
# - first argument is the pixel mask i.e. a list of pixels and their weights
# - second argument is the image mask
w, h = 3, 3
pix_mask1 = [(0, 0, 1.1), (1, 1, 1.2), (2, 2, 1.3)]
img_mask1 = [[0.0 for x in range(w)] for y in range(h)]
img_mask1[0][0] = 1.1
img_mask1[1][1] = 1.2
img_mask1[2][2] = 1.3
ps.add_roi(pix_mask1, img_mask1)

pix_mask2 = [(0, 0, 2.1), (1, 1, 2.2)]
img_mask2 = [[0.0 for x in range(w)] for y in range(h)]
img_mask2[0][0] = 2.1
img_mask2[1][1] = 2.2
ps.add_roi(pix_mask2, img_mask2)

# add a Fluorescence container
fl = Fluorescence('a toy fluorescence container')
mod.add_data_interface(fl)
# get an ROI table region i.e. a subset of ROIs to create a RoiResponseSeries from
rt_region = ps.create_roi_table_region([0], 'the first of two ROIs')
# make some fake timeseries data
data = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
timestamps = [0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]
rrs = fl.create_roi_response_series('test_roi_response_series', 'RoiResponseSeries integration test',
                                    data, 'lumens', rt_region, timestamps=timestamps)
# write data
with NWBHDF5IO(nwb_path, 'w') as io:
    io.write(nwbfile)

# read data back in
io = NWBHDF5IO(nwb_path, 'r')
nwbfile = io.read()

# get the processing module
mod = nwbfile.get_processing_module('img_seg_example')

# get the PlaneSegmentation from the ImageSegmentation data interface
ps = mod['ImageSegmentation'].get_plane_segmentation()
img_mask1 = ps.get_image_mask(0)
pix_mask1 = ps.get_pixel_mask(0)
img_mask2 = ps.get_image_mask(1)
pix_mask2 = ps.get_pixel_mask(1)

# get the RoiResponseSeries from the Fluorescence data interface
rrs = mod['Fluorescence'].get_roi_response_series()
# get the data...
rrs_data = rrs.data
rrs_timestamps = rrs.timestamps
# and now do something cool!
```

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,7 @@ test:
 flake:
 	$(FLAKE) src/
 	$(FLAKE) tests/
+	$(FLAKE) docs/code/
 
 checkpdb:
 	find {src,tests} -name "*.py" -exec grep -Hn pdb {} \;
--- a/Makefile
+++ b/Makefile
@@ -41,8 +41,13 @@ flake:
 	$(FLAKE) src/
 	$(FLAKE) tests/
 
+checkpdb:
+	find {src,tests} -name "*.py" -exec grep -Hn pdb {} \;
+
 devtest:
-	$(PYTHON) -W ignore test.py && $(MAKE) flake
+	$(PYTHON) -W ignore test.py
+	$(MAKE) flake
+	$(MAKE) checkpdb
 
 apidoc:
 	pip install -r requirements-doc.txt
--- a/docs/code/write-read-ophys.py
+++ b/docs/code/write-read-ophys.py
@@ -1,69 +1,97 @@
+import os
+from datetime import datetime
 
-        # create your NWBFile object
-        nwbfile = NWBFile(...)
+from pynwb import NWBFile, NWBHDF5IO
+from pynwb.ophys import TwoPhotonSeries, OpticalChannel, ImageSegmentation, Fluorescence
 
-        # set up some fake data
-        nwbfile.add_acquisition(ImageSeries(name='test_iS', source='a hypothetical source', dimension=[2],
-                                external_file=['images.tiff'],
-                                starting_frame=[1, 2, 3], format='tiff', timestamps=list()))
+def main():
+    nwb_path = 'ophys_example.nwb'
+    # create your NWBFile object
+    nwbfile = NWBFile('the PyNWB tutorial', 'my first synthetic recording', 'EXAMPLE_ID', datetime.now(),
+                      experimenter='Dr. Bilbo Baggins',
+                      lab='Bag End Laboratory',
+                      institution='University of Middle Earth at the Shire',
+                      experiment_description='I went on an adventure with thirteen dwarves to reclaim vast treasures.',
+                      session_id='LONELYMTN')
 
-        optical_channel = OpticalChannel('test_optical_channel', 'optical channel source',
-                                         'optical channel description', '3.14')
-        imaging_plane = nwbfile.create_imaging_plane('test_imaging_plane',
-                                                     'ophys integration tests',
-                                                     optical_channel,
-                                                     'imaging plane description',
-                                                     'imaging_device_1',
-                                                     '6.28', '2.718', 'GFP', 'somewhere in the brain',
-                                                     (1, 2, 1, 2, 3), 4.0, 'manifold unit', 'A frame to refer to')
+    # create acquisition metadata
+    optical_channel = OpticalChannel('test_optical_channel', 'optical channel source',
+                                     'optical channel description', '3.14')
+    imaging_plane = nwbfile.create_imaging_plane('test_imaging_plane',
+                                                 'ophys integration tests',
+                                                 optical_channel,
+                                                 'imaging plane description',
+                                                 'imaging_device_1',
+                                                 '6.28', '2.718', 'GFP', 'somewhere in the brain',
+                                                 (1, 2, 1, 2, 3), 4.0, 'manifold unit', 'A frame to refer to')
 
-        mod = nwbfile.create_processing_module('img_seg_example', 'a module for demoing ImageSegmentation')
-        img_seg = ImageSegmentation('a toy image segmentation container')
-        mod.add_data_interface(img_seg)
-        pS = img_seg.create_plane_segmentation('integration test PlaneSegmentation', 'plane segmentation description',
-                                               imaging_plane, 'test_plane_seg_name', image_series)
-        # add two ROIs
-        # - first argument is the pixel mask i.e. a list of pixels and their weights
-        # - second argument is the image mask
-        w, h = 5, 5
-        pix_mask1 = [(1, 2, 1.1), (3, 4, 1.2), (5, 6, 1.3)]
-        img_mask1 = [[0.0 for x in range(w)] for y in range(h)]
-        img_mask1[1][2] = 1.1
-        img_mask1[3][4] = 1.2
-        img_mask1[5][6] = 1.3
-        pS.add_roi(pix_mask1, img_mask1)
-
-        pix_mask2 = [(7, 8, 2.1), (9, 10, 2.2)]
-        img_mask2 = [[0.0 for x in range(w)] for y in range(h)]
-        img_mask2[7][8] = 2.1
-        img_mask2[9][10] = 2.2
-        pS.add_roi(pix_mask2, img_mask2)
-
-        # get an ROI table region i.e. a subset of ROIs to create a RoiResponseSeries from
-        rt_region = plane_segmentation.create_roi_table_region([0], 'the first of two ROIs')
-        # make some fake timeseries data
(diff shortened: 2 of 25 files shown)


oruebel (dismissed):
This looks great! I think this will work much better than the original ROI structure.

oruebel (approved):

neuromusic:
Andrew,

I didn't get a chance to review this over the weekend. Am I reading this correctly that there is no longer a way to label ROIs with a cross-file unique identifier (like a database index or a UUID)? They are only defined by the index where they happen to land in the table?

Thanks for the work on this!

ajtritt:
@neuromusic Yes, you are correct. Requiring the user provide a unique ID seemed like another usability hurdle, so I removed it. I can add that back in though.

Andrew

bendichter:
@ajtritt It seems like an optional argument might be best here

ajtritt:
If there is good justification, I can add optional argument for this. I tend to think that optional fields are an all around lose-lose situation. They make development harder--everything that works with ROIs now has to always check for the existence of a unique ID, and it is fundamentally at odds with a standard--how can this be a standard if no field is ever guaranteed to exist. 

Personally, I think requiring a unique ID is perfectly reasonable; however, there has been a lot of pushback to make this component easier to use. I am happy to add unique IDs back in as a required field. If an optional field is really wanted, I will need some justification beyond mere usability to prioritize this feature.

bendichter:
Clearly experimenters that do cross-session linking of ROIs require some field to do this. What about datasets that don't have this? Maybe there is one session per animal so cross-session ROIs are unneeded, or maybe the data is stored before doing ROI matching. If it's a required field you could try to use a placeholder (I am currently using `np.nan` in a bunch of places where I don't have the information), but any placeholder you choose in this case has a risk of being mistaken for a UID that will be mistakenly used to match ROIs.

NWB needs to be flexible enough to be usable by many different labs with different use-cases. I think optional fields are going to be necessary in some cases.

ajtritt:
In terms of data management, placeholders just kick the can down the road and create a new/different problem. I don't want to endorse this solution or make it easy to carry out this bad practice.

If a users wants to use a placeholder, that is their choice; however, I want to avoid making optional fields in the API that sit on top of fields that are required by the schema.

bendichter:
How would you recommend someone express through NWB that they do not have cross-session ROIs?

ajtritt:
Said person can just create unique ids themselves. 

This is how electrodes are set up. The electrode table has a column for a unique identifier, and users populate it on their own

neuromusic:
"Clearly experimenters that do cross-session linking of ROIs require some field to do this."

Yes, this is absolutely 100% critical to our use case. 

One option would be that we could build an an extension that adds identifiers to ROIs by indexing into the ROI table... but then what's the point in having a standard?

At the same time, this use case may be overkill for the broader needs of the community... hence why an "optional" field is appropriate. nullable fields are not "fundamentally at odds with a standard", as they designate where particular kinds of (meta)data should live, even if they are not relevant to all uses. 

The W3C standard for HTML designates what fields can exist in headers, not that every type of field need be present or that every field needs every attribute. ISO-8601 specifies what datestamps should look like, regardless of whether they have month or day or hour or timezone specificity. But if your date has ms precision, there is a clear way to designate it as such and for others to understand it.

Indeed, this kind of flexibility is precisely what makes it a community standard as opposed to an institutional or laboratory standard. A single lab can enforce a comprehensive specification internally... as you note, this is technically easier to do than build a community standard. But the whole point of this endeavor is a community standard.

This is as good a time as any to re-hash that I think the NWB specification goes WAY overboard in enforcing required fields that should be optional (likely to the detriment of the specification). In fact, I would welcome making ROIs optional for RoiResponseSeries, period.

also, "source" should be optional: https://github.com/NeurodataWithoutBorders/nwb-schema/issues/53#issuecomment-329247562

and "session_start_time" https://github.com/NeurodataWithoutBorders/nwb-schema/issues/49#issuecomment-326325025

p.s. @ajtritt this PR effectively solves #61 with the `dF/F  <->  ROIs  <->  segmentation` schema I proposed on that issue https://github.com/NeurodataWithoutBorders/nwb-schema/issues/61#issuecomment-332046841

oruebel:
> The W3C standard for HTML

This is comparing apples and oranges. HTML is a standard markup language not a data format. You don't need to use all parts of a language (e.g. HTML) in order to create a sensible document. However, in order to understand and interpret the content of a document (i.e., the data) you need to have the complete document and its context. In that regard:

> In fact, I would welcome making ROIs optional for RoiResponseSeries, period.

How do you know what an RoiResponseSeries means without knowing what the ROIs are?

The purpose of a data standard is not just to define what things should be called but it's also there to make sure that the data is interpretable. A data standard that makes everything optional is really not a standard because it means that folks omit a lot of information that is ultimately needed to understand the data. I understand that required fields often mean that folks have to think how to actually get that information but I don't think that is necessarily a bad thing. Folks should think about their data and not just leave out important information because it makes things easier at first. 

Ultimately, there needs to be a balance between what should be required in order to ensure that data is usable and what can be optional. In terms of what data is being required/optional NWB 2 is still very close to NWB 1.x. I don't mind if a field is optional if it is not needed to understand the data but if the information in a field is needed to understand and interpret the data than it should be required.  After all, once a field is optional you can no longer rely on the field in your data analysis. Optional fields are a double-edged-sword, if a field is optional it may make it "easier" to write data but it also makes it harder to write consistent APIs, analysis codes, and tools that may rely on or change behavior based on the corresponding information.

nicain:
FWIW I am contributing to an analysis project with df/f signals that are computed from spike trains by a model (mlspike) .  I would like to treat these signals as similar as possible to real in vivo recordings, but as it currently stands I dont think I can store this signal in an nwb file (it is not associated with an ROI, so no RoiResponseSeries).

oruebel:
@nicain maybe adding a simpler base-type that then ROIReponseSeries would extend might be the way to go here? 

At this point I would like us to be able to merge this refactor of the ROI stuff with dev to get it out. From what I can tell, this refactor seems to be major improvement over what we have, so lets role it out and then we can move the discussion about making things optional to a new thread.

neuromusic:
How is this going to effect your reading of legacy Brain Observatory data, @nicain?

neuromusic:
"How do you know what an RoiResponseSeries means without knowing what the ROIs are?"

That depends on the analysis. For many analyses (and indeed, for most of the preprints that have been posted on Brain Observatory data), it doesn't matter... all that matters is that the dimensions of the array (neurons and time) are known. The provenance of the traces is secondary and not required for most analyses.

neuromusic:
"HTML is a standard markup language not a data format."

fine. NetCDF, BIDS, Neo, Exif...

oruebel:
@neuromusic I'd suggest the following:
1. create a new issue for making fields optional
1. create a list of fields that are currently required but should be optional and why
1. lets go through the list at the Allen hackthon to decide on changes
1. identify a person that should implement those changes  (preferably someone other than @ajtritt since he is already busy with lots of other changes and these things should be fairly straight forward for other to implement)

PR #396, opened 2018-03-16
Enh/epoch refactor

ajtritt:
## Motivation

Epochs were being stored inefficiently

## How to test the behavior?
``` python
from pynwb import NWBFile 
from pynwb.ecephys import 

f = NWBFile(...)
ephys_ts = ElectricalSeries(...)
f.add_acquisition(ephys_ts)

spatial_ts = SpatialSeries(...)
f.add_acquisition(spatial_ts)

f.create_epoch(name='epoch1', start_time=0.0, stop_time=1.0, tags=epoch_tags,
               description="the first test epoch", timeseries=[ephys_ts, spatial_ts])

f.create_epoch(name='epoch2', start_time=0.0, stop_time=1.0, tags=epoch_tags,
               description="the second test epoch", timeseries=[ephys_ts, spatial_ts])
```

## Checklist

### DIFF
--- a/docs/code/creating-and-writing-nwbfile.py
+++ b/docs/code/creating-and-writing-nwbfile.py
@@ -39,18 +39,6 @@ def main():
 
     os.remove(filename)
 
-    # create-epochs: start
-    epoch_tags = ('example_epoch',)
-
-    ep1 = f.create_epoch(source='an hypothetical source', name='epoch1', start=0.0, stop=1.0,
-                         tags=epoch_tags,
-                         description="the first test epoch")
-
-    ep2 = f.create_epoch(source='an hypothetical source', name='epoch2', start=0.0, stop=1.0,
-                         tags=epoch_tags,
-                         description="the second test epoch")
-    # create-epochs: end
-
     # create-device: start
     device = f.create_device(name='trodes_rig123', source="a source")
     # create-device: end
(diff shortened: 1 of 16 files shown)


neuromusic:
so if I have 4000 epochs, I need to define a source and description for each and every one?

oruebel:
@neuromusic I'm glad your are reviewing these changes. Instead of taking a critical approach to comments and feedback:

> so if I have 4000 epochs, I need to define a source and description for each and every one?

I'd appreciate it if you could try to formulate these kind of question in a more inclusive and constructive fashion. E.g., you could say something like: "Thanks for this change. Do I understand this structure correctly that every Epoch requires a separate source and description? In cases where users need to create large numbers of Epochs I think this could be problematic. Do you have any suggestions on how to address this potential problem?" We are all working very hard on this and constructive feedback is always better than criticism. Thanks again for all your efforts on this project!

ajtritt:
@neuromusic For _source_, no--this is a typo. For _description_, yes--if you are going to set 4000 start/stop times, you need to differentiate them and provide information about what's starting/stopping. 

If you are trying to store trial data, then epochs alone is probably the wrong solution.

oruebel (dismissed):

oruebel (dismissed):
Thanks for the additional fix. Looks good.

ajtritt:
@neuromusic I added a type called "EventTable" that has three columns: _start_time_, _stop_time_, and _tags_ (_tags_ is just a string of tags). The PyNWB class is in pynwb.epochs.EpochTable. You can extend this for your needs.

bendichter:
@ajtritt I'm having a bit of trouble incorporating this into my conversion script. Previously, I was able to link a single epoch to multiple timeseries objects at once. Now, when I add a row, the `timeseries` argument expects a `RegionSlicer` object, and throws an error when I try to use a list of `TimeSeries` objects.

bendichter:
or am I supposed to use `pynwb.epoch.Epochs.add_epoch`

PR #400, opened 2018-03-20
Enh/hdf5comp

oruebel:
## Motivation

Enable users to  directly specify advanced I/O options for the writing of datasets, specifically chunking and compression. 

# Summary of changes
* Expanded H5DataIO to enable users to specify more advanced I/O settings
* Updated the HDF5IO backend to support the use of the expanded I/O settings of H5DataIO
* Removed direct support Iterable from H5DataIO.write_dataset and instead require users to explicitly wrap iterables using DataChunkIterator. This is to avoid implicit behavior. For convenience I added a factory function for DataChunkIterator to generate it from an Iterable. 
* Added max_shape and dtype functions to AbstractDataChunkIterator to ensure we only assume functionality that is on the base class, rather than DataChunkIterator.
* Added/expanded tests to test the new options

## How to test the behavior?

The following is a direct test derived from the test suit that uses the FORM API only. 

```
import tempfile
import numpy as np
from pynwb.form.backends.hdf5.h5tools import HDF5IO
from pynwb.form.backends.hdf5 import H5DataIO
from pynwb.form.build import DatasetBuilder
test_temp_file = tempfile.NamedTemporaryFile()
test_temp_file.close()
io = HDF5IO(test_temp_file.name)
f = io._file
a = H5DataIO(np.arange(30).reshape(5, 2, 3), compression='gzip', compression_opts=5, shuffle=True, fletcher32=True)
io.write_dataset(f, DatasetBuilder('test_dataset', a, attributes={}))
dset = f['test_dataset']
print("Compression: %s" % str(dset.compression))
print("Compression Options: %s" % str(dset.compression_opts))
print("Shuffle: %s" % str(dset.shuffle))
print("Fletcher32: %s" %  str(dset.fletcher32))
io.close()
```

When using PyNWB containers one woulld hand the H5DataIO object as an input for a specific dataset, e.g., the data of a TimeSeries.

## Checklist


oruebel:
@ajtritt can you have a look at this. I think this should be ready to be merged now. This mainly updates H5DataIO and HDF5IO to be able to define IO options for datasets, e.g., chunking and compression, but also linking/copy of h5py.Dataset. For the latter I also modified FORMIO to be able to pass parameters from FROMIO.write to FORMIO.write_builder.

bendichter:
@jcfr had some good ideas about using sha tags to ensure that linked files are correct.

ajtritt (approved):

PR #405, opened 2018-03-22
[WORK IN PROGRESS] Read api tests

dorukozturk:
## Motivation

Adding new tests for asserting read api is working with different files. This is a "work in progress" pr where I will be experimenting/finding a balance between speed of tests and convenience.


ajtritt:
I noticed you added a shell script to download files. Will this work with testing on Windows?

dorukozturk:
@ajtritt Good point. I totally forgot about it. I will go with python to download the files. Can you or @nicain help me fixing the legacy_read tests?

ajtritt:
2.0 has diverged from 1.x enough to make backwards compatibility non-trivial. At this point, there are other issues that are higher priority, so I turned off the test for now. I think it is in our best interest to not get bogged down making old 1.x files work and keep this issue focused on 2.0 files.

jcfr:
> 2.0 has diverged from 1.x enough to make backwards compatibility non-trivial.

:+1: 

Then, documenting how the data can be (are being / will be) processed and updated to work with the format will be sensible i think. This is not something core developer have to do, I envision it will be a collaborative endeavor. 

As mentioned by @ajtritt , the current pynwb doesn't have to read the older format. It would make the code base too complex and harder to maintain.

That said, having hosted at [place to be defined] scripts, documentation and guidelines on how to update older datasets may help.

Cc: @nicain

dorukozturk:
@ajtritt Ok, that sounds good. I assume labs are waiting for 2.0 schema to be stabilized/finalized to generate their files according to 2.0 schema. Is this more or less correct? :)

ajtritt:
@dorukozturk Yes. I think they are also waiting for documentation, too :)

@jcfr 
> Then, documenting how the data can be (are being / will be) processed and updated to work with the format will be sensible i think. This is not something core developer have to do, I envision it will be a collaborative endeavor.

I'm not quite sure what you mean. I don't think it benefits users to talk about the details of how these things have changed. We have been documenting schema changes--they can go look at it if they want, but it should not be a focal point for user training. Moving forward, we should xdirect users to the PyNWB front end and the 2.0 schema.

> That said, having hosted at [place to be defined] scripts, documentation and guidelines on how to update older datasets may help.

Anyone that has data in 1.x converted to 1.x from the format it was recording in and is reading 1.x files it with h5py. The majority of 1.x users know how to get their data, either from it's original source or out of 1.x files. Since they already know how to get their data out, and will be using PyNWB moving forward anyway, the most effective "scripts, documentation and guidelines on how to update older datasets" is good documentation on how to write 2.0 with PyNWB

My general stance on 1.x vs 2.0 is that we do not spend time on 1.x since it is effectively unsupported and was insufficient to begin with. Our limited resources are better spent helping users learn how to use the new API and developing the features they need/want.

oruebel:
> Our limited resources are better spent helping users learn how to use the new API and developing the features they need/want.

I absolutely agree!

jcfr:
>I'm not quite sure what you mean. I don't think it benefits users to talk about the details of how these things have changed. 
> We have been documenting schema changes--they can go look at it if they want, but it should not be a focal point for user training. Moving forward, we should direct users to the PyNWB front end and the 2.0 schema.

This makes complete sense. And there is already a complete documentation for the schema :+1: 

I think this page is particularly important: http://nwb-schema.readthedocs.io/en/latest/format_release_notes.html#specification-language-changes


> Anyone that has data in 1.x converted to 1.x from the format it was recording in and is reading 1.x files it 
> with h5py. The majority of 1.x users know how to get their data, either from it's original source or out of 1.x files

This is great. 

@nicain  What is currently the barrier preventing the data from being converted to the new schema ? Are you just waiting for the stable release of the version 2.0 of the schema ? 

All of that said, I still wonder if having a "place" that users from the "community" could improve and maintain by sharing the lesson learned, scripts, ...  would be useful.

PR #409, opened 2018-03-30
Fix virtualenv activation

t-b:
The activate script is not executable on linux. Source it which works with sh and bash.


dorukozturk:
@t-b Nice catch. Thanks !!!

dorukozturk (approved):

PR #417, opened 2018-04-04
Use NaN as default resolution for TimeSeries objects

t-b:
The NWB schema documentation says

Smallest meaningful difference between values in data, stored in the specified by unit.
Comment: E.g., the change in value of the least significant bit, or a
larger number if signal noise is known to be present. If unknown, use NaN.

So using NaN instead of zero is a more natural choice as a resolution of zero
means that you have perfect experimental conditions which are in general
rarely available.


t-b:
Okay I'm not understanding the tests well enough to get them fixed. Somehow the resolution is still stuck to 0 at some place.

t-b:
@ajtritt Tests work :)

tests/integration/ui_write/base.py
```diff
@@ -45,6 +47,22 @@ def container(self):
...
+    def assertDictEqual(self, d1, d2, msg=None):
+        """
+        Custom dictionary comparison routine treating NaN == NaN being true
+        """
+        self.assertIsInstance(d1, dict, 'First argument is not a dictionary')
+        self.assertIsInstance(d2, dict, 'Second argument is not a dictionary')
+
+        if d1 != d2 and len('\n'.join(difflib.context_diff(pprint.pformat(d1).splitlines(),
+            pprint.pformat(d2).splitlines()))) > 0:
+            standardMsg = '%s != %s' % _common_shorten_repr(d1, d2)
```
ajtritt: Looks like you missed the function `_common_shorten_repr`  when copying over the assert method.

ajtritt (changes requested):
Fix missing function issue

tests/integration/ui_write/base.py
t-b: @ajtritt Thanks, I'll fix that.

t-b: @ajtritt Done.

t-b:
@ajtritt Fixed.

t-b:
@ajtritt Any hints why the circleci python 2.7 job does not pass?

ajtritt:
Looks like the equality comparison is breaking on unicode-ascii comparisons.

t-b:
@ajtritt Hmm... But the miniconda27 passes. I look into that when I get a change.

t-b:
I can reproduce the failure locally with python 2.7.

t-b:
@ajtritt I finally found a not too hacky way...

t-b:
@ajtritt Regarding the review: I (think) I adressed your comments. Only github seems not to agree.

t-b:
@oruebel @ajtritt With the approaching release, is that good to go?

ajtritt:
@t-b #610 addressed this. I changed resolution to default to -1.0. The logic being, since resolution is a difference, it should never be positive, so it's just as informative in this context as NaN. Using -1.0 has the added benefit of being a floating point number that behaves as one would expect a floating point number, namely `-1.0 == -1.0` is *true*, while `NaN == NaN` is *false* 

Does this meet your needs?

t-b:
@ajtritt Thanks for the update. From a purist technical POV I still prefer `Nan` but a negative resolution does the trick as well.

PR #426, opened 2018-04-04
Fix timestamps

t-b:
The timestamps in `/start_sesssion_time` and `/file_create_date` are with that change now both in UTC timezone.


oruebel:
@t-b should this also be mentioned in the schema itself?

t-b:
@oruebel The schema itself is pretty ambigous, see https://github.com/NeurodataWithoutBorders/nwb-schema/pull/71. I'm happy to discuss that PR here at the hackathon.

ajtritt (approved):

src/pynwb/file.py
```diff
@@ -216,7 +216,7 @@ def __init__(self, **kwargs):
             self.__session_start_time = parse_date(self.__session_start_time)
```
bendichter, line 216: Isn't this the one you wanted to change?

src/pynwb/file.py
```diff
@@ -203,7 +203,7 @@ def __init__(self, **kwargs):
         pargs, pkwargs = fmt_docval_args(super(NWBFile, self).__init__, kwargs)
         pkwargs['name'] = 'root'
         super(NWBFile, self).__init__(*pargs, **pkwargs)
-        self.__start_time = datetime.utcnow()
+        self.__start_time = datetime.utcnow().isoformat() + "Z"
```
bendichter, line 206: According to the [schema documentation](http://pynwb.readthedocs.io/en/latest/pynwb.file.html?highlight=NWBFile) there's a `session_start_time` and a `file_create_date` but no `start_time` in `NWBFile`, so I don't know what this is. An accident perhaps? I think the value you meant to change is below.

bendichter (commented):
Sorry for the late review here. I think there might have been a mistake in which field was edited, which I've marked in the code, but could we also discuss the implementation?

If times are all in UTC, does that mean a user would not know what time of day it was when the session started? I think that might be more important than having a worldwide uniform clock. I suppose you could look up the timezone of the institution, but that's a chore that's difficult to automate and will fail for a variety of reasons.

Also, if the experimenters, like normal human beings, keep the clocks of their computers in local time, there will be a mismatch between when their computers said the session was recorded and when the NWB file reports.

This behavior goes against what I would expect as a user. At least this seems to require an addition of a timezone field. I realize UTC is standard practice in CS, but it isn't that standard in neuroscience. On the other hand maybe if we are building this tool for worldwide sharing of data it's worthwhile to revisit these conventions. Could someone chime in with the big PROs of this approach?

jeffteeters:
I think a good approach would be to store the local time and include the timezone, perhaps as an offset to UTC.  Doing that would provide information about the local time, the location, and also the universal time.   For example, "2018-04-13T16:10:15.333-08:00" (ISO 8601 format) would indicate 15.333 seconds after 4:10PM local time (16:10:15.333), the -08:00 would indicate Pacific Time (the timezone is 8 hours behind UTC).

bendichter:
Yeah, I agree, that would be a good compromise. We can infer the timezone with 
```python
import pendulum
dt = pendulum.now().to_iso8601_string()
```
or there may be a way to do it without adding a dependency.

Issue #431, opened 2018-04-05
Associate cluster waveforms and electrode table regions with unit times


NileGraddis:
UnitTimes provide a nice way to store event times associated with sorted units. There should be a way to associate mean/std waveforms (such as those in ClusterWaveforms) as well as electrodes (such as those in ElectrodeTableRegion) to the units defined by a UnitTimes container.

Since UnitTimes supports simulated data and ophys data as well as ephys data, the solution to this problem should probably involve referencing units in UnitTimes from other containers rather than altering UnitTimes itself.

ajtritt:
What about a type that maps between the different IDs? It could be a table with two columns. For example, in the ElectrodeTable to UnitTimes case, the first column would be an id from the electrode table and the second column would be an id from UnitTimes.unit_id.

tjd2002:
I support a separate class to represent Units (independent of UnitTimes/ClusterWaveforms).And NB we definitely want to refer to electrodetableregions and not just electrodeids.

What is the reason to separate this list of Units from the UnitMetrics class as proposed at PR #435? Are there cases where there will not be a 1:1 correspondence between UnitMetrics and Units?

NileGraddis:
The way that I am currently leaning towards implementing Units is as a group with two datasets:

1. a list of unit ids
2. a table of associations, with one column of unit ids and another of links, like this:
```
  - neurodata_type_def: UnitAssociations
    neurodata_type_inc: NWBData
    doc: table of unit associations
    attributes:
      - name: help
        doc: a help string
        dtype: text
        value: table of unit associations
    dtype:
      - name: unit_id
        doc: id of unit
        dtype: int
      - name: association
        doc: a reference
        dtype:
          target_type: NWBBaseType
          reftype: object
``` 

This would allow us to attach electrode groups, event times, waveforms and other data to units while preserving the type of the attached data. We could then units obtained from different modalities (ecephys, ophys, models) in a cohesive framework.

I think that @tjd2002's suggestion is reasonable. The UnitMetrics could be attached to the file as a subgroup of the Units

bendichter:
These all sound like good solutions. I just want to make sure that whatever we pick can support, either by links or some other means, indicating the relationship between individual units and:

* average waveform
* continuous variable like mean firing rate
* categorical variable like cell type
* spike train
* tuning properties/receptive field (often described as a matrix, e.g. STRF, location preference, etc.)
any more?

NileGraddis:
[...]

PR #435, opened 2018-04-05
Unitwise metrics

### LINKED ISSUE #427: Generalize cluster quality metrics
The [clustering container](https://pynwb.readthedocs.io/en/latest/pynwb.ecephys.html#pynwb.ecephys.Clustering) has a dataset called "peak_over_rms". This is a metric of the quality of clustered units. Such metrics will likely be produced by every extracellular ephys experiment, but will not necessarily be peak_over_rms.  The peak_over_rms field should probably be replaced by a more generic quality metric field.

I will take a stab at this if there are no objections.

NileGraddis:
Resolves #427 

Adds a generic class for unit-wise metrics (such as the mean firing rate of a sorted ephys unit) and a specialization for peak_over_rms (the existing version of a unit metric). Adds a Multicontainer Interface for storing these metrics, which also stores a link to a unittimes group. Adds unit and integration tests for theses features

### DIFF
--- a/src/pynwb/form/spec/write.py
+++ b/src/pynwb/form/spec/write.py
@@ -33,7 +33,7 @@ class YAMLSpecWriter(SpecWriter):
         self.__outdir = getargs('outdir', kwargs)
 
     def __dump_spec(self, specs, stream):
-        yaml.safe_dump(json.loads(json.dumps(specs)), stream, default_flow_style=False)
+        yaml.main.safe_dump(json.loads(json.dumps(specs)), stream, default_flow_style=False)
 
     def write_spec(self, spec_file_dict, path):
         with open(os.path.join(self.__outdir, path), 'w') as stream:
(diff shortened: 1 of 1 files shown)


src/pynwb/data/nwb.misc.yaml
```diff
@@ -139,3 +139,38 @@ groups:
...
+    - name: unit_times
+      target_type: UnitTimes
+      doc: defines the units from which these metrics were computed
+      quantity: 1
+  datasets: 
+    - neurodata_type_inc: UnitMetric
+      doc: a metric
+      quantity: '*'
+datasets:
+- neurodata_type_def: UnitMetric
```
ajtritt: Can we add a required description field here? If we allow for arbitrary metrics, we should also require that users provide a description of the metric they are storing.

ajtritt (changes requested):

tjd2002:
Per discussion with @NileGraddis it seems like a good idea to expand this PR to also include the proposed Unit class discussed at #431 . While this class would stand alone it seems preferable to refactor all the Unit-wise stuff together.

src/pynwb/data/nwb.misc.yaml
tjd2002: Agreed

bendichter:
@ajtritt can we close this?

PR #437, opened 2018-04-06
Exclude example and integration tests from coverage

### LINKED ISSUE #423: Run a coverage report without integration tests
The coverage tests report too high a coverage value because the integration tests are inflating the value and hiding the actual coverage of unit tests.

dorukozturk:
## Motivation

Fixes #423 
Coverage is only useful in the context of unit testing. Integration tests create the illusion of a higher coverage than what we currently have. Example tests are just executing tutorial code but they are not asserting any conditions explicitly so I excluded also the example tests from coverage. 

## How to test the behavior?
```
make coverage
```

## Checklist

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,7 +23,7 @@ commands =
 basepython = python3.6
 
 commands =
-    coverage run test.py
+    coverage run test.py --form --pynwb
     coverage html -d tests/coverage/htmlcov
 
 # Envs that builds wheels and source distribution
@@ -47,7 +47,7 @@ commands = {[testenv:build]commands}
 # Envs that will only be executed on CI that does coverage reporting
 [testenv:coverage]
 commands =
-    coverage run test.py
+    coverage run test.py --form --pynwb
     coverage report -m
     codecov -X fix
 


tox.ini
```diff
@@ -47,7 +47,7 @@ commands = {[testenv:build]commands}
 # Envs that will only be executed on CI that does coverage reporting
 [testenv:coverage]
 commands =
-    coverage run test.py
+    coverage run test.py -f -p
```
jcfr: I suggest to use the long argument ``--form``  and ``--pynwb``

dorukozturk: @jcfr I changed it to use the long argument names. Is it good now?

dorukozturk:
@ajtritt Jc is at vacation and might not see this. Can you review it?

ajtritt (approved):

waxlamp:
@dorukozturk, will this trigger an updated coverage report? Where can we see that? Thanks for this PR!

dorukozturk:
@waxlamp Yes, an update already happened in the report. Coverage was %77 before and now it is %68 due to the fact that we excluded example and integration tests. [This](https://codecov.io/gh/NeurodataWithoutBorders/pynwb/tree/dev/src/pynwb) is the current coverage.

waxlamp:
Awesome!

Should we exclude `legacy` from coverage as well, or is the testing there needed?

dorukozturk:
I think we want to keep it and even improve testing. @nicain Is that the case?

Issue #441, opened 2018-04-06
Make setUpBuilder optional in test case


NileGraddis:
Having to write out the builders manually makes writing roundtrip tests more cumbersome, though it is helpful for debugging new containers. It would be nice (and make contributing easier!) if this were optional. As far as I can tell, the relevant class is tests.integration.ui_write.base.TestMapNWBContainer - specifically the setUp and test_build methods.

thoughts?

ajtritt:
relates to #440

bendichter:
Agreed. Thanks for bringing this up, Nile!

ajtritt:
closed by #456

Issue #440, opened 2018-04-06
Add docs for creating a new round trip test


mgrauer:
We should make it easier to add a new round trip test, at least some documentation of the process and the point of adding a round trip test, and a pointer to an existing example (or ideally new example that is brought in at the same time as the docs to exercise the new docs).

PR #450, opened 2018-04-09
Round trips2 and UnitTimes bug fix

### LINKED ISSUE #449: UnitTime roundtrip changes type returned by `get_unit_spike_times`
## Bug

Before IO, `UnitTimes` returns spike times as a list. After IO, spike times are returned as a numpy array. I guess either is fine. I prefer np.arrays, but it should be consistent. 

### Steps to Reproduce
```python
from pynwb.misc import UnitTimes
from base import ContainerRoundTrip

ut = UnitTimes('UnitTimes add_spike_times unit test')
ut.add_spike_times(0, [0, 1, 2])
print(type(ut.get_unit_spike_times(0)))

with ContainerRoundTrip(ut) as ut2:
    print(type(ut2.get_unit_spike_times(0)))
```

```
<class 'list'>
<class 'numpy.ndarray'>
```

(ContainerRoundTrip is in #450)
### Environment
Please describe your environment according to the following bullet points.
- **Python Executable:** Conda
- **Python Version:** Python 3.6
- **Operating System:** macOS
- **Pynwb Version:** latest dev

## Checklist

bendichter:
## Motivation

Add a way to compactly round-trip containers. Used it to round trip `UnitTimes` and found a little bug: the type of the return of `UnitTimes.get_spike_unit_times` is different before and after IO. fix #449 

## Checklist


ajtritt:
This duplicates
https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/tests/integration/ui_write/base.py#L106
and 
https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/tests/integration/ui_write/base.py#L154

Relates to #440 and #441

bendichter:
@ajtritt I've found those classes pretty difficult to understand and to use. I think they require builders? I'm really interested in making sure the read and write functions work as expected and for that purpose I think the use of contexts makes this a lot simpler. Look, the round-trip part of the test in `test_misc.py` is one line! (line 85)

ajtritt:
Yeah, I get that they're difficult to use because of the Builder issue, but this PR adds duplicate code that requires the test-writer to explicitly check that attributes are equal. 

If the setUpBuilder implementation gets removed (i.e. #441 gets resolved), you only need to construct an object and no other test code. 

There are also many other test cases implemented using the existing framework.

ajtritt:
Also, a UnitTimes roundtrip test already exists in [here](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/tests/integration/ui_write/test_ecephys.py#L13)

src/pynwb/misc.py
```diff
@@ -231,7 +231,7 @@ def __init__(self, **kwargs):
              'doc': 'the index of the unit in unit_ids to retrieve spike times for'})
     def get_unit_spike_times(self, **kwargs):
         index = getargs('index', kwargs)
-        return self.__iv.get_vector(index)
+        return np.array(self.__iv.get_vector(index))
```
ajtritt, line 234: This is okay if referential equality if never required for the return values of this method.  This now creates the property that `self.get_unit_spike_times(i) != self.get_unit_spike_times(i)`. 

This will also duplicate whatever `IndexedVector.get_vector` returns. Assuming we never want `IndexedVector.get_vector` to cache, this probably won't do much to memory footprint, since the GC will pick up the first copy, but it does incur the extra overhead allocating/copying.

waxlamp:
It seems a lot like there may be a gap in documentation in how to use the existing testing infrastructure? If that's not true, @ajtritt, can you point me to the docs? And if the docs really are missing, we should write some, and maybe take this opportunity to "impedance match" between what exists and what @bendichter would want to have instead.

ajtritt:
Thanks @waxlamp. We already have this issue #440 to address the lack of documentation for writing tests, which I am currently working on.

waxlamp:
Excellent, thanks @ajtritt

ajtritt:
@bendichter #456 should address the motivation behind this PR. can we close this and move the change for consistent results from `get_unit_spike_times` to a new PR?

@waxlamp [documentation](http://pynwb.readthedocs.io/en/latest/make_roundtrip_test.html) for roundtrip tests.

bendichter:
Sure

PR #456, opened 2018-04-10
Enh/roundtrip tests

### LINKED ISSUE #440: Add docs for creating a new round trip test

### LINKED ISSUE #441: Make setUpBuilder optional in test case

ajtritt:
## Motivation

Make writing roundtrip tests easier.   Fix #440, #441 

## How to test the behavior?
```python
from . import base

# Roundtrip an NWBContainer i.e. something that isn't allowed in 'acquisition`
class TestElectrodeGroupIO(base.TestMapRoundTrip):

    def setUpContainer(self):
        self.dev1 = Device('dev1', 'a test source')  # noqa: F405
        return ElectrodeGroup('elec1', 'a test source',  # noqa: F405
                                        'a test ElectrodeGroup',
                                        'a nonexistent place',
                                        self.dev1)

    def addContainer(self, nwbfile):
        ''' Should take an NWBFile object and add the container to it '''
        nwbfile.add_device(self.dev1)
        nwbfile.add_electrode_group(self.container)

    def getContainer(self, nwbfile):
        return nwbfile.get_electrode_group(self.container.name)

# Roundtrip an NWBDataInterface i.e. something that can go in 'acquisition`
class TestUnitTimes(base.TestDataInterfaceIO):

    def setUpContainer(self):
        ut = UnitTimes('UnitTimes add_spike_times unit test')
        ut.add_spike_times(0, [0, 1, 2])
        ut.add_spike_times(1, [3, 4, 5])

    # roundtrip the container, and get it back for some additional tests
    def test_my_special_test(self)
        roundtripped = self.roundtripContainer()
        # now do something in addition to testing attribute equality

```

relates to #450 

## Checklist

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -43,6 +43,7 @@ breaking down the barriers to data sharing in neuroscience.
    validation
    api_docs
    software_process
+   make_roundtrip_test
 
 .. toctree::
    :maxdepth: 2
--- a/docs/source/make_roundtrip_test.rst
+++ b/docs/source/make_roundtrip_test.rst
@@ -66,7 +66,7 @@ also where you can add additonial containers that your container of interest dep
 :py:class:`~pynwb.ecephys.Device` dependencies.
 
 
-Continuing from our example above, we will add the method for adding a generic :py:class:`~pynwb.base.TimeSeries`: instance:
+Continuing from our example above, we will add the method for adding a generic :py:class:`~pynwb.base.TimeSeries` instance:
 
 
 .. code-block:: python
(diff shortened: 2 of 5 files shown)


oruebel (dismissed):

oruebel:
@ajtritt just to make sure I understand this change correctly. If I see this correctly than all that one now needs to do to implement a round-trip-test for a specific container type is to implement the simple``setUpContainer``, ``addContainer``, and ``getContainer`` methods. This is great! :+1:

Looking at the design, it seems like this would also be fairly simple to extend in the future to the case where one might want to test combinations of containers basically by allowing ``TestMapRoundTrip.container`` to be a list of containers.

oruebel (approved):

PR #462, opened 2018-04-13
PEP8Speaks: Only scan the diff of the changed files for violations

t-b:
Currently PEP8Speaks checks all changed files. This results in warning
messages for lines which were not touched in a given PR.

This is, at least for casual contributors, quite annoying as it basically
forces them to cleanup after other people's contributions.

Let's scan only the changed lines to make it less annoying.

### DIFF
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -2,3 +2,5 @@ pycodestyle:
     max-line-length: 120
     ignore:
         - E402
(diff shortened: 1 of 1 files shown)


dorukozturk:
@t-b This makes sense and I think it is much better.

dorukozturk (approved):

PR #476, opened 2018-04-21
remove NWBFile.__start_time

bendichter:
nwb-schema specifies a `file_created_date` and a `session_start_time` but no `start_time` for NWBFile. Also, this value appears to be unused anywhere else. I think it was added by mistake

## Checklist


tjd2002:
The field also doesn't appear in src/pynwb/data/nwb.file.yaml.

A little digging with git-blame shows that this field was introduced in 44b892242d0162d3a5a53e47df290fcba597241d in December 2016. Comments in [file.py at that commit ](https://github.com/NeurodataWithoutBorders/pynwb/blob/44b892242d0162d3a5a53e47df290fcba597241d/src/pynwb/ui/file.py) indicate that this was intended to represent same info as session_start_time:

```
*start_time* (text -- optional) This is the starting time of the 
experiment.  If this isn't provided, the start time of the 
experiment will default to the time that the file is created
```

The changes in Pull request 426 should be reapplied to the correct field, (as @bendichter noted in his comments there after the merge)

[Should this be picked up by a test? I.e. if fields are present that aren't in the spec?]

bendichter:
Thanks for tracking that down, @tjd2002! 

> Should this be picked up by a test? I.e. if fields are present that aren't in the spec?

I don't think so. I think there are situations where you would want the ability to input fields that are not in the spec, for instance intermediate variables, or specific instructions for how to write something to disk.

jeffteeters:
> Should this be picked up by a test? I.e. if fields are present that aren't in the spec?
> 
> > I don't think so. I think there are situations where you would want the ability to input fields that are not in the spec, for instance intermediate variables, or specific instructions for how to write something to disk.

If it's allowed to include content in the file which are not in the spec, it might be good to have a convention so that such content is easily distinguished from content that is in the spec.   NWB version 1.x allowed such "custom" content in the files, and required that it be flagged using attribute "neurodata_type" with value "Custom".  But my understanding is that "neurodata_type:Custom" is no longer supported in PyNWB.  Is the plan for PyNWB to allow custom content without requiring that it be flagged as custom?

bendichter:
@jeffteeters To clarify, we are talking about an argument into a pynwb class that never gets written to the NWB file. We plan on keeping it that you must explicitly define all data that gets written to an NWB file. 

In this case `start_time` is defined as an attribute to `NWBFile` but is not in the spec and never gets written to the file. It does nothing, and it looks like it's there by mistake. The question (as I understand it) was whether we should enforce that all data-writing python classes only take arguments that are defined by the spec in order to catch such mistakes.

oruebel:
> We plan on keeping it that you must explicitly define all data that gets written to an NWB file.

Agreed!

> whether we should enforce that all data-writing python classes only take arguments that are defined by the spec

@bendichter  I agree that  while this probably true for most classes right now, that this something we don't want make a rule. For example, instead of having to provide some particular field directly you might want to have the API compute the information to be stored based on some other inputs. As such, the API should be allowed to be different from the storage spec. You could certainly write some generic test to check this but then add it to the test suite for the container classes on a case-by-case basis.

oruebel (approved):

PR #477, opened 2018-04-21
fix manifold shape

bendichter:
Change description and test of manifold to reflect the shape in the schema.

* Change the shape of the test manifold to (50,100,3) to reflect the spec (null, null, 3)
* Change the documentation to say "heigh","width","xyz"


ajtritt (approved):

ajtritt:
addressed in #482

PR #478, opened 2018-04-21
names cannot include '/'

### LINKED ISSUE #129: pynwb mangles names with "/"
on Python 2.7

I created an RoiResponseSeries and named it "dF/F"

https://gist.github.com/neuromusic/b44fa185f03b7e588af7f5e7ccb52bdd#file-ophys_example-py-L82

I can't validate the created NWB file because of #127, however, when opening with h5py, the object gets split into a hierarchy of groups "dF" & "F"

image

bendichter:
The '/' character causes a problem for names of datasets because h5py interprets and creates subgroups, which is not the desired behavior.

Add ValueError in constructor for all data objects if '/' is present
Add test that '/' raises error

## Motivation

fix #129 

## How to test the behavior?
```python
from pynwb import NWBContainer
NWBContainer('a', 'b/c')
```
should raise: `ValueError: name 'b/c' cannot contain '/'`

## Checklist

### DIFF
--- a/src/pynwb/core.py
+++ b/src/pynwb/core.py
@@ -101,7 +101,10 @@ class NWBBaseType(with_metaclass(ExtenderMeta)):
         super(NWBBaseType, self).__init__()
         self.__fields = dict()
         self.__parent = None
-        self.__name = getargs('name', kwargs)
+        name = getargs('name', kwargs)
+        if '/' in name:
+            raise ValueError("name '" + name + "' cannot contain '/'")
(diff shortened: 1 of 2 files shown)


ajtritt (approved):

neuromusic:
💯

ajtritt (changes requested):
This should now go in Container, in src/pynwb/form/container.py

ajtritt:
addressed in #482

PR #479, opened 2018-04-21
change ophys wavelength from str to float

bendichter:
depends on https://github.com/NeurodataWithoutBorders/nwb-schema/pull/137, which addresses https://github.com/NeurodataWithoutBorders/nwb-schema/issues/136

Tests not passing due to Builder issues. I could use a hint @ajtritt 

## Checklist


ajtritt (commented):
Don't forget to change the schema.
Here:
https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/data/nwb.ophys.yaml#L230

And here:
https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/data/nwb.ophys.yaml#L286

bendichter:
@ajtritt Yeah see above. I referenced a schema PR that needs to occur simultaneously. Is the builder using the schema? Is that where the errors are coming from?

ajtritt:
Oops, didn't read the PR description. You'll have to change it to get the tests running. 

I've concluded that based on this part of the error message: 

```
  File "/home/circleci/project/src/pynwb/form/build/map.py", line 675, in construct
    obj = cls(**kwargs)
  File "/home/circleci/project/src/pynwb/form/utils.py", line 346, in func_call
    raise_from(TypeError(msg), None)
  File "<string>", line 3, in raise_from
TypeError: incorrect type for 'emission_lambda' (got 'str', expected 'float')
```

This tells me wrong type is getting passed into the constructor.  In the test suite, there are two ways the constructor gets called: 1) when you explicitly call it in the tests and 2) when the middle layer builds the objects that get read in. We know the the first way shouldn't raise this error, since we can look at the code. So, it must be the second way. When the *emission_lamba* gets read in, it's string. You gave it a float, so at some point along the way, it got converted to string, either before or after it was written. 

It happens before it's written. I know this because I can look at the results of `test_build`.

```
...
-               'emission_lambda': {'attributes': {}, 'data': '500.0'},
?                                                             -     -

+               'emission_lambda': {'attributes': {}, 'data': 500.0},
...
```

The handwritten Builder had a float (you just changed those), but the Builder produced by the middle layer had strings. So, the middle layer converted the floats to strings. 

Why would such a thing happen? Because the specification says *emission_lambda* should be `text`, it converts it from a float to string. The code that does this is [here](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/form/build/map.py#L380)

Hope that helps.

bendichter:
Ahhh thanks! @ajtritt 

So what's the best way to proceed in these cases? As soon as we merge the nwb-schema changes, there will probably be incompatibilities between the nwb-schema/dev and pynwb/dev until we merge these changes. Is there a way to handle this so that we are never out of sync? Is there a way of changing which nwb-schema branch is used in the pynwb test? 

@dorukozturk?

ajtritt:
Just make the changes you need in pynwb, and then copy over the changes to nwb-schema. This is how I've been operating.

mgrauer:
I'm not addressing the CI side of this, but part of the story here is to version the schema and version PyNWB with more frequent minor versions, and keep track of which versions go together (even if just in documentation).

ajtritt:
addressed in #482

PR #481, opened 2018-04-23
Run tests against miniconda on circleci

dorukozturk:
## Motivation

We were only checking installation on miniconda. Now we also run the tests against it.


ajtritt (approved):
Thanks for the speedy fix!

dorukozturk:
Sure!!! Thanks for catching it.

PR #484, opened 2018-04-24
sphinx-gallery tutorial for brain observatory files first version

### LINKED ISSUE #483: Ophys brain observatory tutorial to sphinx-gallery
Add a tutorial based on writing Allen Institute Brain Observatory data.

nicain:
## Motivation

Resolves #483 by adding a tutorial example

### DIFF
--- a/docs/gallery/domain/brain_observatory.py
+++ b/docs/gallery/domain/brain_observatory.py
@@ -1,0 +1,187 @@
+"""
+Allen Brain Observatory
+=================================
+
+Create an nwb file from Allen Brain Observatory data.
+"""
+
+########################################
+# This example demostrates the basic functionality of several parts of the pynwb write API, centered around the optical physiology submodule (pynwb.ophys). We will follow and extract/transform/load (ETL) methodology, using the allensdk as a read API while leveraging the pynwb data model and write api to transform and write the data back to disk. # noqa: E501
+
+########################################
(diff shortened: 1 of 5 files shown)


docs/gallery/domain/brain_observatory.py
```diff
@@ -0,0 +1,187 @@
...
+import allensdk.brain_observatory.stimulus_info as si
+
+from pynwb import NWBFile, NWBHDF5IO, TimeSeries
+from pynwb.ophys import OpticalChannel, DfOverF, ImageSegmentation
+from pynwb.image import ImageSeries, IndexSeries
+
+
+# Settings:
+ophys_experiment_id = 562095852
+save_file_name = 'basic_example.nwb'
```
ajtritt: Can you change this name? Another tutorial already has this name, I think. In the event that someone wants to run the test suite and view the files, one of these would get clobbered.

nicain: Aha, OK, will do.  I am also working on getting this thorugh some sort of sane testing, since allensdk is a dependency for getting this to work :/

nicain:
@dorukozturk The sphinx gallery example that I added sucked in a tricky dependency: allensdk.  Additionally, --gallery was turned on in test.py by default.  It makes more sense to have a separate test case for the sphinx-gallery examples, which I have built by creating a new circleci build, and subsequent tox.ini environment, for testing just the gallery examples in isolation.  Can you take a peek at my changes to this testing infrastructure, and let me know if you have any suggestions?

docs/gallery/domain/brain_observatory.py
nicain: OK, this is done.

nicain:
OK, downloading an NWB file for testing the brain_observatory.py tutorial script takes too long, and circle is timing out.  For now, I am going to blacklist this script.

dorukozturk:
@nicain This looks very good. I have 3 questions/comments. 
1) We can cache the nwb file on circle so that we don't download it unless it changes
2) Do we want to test gallery only on python3.6?
3) We should ignore manifest.json file so that somebody does not commit it by accident

I will submit a PR for these issues and maybe you review it?

PR #486, opened 2018-04-25
Cache allen sdk files on circle for speeding up tests

dorukozturk:
## Motivation

Caches big data file so that circle does not have to download it every time. This will speed up gallery tests significantly. Also we run the gallery tests for all 3 python versions now. @nicain that would be great if you can review it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,6 +83,24 @@ conda-steps: &conda-steps
         command: |
           tox -e $TEST_WHEELINSTALL_ENV --recreate --installpkg dist/*-none-any.whl
 
+gallery-steps: &gallery-steps
+  steps:
+    - checkout
+    - restore_cache:
+        keys:
+          - ophys-data-cache
+    - run:
+        <<: *initialize-venv
+    - run:
+        name: Run the gallery tests
+        command: |
+          . ../venv/bin/activate
+          pip install tox
+          tox -e $TEST_TOX_ENV
+    - save_cache:
+        key: ophys-data-cache
+        paths: ophys_experiment_data
+
 version: 2
 jobs:
   flake8:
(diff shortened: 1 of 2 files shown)


nicain (approved):
LGTM!

.circleci/config.yml
```diff
@@ -83,6 +83,24 @@ conda-steps: &conda-steps
...
+    - run:
+        <<: *initialize-venv
+    - run:
+        name: Run the gallery tests
+        command: |
+          . ../venv/bin/activate
+          pip install tox
+          tox -e $TEST_TOX_ENV
+    - save_cache:
+        key: ophys-data-cache
```
jcfr, line 101: I suggest to include the revision. The problem with this approach is that the cache will never be invalidated.

Consider that the prefix `ophys-data-cache` has already been used, you could instead use `gallery-test-{{ .Revision }}`

```
- restore_cache:
    keys: gallery-test-

[...]

- save_cache:
    key: gallery-test-{{ .Revision }}
```

From  https://circleci.com/docs/2.0/caching/#clearing-cache : 

> Tip: Caches are immutable so it is useful to start all your cache keys with a version prefix, for example v1-.... This 
> enables you to regenerate all of your caches by incrementing the version in this prefix.

dorukozturk: Oh ok, I see your point. The problem with using revision will be downloading it way too frequently. It takes ~12 minutes to download the file. Do you think it is better to cache based on latest git tag like (gallery-test-0.3.0)? This way we also ensure the cached file will work for a specific version. 

So if the allensdk file test is broken that means we have to bump the number so that the cache will have a chance to get invalidated? If the test is broken at some point that means either we broke backwards compatibility or it is time to replace that file. Either way it is good to invalidate cache at that point by changing the key. When I say invalidate I mean using a different cache key. Does that make sense?

jcfr: The idea is the following:
* Use revision for caching. Then specifying `gallery-test-` ensure the latest available version is restored and will avoid unwanted download. See [Restoring Cache](https://circleci.com/docs/2.0/caching/#restoring-cache)
* To ensure the updated version is effectively downloaded, checksum of the current version (restored from cache) should be computed and compared with the expected one ... if not the expected one, the data would effectively be re-downloaded and cached.

PR #492, opened 2018-04-26
Ignore brain_observatory file's flake8 errors

### LINKED ISSUE #491: flake8 config not ignoring gallery example
In https://github.com/NeurodataWithoutBorders/pynwb/commit/72883b017b0eb0ae4fb3e37630db09b5c2082ab5 I added a line to ignore line-too-long (i.e. E501) to the flake8 circle build.  However, it looks like this ignore is not actually working, for some reason.  I had to force-merge PR https://github.com/NeurodataWithoutBorders/pynwb/pull/490 to get my changes for my tutorial through for the hackathon, but this needs to be sorted out. @dorukozturk am I missing something obvious?

As a side note, I need this long line so that sphinx gallery can find the natural line break to use for typesetting the documentation in the html and jupyter notebooks that it creates; I could tell it where to line break, but it is better to let those platforms figure it out for themselves depending on monitor resolution, etc.  If I add noqa inline at the end of the line, it gets rendered inside the documentation as plaintext :/

dorukozturk:
## Motivation

Fixes #491

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ run-style-check: &run-style-check
     pip install flake8
     flake8 src/pynwb
     flake8 tests
-    flake8 --ignore E402 --ignore E501 docs/gallery
+    flake8 --ignore E402,E501 docs/gallery
 
 initialize-venv: &initialize-venv
   name: Initialize Virtual Environment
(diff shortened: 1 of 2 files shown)


nicain (approved):
Thanks!

PR #493, opened 2018-04-26
[DON'T MERGE] Introducing ci

dorukozturk:


ajtritt:
@dorukozturk Does this need to be open anymore?

dorukozturk:
@ajtritt Sure we can close it. It was only for instructional purposes.

PR #496, opened 2018-04-27
Add codecov config file for better defaults

dorukozturk:
## Motivation

This PR adds to conditions. 

1) We make the coverage checks fail if a PR adds a lot of code without tests so that it drops overall coverage by 2%. 

2) We make the coverage checks fail if a PR adds code but covers less than 75% of the changed lines.

We can always adjust these numbers if they become annoying. We use these numbers in some of our projects and they have been serving well so far.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -2,9 +2,11 @@ coverage:
   status:
     project:
       default:
-        target: auto
-        threshold: 1%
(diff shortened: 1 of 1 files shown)


jcfr:
@dorukozturk I missed your PR.  I am rebased your topic

Issue #502, opened 2018-04-27
NWB Extensions tutorial has incorrect argument order in call to NWBAttributeSpec


jeffteeters:
In the NWB Extensions tutorial, at:
https://pynwb.readthedocs.io/en/latest/tutorials/general/extensions.html#sphx-glr-tutorials-general-extensions-py

Parameters to NWBAttributeSpec should be:

`pynwb.spec.NWBAttributeSpec(name, doc, dtype, shape=None, dims=None, required=True, parent=None, value=None, default_value=None)`
(According to the API specification at:
http://pynwb.readthedocs.io/en/latest/pynwb.spec.html?highlight=NWBAttributeSpec)

But the call in the tutorial is:

`NWBAttributeSpec('trode_id', 'int', 'the tetrode id')`

(The doc and dtype are reversed).

This results in the following yaml file (mylab.extensions.yaml) being generated:

```
groups:
- attributes:
  - doc: int
    dtype: the tetrode id
    name: trode_id
  doc: A custom ElectricalSeries for my lab
  neurodata_type_def: TetrodeSeries
  neurodata_type_inc: ElectricalSeries
```

With the `doc` & `dtype` being reversed.

PR #503, opened 2018-04-27
Fix NWB Extensions tutorial order of arguments in call to NWBAttributeSpec

jeffteeters:
This fixes issue issue #502 

(Corrects the order of the arguments in the call to NWBAttributeSpec in the tutorial at:
https://pynwb.readthedocs.io/en/latest/tutorials/general/extensions.html#sphx-glr-tutorials-general-extensions-py ).


mgrauer:
@dorukozturk 

This PR failed the coverage test with a decrease of 0.04%, which is below the threshold of 2%, but the test failed. Can you take a look at the coverage threshold tests?

oruebel:
This PR just fixes a bug in a tutorial so the problem with the codecov test is not critical (i.e., the problem seems to be with codecov not the PR).

PR #506, opened 2018-04-28
Add script to update schema files and update files

jcfr:


jcfr:
Cc: @bendichter @nicain @mgrauer @dorukozturk @oruebel @ajtritt

src/pynwb/data/UpdateFromUpstream.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+thirdparty_module_name='nwb-schema'
+
+upstream_git_url='git://github.com/NeurodataWithoutBorders/nwb-schema'
+upstream_git_branch='dev'
+github_compare=true
+
+snapshot_author_name='nwb-schema Upstream'
+snapshot_author_email='neurodatawithoutborders@googlegroups.com'
```
jcfr, line 10: @oruebel I suggest we associate the email `neurodatawithoutborders@googlegroups.com` with @nwb-bot

jcfr:
@Lawrence I am planning to update matnwb adding a script similar to [src/pynwb/data/UpdateFromUpstream.sh](https://github.com/NeurodataWithoutBorders/pynwb/pull/506/files)

src/pynwb/data/UpdateFromUpstream.sh
jcfr: @oruebel If this sounds reasonable, is it something you could do ?

oruebel: @jcfr what is the purpose of linking nwb-bot to the googlegroup?

jcfr: This allows to have commit message done by the bot to have an email, allowing it to link back to the bot account.

The email doesn't have to be verified, it should just be added in the settings.

rly:
Our current workflow is to make changes to the schema in pynwb and then copy those over to nwb-schema. This PR would add a script to do the reverse - it copies the schema from nwb-schema over to pynwb. As such, we are closing this PR. We could change the script to do the reverse if this is desired.

Issue #520, opened 2018-05-26
trouble writing extensions with str type


bendichter:
update:

```python
name = 'test_str'
ns_path = name + ".namespace.yaml"
ext_source = name + ".extensions.yaml"


test_type = NWBGroupSpec(neurodata_type_def='TestType', datasets=[
    NWBDatasetSpec(name='str_type', dtype='text', doc='genotype',
                   quantity='?')
], doc='doc')


ns_builder = NWBNamespaceBuilder(name + ' extensions', name)
ns_builder.add_spec(ext_source, test_type)
ns_builder.export(ns_path)


from pynwb import load_namespaces, get_class

load_namespaces(ns_path)
test_type = get_class('TestType', name)

test_type(name='test_name', str_type='hello')
```

error:
```python
Traceback (most recent call last):
  File "/Users/bendichter/Library/Preferences/PyCharmCE2019.1/scratches/scratch_17.py", line 24, in <module>
    test_type(name='test_name', str_type='hello')
  File "/Users/bendichter/dev/hdmf/src/hdmf/utils.py", line 379, in func_call
    raise_from(ExceptionType(msg), None)
  File "<string>", line 3, in raise_from
TypeError: incorrect type for 'str_type' (got 'str', expected 'ndarray, list, tuple, Dataset, AbstractDataChunkIterator, HDMFDataset or DataIO')

```

bendichter:
@ajtritt

oruebel:
@bendichter try to change ``dtype='str'`` to ``dtype='text'``. I think ``str`` was removed when we cleaned up the specification language. 

BTW, generally I would discourage using datasets to store single single strings and small fixed-size data in general. While it's perfectly Ok to it, in HDF5 Attributes are typically the better choice for this kind of small data.

bendichter:
Thanks!

On May 28, 2018 1:57 AM, "Oliver Ruebel" <notifications@github.com> wrote:

@bendichter <https://github.com/bendichter> try to change dtype='str' to
dtype='text'. I think str was removed when cleaned up the specification
language.

BTW, generally I would discourage using datasets to store single single
strings and small fixed-size data in general. While it's perfectly Ok to
it, in HDF5 Attributes are typically the better choice for this kind of
small data.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/NeurodataWithoutBorders/pynwb/issues/520#issuecomment-392389149>,
or mute the thread

[...]

PR #523, opened 2018-05-29
Enh/validatedtype

### LINKED ISSUE #521: NWBAttributeSpec dtype takes any string
I looks like `NWBAttributeSpec` will accept any `str` as the dtype argument. `pynwb.spec.NWBAttributeSpec(name='my_name', dtype='sdfsd', doc='my doc')` should throw an error.

oruebel:
## Motivation

This pull request fixes several issues:

* Fix #521 . Add validation of dtype strings for ``AttributeSpec`` and ``DatasetSpec``.  This is in part also in response to #520 
* ``nwb.misc.yaml`` specified ``dtype: None`` which is not allowed. Assigned appropriate dtype values instead.
* The following spec files contained `` dtype: number`` which is not allowed. Assigned approbriate dtype values instead:``src/pynwb/data/nwb.behavior.yaml``, ``src/pynwb/data/nwb.ecephys.yaml``, ``src/pynwb/data/nwb.icephys.yaml``, ``src/pynwb/data/nwb.image.yaml``
* Several tests used ``dtype='str'`` in the specification of attributes/datasets. Fixed those tests to use ``dtype='text'`` instead. 
* Updated the validator to use the data type lists from the new ``DtypeHelper`` class to avoid duplication of code 
* Added tests for the ``DtypeHelper`` class and updated constructor behavior of AttributeSpec and DatasetSpec
* Bug fix. ``shape`` constructor argument was not set in ``AttributeSpec``
* Updated ``DtypeSpec.is_ref`` function to also check if a ``DtypeSpec`` defines a reference
* Add some more test cases for the various spec classes

## Checklist

### DIFF
--- a/src/pynwb/data/nwb.icephys.yaml
+++ b/src/pynwb/data/nwb.icephys.yaml
@@ -8,7 +8,7 @@ groups:
   - dims:
     - num_times
     doc: Recorded voltage or current.
-    dtype: number
+    dtype:
     name: data
     shape:
     - null
--- a/src/pynwb/data/nwb.icephys.yaml
+++ b/src/pynwb/data/nwb.icephys.yaml
@@ -8,7 +8,7 @@ groups:
   - dims:
     - num_times
     doc: Recorded voltage or current.
-    dtype:
+    dtype: float
     name: data
     shape:
     - null
--- a/src/pynwb/data/nwb.behavior.yaml
+++ b/src/pynwb/data/nwb.behavior.yaml
@@ -20,7 +20,7 @@ groups:
     - num_times
     - num_features
     doc: 2-D array storing position or direction relative to some reference frame.
-    dtype: number
+    dtype: float
     name: data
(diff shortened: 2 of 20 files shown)


src/pynwb/data/nwb.icephys.yaml
```diff
@@ -8,7 +8,7 @@ groups:
   - dims:
     - num_times
     doc: Recorded voltage or current.
-    dtype: number
+    dtype:
```
ajtritt: Is this supposed to be an empty dtype?

oruebel: Oops, that should be float. I'll fix that and also add some more tests to up the coverage for spec.py

bendichter:
Woah, this is great! Thanks for working on this so quickly! :-) :-)

@oruebel @ajtritt

ajtritt (approved):

PR #573, opened 2018-07-30
Create a separate file for the Device object

### LINKED ISSUE #443: Device class used for ecephy, icephys and ophys but lives in ecephys.py
The class `Device` is used for Intracellular Ephys and Extracellular Ephys. See https://github.com/NeurodataWithoutBorders/pynwb/blob/3afc38fce1cad67c54efa55068c01486ff15478f/src/pynwb/data/nwb.ecephys.yaml#L312 and https://github.com/NeurodataWithoutBorders/pynwb/blob/3afc38fce1cad67c54efa55068c01486ff15478f/src/pynwb/data/nwb.icephys.yaml#L194.

But it lives in https://github.com/NeurodataWithoutBorders/pynwb/blob/3afc38fce1cad67c54efa55068c01486ff15478f/src/pynwb/ecephys.py#L14.

t-b:
The Device object is currently defined in ecephys.py although both
icephys and ecephys use it. This is unfortunate and will be changed in
this commit.

Fix #443.

### DIFF
--- a/docs/gallery/domain/ecephys.py
+++ b/docs/gallery/domain/ecephys.py
@@ -45,7 +45,7 @@ nwbfile = NWBFile('the PyNWB tutorial', 'my first synthetic recording', 'EXAMPLE
 device = nwbfile.create_device(name='trodes_rig123', source="a source")
 
 #######################
-# Once you have created the :py:class:`~pynwb.ecephys.Device`, you can create an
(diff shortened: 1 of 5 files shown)


bendichter:
Thanks for tackling this, @t-b!

bendichter (approved):

PR #574, opened 2018-07-30
add MultiContainerInterface to tutorials

bendichter:
I had some trouble with this so I am providing a guide for others.

### DIFF
--- a/docs/gallery/general/extensions.py
+++ b/docs/gallery/general/extensions.py
@@ -291,8 +291,6 @@ class Potato(NWBContainer):
 @register_class('PotatoSack', name)
 class PotatoSack(MultiContainerInterface):
 
-    __nwbfields__ = ('potatos',)
-
     __clsconf__ = {
         'attr': 'potatos',
         'type': Potato,
--- a/docs/gallery/general/extensions.py
+++ b/docs/gallery/general/extensions.py
@@ -273,6 +273,9 @@ ns_builder.export(ns_path)
 ####################
 # Then create DataMapper classes registered to the new data types (this is generally done in a new file)
 
+from pynwb import register_class, load_namespaces
+from pynwb.file import MultiContainerInterface, NWBContainer
+
 load_namespaces(ns_path)
 
 @register_class('Potato', name)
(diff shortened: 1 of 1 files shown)


docs/gallery/general/extensions.py
```diff
@@ -225,3 +225,93 @@ def __init__(self, **kwargs):
...
+                                                dtype='text',
+                                                value="It's a sack of potatoes")
+                           ])
+
+ns_builder = NWBNamespaceBuilder(name + ' extensions', name)
+ns_builder.add_spec(ext_source, potato_sack)
+ns_builder.export(ns_path)
+
+####################
+# Then create DataMapper classes registered to the new data types (this is generally done in a new file)
```
ajtritt: Your use of *DataMapper* here suggests this is a construct in the API. I'm afraid this will confuse users. Can you change the language ?

bendichter: Sure, should I just call it a class?

docs/gallery/general/extensions.py
```diff
@@ -225,3 +225,93 @@ def __init__(self, **kwargs):
...
+
+load_namespaces(ns_path)
+
+@register_class('Potato', name)
+class Potato(NWBContainer):
+    __nwbfields__ = ('name', 'weight', 'age', 'source')
+
+    def __init__(self, **kwargs):
```
ajtritt: You could also demonstrate the use of docval/getargs/call_docval_func. Not a change I'm requesting, but thought I would mention it.

ajtritt (commented):
Please remove the use of *DataMapper* ( see comment on line 274

docs/gallery/general/extensions.py
ajtritt: *Container class* would be good. That's how we were refer to this layer in the software architecture overview: http://pynwb.readthedocs.io/en/latest/overview_software_architecture.html#main-concepts

docs/gallery/general/extensions.py
bendichter: Yeah, might as well show best practices

docs/gallery/general/extensions.py
bendichter: ok, will do

PR #578, opened 2018-07-31
fix ImageSeries dimension default value

### LINKED ISSUE #528: ImageSeries dimension default arg
The default value of `dimension` in `ImageSeries` in `[np.nan]`, however `dimension` is of type `int32`, so it gets saved as `-2147483648`. I suppose we could leave it like this, but it's a bit confusing. Maybe the default value should be `[-1]`?

bendichter:
## Motivation

The default value of `dimension` in `ImageSeries` was `[np.nan]`, however `dimension` is of type `int32`, so it got saved as `-2147483648`. Default value is now changed to `[-1]`.

fix #528 

## Checklist


t-b:
Looks good. If I read https://github.com/NeurodataWithoutBorders/nwb-schema/blob/997b52423ce242f62cd861c948731ba57aea5738/core/nwb.image.yaml#L35 correctly we don't store the default value of dimension in the schema.

bendichter:
actually you just pointed out a discrepancy. Default val should be `None`.

ajtritt (approved):

PR #580, opened 2018-08-01
add args check to timeseries

### LINKED ISSUE #579: pynwb.base.TimeSeries.starting_time & rate does not return the floats given in the argument
## 1) Bug
I am able to create a TimeSeries object but while checking if the object was what I intended to create I noticed something weird.
If `ts = TimeSeries (...)` (as shown below), I can get the exact **float** values I gave in the argument for `ts.resolution` and `ts.conversion`.
But for `ts.starting_time` and `ts.rate`, they do not return floats instead they return **None**; which is the pynwb default value.

### Steps to Reproduce
```
from pynwb import TimeSeries

ts = TimeSeries( name = "dummy",
                 source = "soma",
                 data = list(range(0, 100, 10)),
                 unit = "milliVolt",
                 resolution = 0.01,
                 conversion = 1000.0,
                 timestamps = list(range(10)),
                 starting_time = 0.0,
                 rate = 1/0.01 )

>> ts.resolution # same for ts.conversion
0.01

>> type(ts.conversion) # same for ts.resolution
<type 'float'> 

>> ts.starting_time # same for ts.rate
None

>> type(ts.rate) # same for ts.starting_time
<type 'NoneType'>
```
### Environment
Please describe your environment according to the following bullet points.
- **Python Executable:** Python
- **Python Version:** Python 2.7
- **Operating System:**  Ubuntu 16.04
- **Pynwb Version:**  0.4.1

bendichter:
## Motivation

`TimeSeries` had bad behavior when you enter `rate` and `timestamps`. A check is added that throws an error if both are provided.

fix #579 

## How to test the behavior?
```python
TimeSeries('test_ts2', 'unit test test_data_timeseries', [10, 11, 12, 13, 14, 15], 'grams', rate=30.,
                       timestamps=[.3, .4, .5, .6, .7, .8])
```

## Checklist


src/pynwb/base.py
```diff
@@ -159,6 +159,8 @@ def __init__(self, **kwargs):
         timestamps = kwargs.get('timestamps')
         starting_time = kwargs.get('starting_time')
         rate = kwargs.get('rate')
+        if rate is not None and timestamps is not None:
+            raise ValueError('specify rate[+starting_time] OR timestamps, not both')
```
t-b: @bendichter How about checking starting_time as well?

```python
if timestamps is not None:
    if rate is not None:
         raise ValueError('Specifying rate and timestamps is not supported.')
    elif starting_time is not None:
         raise ValueError('Specifying starting_time and timestamps is not supported.')
```

and then we need one more test as well.

t-b: @bendichter And if we code it up like that we should also change the schema as https://github.com/NeurodataWithoutBorders/pynwb/blob/5ac6603fb4e9c570dd1d32a627138d75729e7f89/src/pynwb/data/nwb.base.yaml#L139-L141 says that rate is required.

bendichter:
Yes we should. Good catch.

t-b:
@bendichter Looks good! I would now squash all commits into one but that is purley a matter of taste.

bendichter:
Ah yes I've been meaning to learn how to do that

t-b:
@bendichter Nothing easier than that

```
git fetch --all
git checkout $this_branch
git branch bak # in case it goes wrong
git rebase -i origin/dev
# mark all commits, except the first as `fixup` instead of `pick`
git commit --amend --verbose # write a good commit message and be able to view the diff
git push --force-with-lease
```

src/pynwb/base.py
bendichter: Now that I look at it again, the format of the HDF5 is that `starting_time` is a value and `rate` is a mandatory attribute. I think the logic was so that `starting_time` and `rate` go hand-in-hand. I would personally prefer it to be reversed and have `starting_time` be an attribute of `rate`, since `starting_time` is usually `0.0` and unimportant and `rate` is always important. Nevertheless I believe `rate` is intentionally required here.

bendichter:
@t-b Actually I think this is the easiest way (for me). Github ftw

image

ajtritt (approved):

t-b:
@bendichter This would not create a merge commit. Also in https://github.com/NeurodataWithoutBorders/pynwb/commit/b24ace0b712f72ac612cb406a3229bf9adaf406a this is now one commit, instead of the two I carefully crafted.

bendichter:
Ah whoops I guess I was overzealous with the squashing

PR #581, opened 2018-08-01
TimeSeries: Rename rate_unit variable and use it correctly

t-b:
Since 575da105 (moved base TimeSeries class into submodule core,
2017-01-30) the unit of the rate is defined as "Seconds". But in
TimeSeriesMap it is used as starting time unit.

Rename the class variable therefore from rate_unit to
starting_time_unit.

A user visible change is that we now fill starting_time_unit only when a
starting_time is given.


ajtritt (approved):
Good catch

PR #582, opened 2018-08-02
Prefer device over device name and use links

### LINKED ISSUE #412: ImagingPlane device argument should be link to device, not str
https://github.com/NeurodataWithoutBorders/pynwb/blob/ddeaacc464e567f3fc123ca74a6d8cecbfbbce15/src/pynwb/ophys.py#L59

Correct me if I'm wrong, but I think for consistency with ephys, the device argument of ImagingPlane should expect a Device type, not a str.

t-b:
Fixes #412.

I must admit I cargo culted the schema changes and the tests.


t-b:
@nicain I hope the brain observatory changes are good?

bendichter:
phew such a small change touches so many files! Thanks for grinding through this. Looks good to me. It would be nice to give a deprecation-style message to users who try to use a string explaining the change, but I'm not sure how to do that with docval.

bendichter (approved):

t-b:
@bendichter Thx.

PR #585, opened 2018-08-07
fixed spelling

engnadeau:
## Motivation
Spelling is important :)

## How to test the behavior?
```
```

## Checklist


dorukozturk:
@nnadeau You should add a space here to match the indentation with the top line. https://github.com/NeurodataWithoutBorders/pynwb/blob/d79109b0762b0fa5cbe38399c456784b11cb2374/docs/gallery/general/extensions.py#L330

docs/gallery/domain/ophys.py
```diff
@@ -243,4 +243,4 @@
 # .. [#] If you added more than one :py:class:`~pynwb.ophys.RoiResponseSeries`, you will need to
 #    provide the name of the :py:class:`~pynwb.ophys.RoiResponseSeries` you want to retrieve to
 #    :py:func:`~pynwb.ophys.Fluorescence.create_roi_response_series`. This same behavior is exhibited
-#    with :py:class:`~pynwb.ophys.ImageSegmentation` and various other objects throught the PyNWB API.
+#    with :py:class:`~pynwb.ophys.ImageSegmentation` and various other objects thought the PyNWB API.
```
bendichter: through

docs/gallery/general/extensions.py
```diff
@@ -310,7 +310,7 @@ def __init__(self, **kwargs):
 class PotatoSack(MultiContainerInterface):
 
     __clsconf__ = {
-        'attr': 'potatos',
+        'attr': 'potatoes',
```
bendichter: actually this was intentional. I know it's misspelled but the api expects adding an 's'.

docs/gallery/general/extensions.py
```diff
@@ -326,8 +326,8 @@ class PotatoSack(MultiContainerInterface):
 from pynwb import NWBHDF5IO, NWBFile
 
 potato_sack = PotatoSack(source='pantry',
-                         potatos=Potato(name='potato1', age=2.3, weight=3.0,
-                                        source='the ground'))
+                         potatoes=Potato(name='potato1', age=2.3, weight=3.0,
+                                         source='the ground'))
```
bendichter: This was also intentional. This is the way you need to write it to get the api to work. We should really file an issue about this.

engnadeau: So, the `PotatoSack` class builds its attributes using `__clsconf__` (why not use [`attrs`](https://github.com/python-attrs/attrs)?)

My initial PR changes broke your CI due to the mismatch between the `__clsconf__` attribute and the actual keyword assignment. The latest version of the PR fixed the spelling in both places.

What is the current intentional behaviour? All the tests pass, but is there something not being tested?

bendichter: Yes, I think there is something not being tested. As written now I believe it will create an orphaned dataset, but catching it will require reading in the file created in the demo. That would help for pedagogical completeness as well, so I'll contribute a PR adding that to the demo.

bendichter:
@nnadeau I have added code to reveal the "potato" problem. Could you please remove the "potato"-related edits so we can incorporate all the others? I agree that the "potato" thing is a problem, and I'm going to open a specific issue about it, since it will require additional effort.

bendichter (approved):

Issue #312, opened 2018-01-03
enhanced repr


neuromusic:
### Problem/Use Case
Constructing NWBFiles in interactive environments like Jupyter would be facilitated by a more comprehensive __repr__ which displays the values of the attributes saved in the file

for example, printing an xarray DataArray or Dataset displays important information about the dimensions and metadata associated with the data object 

```
In [5]: data = xr.DataArray(np.random.randn(2, 3), coords={'x': ['a', 'b']}, dims=('x', 'y'))
In [6]: data
Out[6]: 
<xarray.DataArray (x: 2, y: 3)>
array([[-1.294524,  0.413738,  0.276662],
       [-0.472035, -0.01396 , -0.362543]])
Coordinates:
  * x        (x) <U1 'a' 'b'
Dimensions without coordinates: y
```

## Checklist

- [x] Have you ensured the feature or change was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues/) ?
- [x] Have you included a brief and descriptive title?
- [x] Have you included a clear description of the problem you are trying to solve?
- [x] Have you included a minimal code snippet that reproduces the issue you are encountering?

ajtritt:
@neuromusic Good idea. Can you take a stab at this?

PR #595, opened 2018-08-20
add repr for NWB classes. Use numpy to print datasets

bendichter:
## Motivation

The `__repr__` method makes it easier to work interactively for instance in a Jupyter notebook. This is carrying out #312 to be actually output as `__repr__` and fixing display of large datasets.
## How to test the behavior?
In a Jupyter notebook
```python
from pynwb.misc import IntervalSeries
import numpy as np

nwbfile = NWBFile("source", "a file with header data", "NB123A", '2018-06-01T00:00:00')
interval_series = IntervalSeries(data=list(np.random.randn(100,100)),
                                 timestamps=list(range(10000)),
                                 name='test_interval_series', source='source')
nwbfile.add_acquisition(interval_series)
nwbfile
```
old output:
```
<pynwb.file.NWBFile at 0x1075d36d8>
```
new output:
```
root <class 'pynwb.file.NWBFile'>
Fields:
  source:
    source
  acquisition:
    {'test_interval_series': test_interval_series <class 'pynwb.misc.IntervalSeries'>
    Fields:
      source:
        source
      resolution:
        0.0
      comments:
        no comments
      description:
        no description
      conversion:
        1.0
      unit:
        n/a
      data:
    [[ 0.38033257  1.13714832 -1.37388252 ... -1.05715949  2.07389766
       0.51760853]
     [-0.48956633 -0.01494456 -0.36479737 ...  0.37393639  1.30577923
      -0.66345507]
     [ 1.59800002  0.52469375 -0.10094651 ...  0.17999466 -0.12167376
       0.51656919]
     ...
     [ 0.37120138  0.4033322   0.90104058 ...  0.87949093 -0.12655733
      -0.72993745]
     [ 0.2963325   0.14295951 -1.37639021 ...  0.44836196 -0.56332384
      -2.12629588]
     [ 1.7443997   0.05677955  0.87893873 ...  1.71913344 -1.69006126
       0.04939431]]
      num_samples:
        100
      timestamps:
    [   0    1    2 ... 9997 9998 9999]
      timestamps_unit:
        Seconds
      interval:
        1
    }
  analysis:
    {}
  stimulus:
    {}
  stimulus_template:
    {}
  modules:
    {}
  devices:
    {}
  ec_electrode_groups:
    {}
  imaging_planes:
    {}
  ic_electrodes:
    {}
  ogen_sites:
    {}
  epoch_tags:
    set()
```

## Checklist


src/pynwb/core.py
```diff
@@ -60,6 +60,26 @@ def prepend_string(string, prepend='    '):
     return prepend + prepend.join(string.splitlines(True))
 
 
+def list2str(v):
+    return str(np.array(v)) + '\n'
+
+
+""" This works for 1D arrays but not 2D arrays
```
oruebel: Which function object (e.g,. function) does this comment belong to, it seems to be a bit out of place in the middle of LabelledDict class?

https://github.com/NeurodataWithoutBorders/pynwb/blob/c1a616fd12c9e0e56fa847cbe607bc7511098ef1/src/pynwb/core.py#L67-L80

oruebel (dismissed):
Other than the one question I added this looks fine to me.

src/pynwb/core.py
bendichter: this PR is a bit of a work-in-progress. That is code I was using to attempt to do the printing myself for datasets without converting them to numpy. It works well for 1D lists, but once you get into multidimensional arrays the problem gets a bit harder. I'd prefer to just convert to numpy on the fly whenever people request a print, since that library handles this problem very well, but I realize that's extra overhead we may not want. This code is my attempt to try to do it myself, and I didn't want to throw it away just yet in case the team wanted to do this the long way.

bendichter:
I think there could be room for improvement here. I can imagine if you had an entire dataset converted this repr would be too busy

bendichter:
@oruebel @ajtritt ok I changed some things to simplify the print-out. Does this still look ok?

src/pynwb/core.py
```diff
@@ -166,8 +144,27 @@ def __gather_nwbfields(cls, name, bases, classdict):
...
 
-    def __str__(self):
-        return nwb_repr(self)
+    def __repr__(self):
+        template = "{} {}\nFields:\n""".format(getattr(self, 'name'), type(self))
+        for k, v in iteritems(self.fields):
+            template += "  {}: {} \n".format(k, self.smart_str(v))
+        return template
+
+    def smart_str(self, v):
```
oruebel: Could you add a docstring for the ``smart_str`` function to describe what it does. Also should smart_str be a private or a public function, i.e., should to be renamed it __smart_str if it is only used internally

oruebel (dismissed):

oruebel (approved):

PR #596, opened 2018-08-20
TwoPhotonSeries input args

### LINKED ISSUE #527: TwoPhotonSeries data warning
TwoPhotonSeries should not throw warning`UserWarning: missing required attribute 'data' for 'object_name' warnings.warn("missing required attribute '%s' for '%s'" % (spec.name, builder.name))` when `format='external'`

bendichter:
* make TwoPhotonSeries data optional
* enforce in pynwb that either external_data or data is input
* add test

## Motivation

fix #527
## How to test the behavior?
```python
from pynwb.ophys import TwoPhotonSeries, OpticalChannel, ImagingPlane
from pynwb.device import Device

oc = OpticalChannel('test_name', 'test_source', 'description', 500.)
device = Device(name='device_name', source='device_source')
ip = ImagingPlane('test_imaging_plane', 'test source', oc, 'description', device, 600.,
                  'imaging_rate', 'indicator', 'location', (50, 100, 3), 4.0, 'unit', 'reference_frame')
TwoPhotonSeries('test_tPS', 'a hypothetical source', unit='unit', field_of_view=list(),
                imaging_plane=ip, pmt_gain=1.0, scan_line_rate=2.0,
                starting_frame=[1, 2, 3], format='tiff', timestamps=list())
```

```
ValueError: must supply either external_file or data to test_tPS
```

```python
TwoPhotonSeries('test_tPS', 'a hypothetical source', unit='unit', field_of_view=list(),
                imaging_plane=ip, pmt_gain=1.0, scan_line_rate=2.0,
                starting_frame=[1, 2, 3], format='tiff', timestamps=list(), data=list())
```
is ok

## Checklist


src/pynwb/data/nwb.ophys.yaml
```diff
@@ -116,6 +116,7 @@ groups:
     doc: Signals from ROIs
     dtype: float32
     name: data
+    quantity: '?'
```
ajtritt: Can you explain what this is for? I'm confused about why data would be optional for a structure that's supposed to hold time series data, or how it fits in with the rest of the changes.

bendichter: Correct me if I'm wrong, but I believe the idea is that data can either be stored in `data` or can be stored in `external_file`. If it's in an external file, then `data` should not be required. The changes to the API ensure that at least one of those is present. If this is correct, maybe I can add this to the documentation on nwb-schema to avoid confusion.

ajtritt: That's correct, but it looks like this change is on RoiResponseSeries, not ImageSeries. Did you mean to add this change around here? 
https://github.com/NeurodataWithoutBorders/pynwb/blob/96b7f081534f26ca1a8a3f3080a6faae914fe5f3/src/pynwb/data/nwb.image.yaml#L24

bendichter: Oh yes I must have meant that. I'll correct it. By the way, would there be an efficient way to change the order of these yaml files? Maybe I'm the only one, but I find alphabetical order very confusing.

ajtritt: You're not alone... it's objectively very annoying! Way back, I tried to to get ruamel to output things in a different order, and couldn't get it to work, so I punted. I think I remember someone making an issue, and adding the correct ruamel code to get the sensible ordering.... I'll see if I can find it and add it to NWB 2.0 milesone

ajtritt (changes requested):
move quantity change from RoiResponseSeries to ImageSeries

src/pynwb/data/nwb.ophys.yaml
bendichter: OK, glad we agree on that and it's good to hear that someone has made an attempt. If you end up finding it I can help usher it through.

bendichter: https://github.com/NeurodataWithoutBorders/nwb-schema/issues/139

ajtritt (dismissed):

bendichter:
These commits are a mess due to merging with the YAML reorder, but the diff is right: https://github.com/NeurodataWithoutBorders/pynwb/compare/dev...bendichter:two_photon_warning

PR #598, opened 2018-08-25
add stimulus_description

### LINKED ISSUE #564: Add method to attach the simulus/protocol name to PatchClampSeries
### Problem

The NWB schema documentation does currently not provide a way to attach the stimulus/protocol name to a PatchClampSeries.

Some existing users of NWBv1, e.g. http://celltypes.brain-map.org/experiment/electrophysiology/474626527, use a text dataset named `aibs_stimulus_description` to store that name. This dataset is labeled as `neurodata_type = Custom` in order to keep the NWB files schema compliant.

### Not viable alternative solutions

Encoding the stimset name in the group name itself is not a viable option. The reasons, taken from https://nwb-users.slack.com/archives/C5XKC14L9/p1532442382000078?thread_ts=1530802845.000026&cid=C5XKC14L9, are:
- The group name has to be unique so you have to attach some numbers to the stimulus name and on reading remove these numbers again. Basically doing a custom serializing/deserialization on the string.
- HDF5 objects names can not contain arbitrary characters ('/' being the most notable one) so you would need to introduce some mapping if you use forbidden characters in your stimset name.

### Proposed solution

- Schema: Add `stimulus_description` as required dataset to PatchClampSeries
- Code: Add `stimulus_description` to the `__init__` method of PatchClampSeries

If there is consent that this is something worth doing I can try to find time doing it.

ajtritt:
## Motivation

fix #564 

@t-b I made `stimulus_description` a required attribute. I had the *PatchClampSeries* constructor set `stimulus_description` to "NA" by default since it broke the fewest tests.  Feel free to change the that around. 

## Checklist

### DIFF
--- a/src/pynwb/data/nwb.icephys.yaml
+++ b/src/pynwb/data/nwb.icephys.yaml
@@ -8,9 +8,9 @@ groups:
     dtype: text
     doc: Value is 'Superclass definition for patch-clamp data'
     value: Superclass definition for patch-clamp data
-  - doc: the protocol/stimulus name for this patch-clamp dataset
+  - name: stimulus_description
     dtype: text
-    name: stimulus_description
+    doc: the protocol/stimulus name for this patch-clamp dataset
   datasets:
   - name: data
     dtype: float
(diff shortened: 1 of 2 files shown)


oruebel (dismissed):

t-b (dismissed):
Looks good! A quick test also showed that it worked.

bendichter (approved):

Issue #592, opened 2018-08-17
IntervalSeries.data.conversion is stored as 'NaN'


bendichter:
IntervalSeries.conversion returns `conversion = 1.0` before and after read:

```python
from pynwb import NWBFile, NWBHDF5IO
from pynwb.misc import IntervalSeries

interval_series = IntervalSeries(data=[-1, 1], timestamps=[1, 2], name='test_interval_series', source='source')

print(interval_series.conversion)

nwbfile = NWBFile("source", "a file with header data", "NB123A", '2018-06-01T00:00:00')
nwbfile.add_acquisition(interval_series)

with NWBHDF5IO('test_interval_series.nwb', 'w') as io:
    io.write(nwbfile)
    
with NWBHDF5IO('test_interval_series.nwb', 'r') as io:
    nwbfile_in = io.read(nwbfile)
    print(nwbfile_in.acquisition['test_interval_series'].conversion)
```

```
1.0
1.0
```

However the data is actually stored as `'NaN'` (the string `'NaN'`, not a NaN float value):
```python
import h5py
with h5py.File('test_interval_series.nwb', 'r') as file:
    print(file['acquisition']['test_interval_series']['data'].attrs['conversion'])
```
```
NaN
```

This is causing issues for matnwb, which expects this value to follow the schema, which defines it to be a float. Is there a reason why it's stored this way?

bendichter:
There's also a discrepancy between the schema and pynwb.

The pynwb API accepts `float` or `str` for `conversion`:
https://pynwb.readthedocs.io/en/latest/pynwb.base.html#pynwb.base.TimeSeries

However the NWB schema only accepts `float`:
https://github.com/NeurodataWithoutBorders/pynwb/blob/631f6db9cf7fa1eda951094108c435bae7b2027b/src/pynwb/data/nwb.base.yaml#L113-L114

JesseLivezey:
@bendichter @ajtritt I tried to trace the `IntervalSeries` code to find where the `NaN` might get converted to a string, but couldn't figure out where attributes like `conversion` get written. I'm somewhat motivated to fix this since Ben suggested I use this for the dataset I'm converting to NWB. If either of you can give me some suggested places to look, I can try and find the problem and make a PR with a fix/test.

I did notice here that `conversion` might have `NaN` as a default value in this `IntervalSeries` yaml entry.
[...]

PR #605, opened 2018-09-10
[WIP] added float and int type check

JesseLivezey:
## Motivation

NaN was being stored as a string and not a float.

## How to test the behavior?

See issue #592

## Checklist

I haven't written a test since I'm not sure this is the right fix for the problem. I'm also not sure whether I would be using the dtype naming conventions from here
https://github.com/NeurodataWithoutBorders/pynwb/blob/c3a1bed686ff291bb0e523573b9b04ea073dcfc7/src/pynwb/form/backends/hdf5/h5tools.py#L404
I'm currently just interpreting the `spec.dtype` strings as numpy dtypes.

### DIFF
--- a/src/pynwb/form/build/map.py
+++ b/src/pynwb/form/build/map.py
@@ -560,7 +560,7 @@ class ObjectMapper(with_metaclass(ExtenderMeta, object)):
                 else:
                     ret = text_type(value)
             elif ('float' in spec.dtype) or ('int' in spec.dtype):
-                ret = np.asscalar(np.array(value, dtype=spec.dtype))
+                ret = np.array(value, dtype=spec.dtype)
         elif isinstance(spec, DatasetSpec):
             # TODO: make sure we can handle specs with data_type_inc set
             if spec.data_type_inc is not None:
--- a/src/pynwb/form/build/map.py
+++ b/src/pynwb/form/build/map.py
@@ -13,6 +13,8 @@ from ..spec.spec import BaseStorageSpec
 from .builders import DatasetBuilder, GroupBuilder, LinkBuilder, Builder, ReferenceBuilder, RegionBuilder, BaseBuilder
 from .warnings import OrphanContainerWarning, MissingRequiredWarning
 
+import numpy as np
+
 
 class Proxy(object):
     """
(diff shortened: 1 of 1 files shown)


JesseLivezey:
@bendichter using your example, this change gives

```python
from pynwb.misc import IntervalSeries

interval_series = IntervalSeries(data=[-1, 1], timestamps=[1, 2],
                                 name='test_interval_series', source='source')

print(interval_series.conversion)

nwbfile = NWBFile("source", "a file with header data", "NB123A", '2018-06-01T00:00:00')
nwbfile.add_acquisition(interval_series)

with NWBHDF5IO('test_interval_series.nwb', 'w') as io:
    io.write(nwbfile)
    
with NWBHDF5IO('test_interval_series.nwb', 'r') as io:
    nwbfile_in = io.read(nwbfile)
    print(nwbfile_in.acquisition['test_interval_series'].conversion)
```
```
1.0
1.0
```
```python
import h5py
with h5py.File('test_interval_series.nwb', 'r') as file:
    print(file['acquisition']['test_interval_series']['data'].attrs['conversion'])
    print(type(file['acquisition']['test_interval_series']['data'].attrs['conversion']))
```
```
nan
<class 'numpy.float64'>
```
Although I don't understand why it goes `1.0 -> nan -> 1.0`.

bendichter:
Thanks @JesseLivezey this is a step in the right direction, but it's not the solution we ultimately need. The problem is interfacing with matnwb. This solution will at least allow matnwb to read the file, but the resolution and conversion won't be the correct value.

EDIT:

I take that back. `resolution` and `conversion` are meant for data that are measurements e.g. what is the resolution of a voltage recording and what's the conversion to get to volts. For IntervalSeries neither of these are really necessary or meaningful, since the data field is not a measurement, so any reasonable default would work fine. The only remaining issue is the rounding stuff that makes the tests fail

bendichter:
and it looks like this:

TypeError: incorrect type for 'starting_frame' (got 'int64', expected 'Iterable')

Might be due to your `asscalar` call

JesseLivezey:
I was wondering whether these two lines would make more sense as `np.nan`
https://github.com/NeurodataWithoutBorders/pynwb/blob/2118dd456c1bb710444c951b66474d1c9d174057/src/pynwb/misc.py#L164
rather than the default `1.0`

bendichter:
Semantically `NaN` is a little more accurate, but on the other hand I want to avoid a bug if you have code that automatically multiplies `data` by `conversion`. That seems like a very natural thing to do, but it would break `IntervalSeries` if `conversion` is `NaN`, so I prefer `1.0`.

JesseLivezey:
Then does it make sense to make the value stored in the HDF5 file 1.0 rather than NaN? I think that would involve modifying this file
https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/misc.py#L164

Does that file need to be consistent with other yamls in the NWB ecosystem?

PR #606, opened 2018-09-10
ENH(DOC): some typos fixed and more consistency in trailing periods

yarikoptic:
## Motivation

I was looking around for what I could spell check and contribute to ;)

On a serious note - I was just familiarizing myself with the project and NWB in general, so was interested to look into metadata definitions.


bendichter (approved):
Thanks for looking over this with a close eye and fixing these errors!

PR #611, opened 2018-09-12
Bug/unit metadata ids

ajtritt:
## Motivation

remove schema bug. Unit metadata DynamicTable was overriding DynamicTable.id type.


oruebel (dismissed):

bendichter (dismissed):
I checked this with matnwb and it works! Thanks for getting this through so fast!

bendichter:
fix https://github.com/NeurodataWithoutBorders/matnwb/issues/58

PR #620, opened 2018-09-17
Add keywords

bendichter:
Add keywords to spec, alter NWBFile to accept them, and add a test

### DIFF
--- a/src/pynwb/file.py
+++ b/src/pynwb/file.py
@@ -148,6 +148,7 @@ class NWBFile(MultiContainerInterface):
                      'description',
                      'experiment_description',
                      'session_id',
+                     'keywords',
                      'lab',
                      'institution',
                      'notes',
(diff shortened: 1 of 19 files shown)


src/pynwb/file.py
```diff
@@ -180,6 +181,7 @@ class NWBFile(MultiContainerInterface):
             {'name': 'session_id', 'type': str, 'doc': 'lab-specific ID for the session', 'default': None},
             {'name': 'institution', 'type': str,
              'doc': 'institution(s) where experiment is performed', 'default': None},
+            {'name': 'keywords', 'type': Iterable, 'doc': 'Terms to search over', 'default': None},
```
oruebel: Instead of ``'type': Iterable`` I believe this should probably be ``'type' : 'array_data'``

oruebel (dismissed):
Aside from the one comment I think this looks fine

src/pynwb/file.py
bendichter: done

oruebel (approved):

Issue #624, opened 2018-09-20
older yaml. incompatibility with 0.5.1


yarikoptic:
While building updated debian pkg across debian/ubuntus, on debian stable and similarly aged ubuntus got bombarded with multiple instances of similar error
```
======================================================================
ERROR: test_load_namespace_with_reftype_attribute_check_autoclass_const (tests.unit.pynwb_tests.test_extension.TestExtension)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/unit/pynwb_tests/test_extension.py", line 72, in test_load_namespace_with_reftype_attribute_check_autoclass_const
    ns_builder.export(self.ns_path, outdir=self.tempdir)
  File "pynwb/form/utils.py", line 337, in func_call
    return func(self, **parsed['args'])
  File "pynwb/form/spec/write.py", line 172, in export
    writer.write_spec(out, path)
  File "pynwb/form/spec/write.py", line 41, in write_spec
    self.reorder_yaml(os.path.join(self.__outdir, path))
  File "pynwb/form/spec/write.py", line 50, in reorder_yaml
    sorted_data = self.sort_keys(data)
  File "pynwb/form/spec/write.py", line 59, in sort_keys
    yaml.RoundTripRepresenter.add_representer(type(None), my_represent_none)
AttributeError: 'module' object has no attribute 'RoundTripRepresenter'

======================================================================
ERROR: test_load_namespace_with_reftype_attribute_check_autoclass_const (tests.unit.pynwb_tests.test_extension.TestExtension)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/unit/pynwb_tests/test_extension.py", line 31, in tearDown
    os.remove(path)
OSError: [Errno 2] No such file or directory: '/tmp/Z8Q4XV664Y_namespace.yaml'
```
due to recently introduced
```
+
+        # Represent None as null
+        def my_represent_none(self, data):
+            return self.represent_scalar(u'tag:yaml.org,2002:null', u'null')
+        yaml.RoundTripRepresenter.add_representer(type(None), my_represent_none)
```

on those systems python-ruamel.yaml (0.13.4-2), when recent ones have 0.15.34 

I will quickly look into backporting that recent release for neurodebian but just wanted to bring to your attention since you might need to boost you version depends or desire to provide a workaround?

Related -- either I lost my git-mind, or bitbucket is having issues since
```shell
smaug:/tmp
[...]

PR #625, opened 2018-09-20
BF: use yaml. submodules to stay compatible with older ruamel-yaml

yarikoptic:
Fixes https://github.com/NeurodataWithoutBorders/pynwb/issues/624


ajtritt (approved):

yarikoptic:
FWIW - reconfirming that fixed test failures while building 0.5.1 on debian 9.0 /ubuntu 17.10

PR #627, opened 2018-09-20
Shape checker

### LINKED ISSUE #619: check shape in docval
pynwb does not enforce that the shape of datasets matches the shape specification in the schema. matnwb expects datasets to follow the size rules specified by the schema, so this is causing a compatibility issue for files that are written in python and read in matlab. For example, it is possible to write a 1-D dataset to the variable `ecephys.`[WaveformClusters](https://github.com/NeurodataWithoutBorders/pynwb/blob/91676cc29bf22c284db63379a676136d64827cab/src/pynwb/ecephys.py#L310)`.waveform_mean`, even though the [schema requires that the dataset be 2-D](https://github.com/NeurodataWithoutBorders/pynwb/blob/91676cc29bf22c284db63379a676136d64827cab/src/pynwb/data/nwb.ecephys.yaml#L159). This dataset will then be written with no problem in pywnb, but will cause a read error in matnwb. 

I suggest we add an optional dataset shape checker to docval. The syntax would be:

```
{'name': 'waveform_mean', 'type': Iterable, 'doc': 'the mean waveform for each cluster', 'shape': (None, None)}
```
The docval would take either a list of lists or an `np.ndarray`, and would enforce that the dataset is 2-D.

@ajtritt, what do you think about this?

### LINKED ISSUE #653: outdated dimension matching tests
## 1) Bug

test_ephys is attempting to check for dimension matching here. 

https://github.com/NeurodataWithoutBorders/pynwb/blob/4b0077c64e015ddfa05a3cb696dba32b92d05dd0/tests/unit/pynwb_tests/test_ephys.py#L198-L228

The tests pass, but for the wrong reason. The tests check that e.g. if `description` and `electrodes` don’t have matching length a `TypeError` is thrown. I was trying to change this to a `ValueError`, and I found that a `TypeError` is in fact thrown, but is due to electrodes being a list, when it should be a region reference. Let’s fix up these tests so they cause errors for the right reason.

bendichter:
## Motivation

Enforce shapes in docval. This is important in order to enforce the spec more faithfully which will help us play nice with matnwb. docval has a new attribute, `'shape'`, which follows the syntax `(None, 2)` means `x.shape` must be `([any], 2)`. I added this to a few of the data types in ecephys and changed the tests accordingly. Allows for multiple shape options e.g. `((None, 2), (None, 3))`

fix #619
fix #653 
## Checklist


bendichter:
@ajtritt There is one iffy part of this that I would like some advice on. I noticed in the `__parse_args` of `docval` that errors were concatenated. All of the errors were tallied up and reported at the end. I like this because it shows the user multiple `TypeError`s in one run, but I wasn't quite sure how to best respect the error management. I wanted to tack my shape errors onto this list, but then they were reported as `TypeError`s, which is not quite right. I decided to just throw a `ValueError` directly in the standard way without concatenating them together, but now I don't like that because any `ValueError` will override your error list. Maybe the best approach is to create 2 threads, a `ValueError` thread for shape issues, and a `TypeError` thread? Maybe we should have a single thread and just stick with a single error type, even if it's less specific?

oruebel:
Having the errors as one list at the end is convenient. The actual type of the error is mainly important for readability for the users. One solution may be to:

1. Define a custom ValdiationError
2. In the text message indicate for every item what type of error each item indicates

bendichter:
thanks @oruebel I like that solution

oruebel:
I think the following might work too (although I'm not sure): 1. as errors occur just add them to a list in docval and  2. then raise a list of errors at the end:

```
raise VaidationError(
    [ValueError('test'), 
     TypeError('test2), 
     ...]
)
```

I'm not sure if the output of this would look good, but it would allow you to use different types of errors. Might be worth a try.

bendichter:
It looks like the classes I am thinking about building already exist:

https://github.com/NeurodataWithoutBorders/pynwb/blob/de391bf1221c03558e883932c20999f84e35720f/src/pynwb/form/validate/errors.py#L97

but are used exclusively in `validator` to validate a saved NWB file, not in `docval` `__args_parse` to validate the docval arguments when writing a file. Is there a reason for this? May I use them for docval as well?

bendichter:
Never mind, that was a bit of a rabbit-hole. I ended up creating a generic error type, `ArgValidationError` where both types of docval errors go (type and shape for now).

src/pynwb/form/utils.py
```diff
@@ -345,7 +377,7 @@ def func_call(*args, **kwargs):
                 parse_err = parsed.get('errors')
                 if parse_err:
                     msg = ', '.join(parse_err)
-                    raise_from(TypeError(msg), None)
+                    raise_from(ArgValidationError(msg), None)
```
ajtritt: Errors with Shape and Type should not be conflated. Any argument validation errors related to shape should be subclasses of [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError)--this is used to indicated argument errors when something isn't the correct type.

ajtritt (dismissed):
Leave argument errors related to *type* as TypeErrors, and make errors related to shape (when type is okay) ValueErrors or a subclass of ValueError

bendichter:
See the above comment about losing the error list. How do you recommend I
resolve that while throwing two different types of errors?

ajtritt:
TypeErrors should take precedence over ValueErrors/ShapeErrors. ValueErrors are secondary to TypeErrors, and should be raised after TypeErrors no longer remain.

bendichter:
Ok I can do that

On Sep 24, 2018 7:54 AM, "Andrew Tritt" <notifications@github.com> wrote:

TypeErrors should take precedence over ValueErrors/ShapeErrors. ValueErrors
are secondary to TypeErrors, and should be raised after TypeErrors no
longer remain.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/NeurodataWithoutBorders/pynwb/pull/627#issuecomment-424066056>,
or mute the thread

bendichter:
@ajtritt I might not get to the changes you requested for a few days. Feel free to implement it yourself in the meantime, particularly if you have strong feelings about how you would like errors to be combined and reported to the user.

bendichter:
As it currently stands on the dev branch, we are throwing [missing arg errors](https://github.com/NeurodataWithoutBorders/pynwb/blob/d4de20f56d10f85b2921fb40909143fb64d9b5f7/tests/unit/form_tests/utils_test/test_core.py#L93-L100) and [extra arg errors](https://github.com/NeurodataWithoutBorders/pynwb/blob/d4de20f56d10f85b2921fb40909143fb64d9b5f7/tests/unit/form_tests/utils_test/test_core.py#L283-L289) as `TypeErrors`, which I don't think is accurate. I like the way I propose here, having missing arg, extra arg, type, and shape errors are all thrown as the `ArgValidationError`. It's vague but it's true, and it allows us to show users all of these errors at once. I suppose the alternative would be to choose an appropriate error type for each of these different types of errors and decide on an order to throw these errors but I think that might be missing the forest for the trees.

oruebel:
Just looking at the discussion, I think the question is in part: 1) what is the order in which different aspects are validated and 2) whether we should stop the validation for an argument once an error has been found. Just to make things concrete. Suppose we have an argument ``data`` that is expected to be an array of ``dtype=float`` with a ``shape=(None,3)`` and the user sets it to be a string, e.g., ``MyContainer(data='my test data')``. In this case you would technically have 3 validation errors: 1) a TypeError because a str is not an array, 2) a TypeError because you didn't get a float, 3) a ``ShapeError``.  For readability and to avoid redundancy, you probably just want to report one error for each argument, where the type-related errors take precedence over the shape-related errors. I think it would be fine to order the validation steps in doc-val accordingly and stop validation for an argument once an error has been found. Alternatively one could run all validations for one arguments and then collect all the errors for one argument into a single error: 

```
ArgValidationError(``data`` argument failed to validate due to: 
       - TypeError : expected ``data`` to be ``float`` got  ``str`` instead
       - ShapeError : expected ``data`` to be array of shape ``(null, 3)`` received ``(1,)``
       - ...)
```
 But I honestly think that ordering the validation steps in docval and stopping on first error is fine.

ajtritt:
`TypeError` is what Python raises for errors related to function arguments i.e. missing, extra, and unexpected keyword arguments. docval should always raise a TypeError for these types of errors.

Given the motivation (i.e. #619) for this change, I think we should revisit the validation-on-write feature. I started that on the [enh/validate_on_write](https://github.com/NeurodataWithoutBorders/pynwb/tree/enh/validate_on_write), and set it aside for some reason. It doesn't have the feature of shape checking at function call, but it would be a step toward solving the problem of writing incorrectly shaped datasets. We already have a [ShapeError](https://pynwb.readthedocs.io/en/latest/pynwb.form.validate.errors.html#pynwb.form.validate.errors.ShapeError), so shape issues should be caught.

bendichter:
> I honestly think that ordering the validation steps in docval and stopping on first error is fine.

well that would make this easy. @ajtritt, do you agree? I know you implemented machinery to collect errors and might not want to see that feature removed.

bendichter:
> `TypeError` is what Python raises for errors related to function arguments i.e. missing, extra, and unexpected keyword arguments.

Huh, I just checked this and it looks like you are right.

This is completely working minus details about error passing. It is better to throw errors as early as we can, so I think we should stay on this route. Validation on write could still be useful, but I don't think it solves this problem.

ajtritt:
How do all the constructors of objects that map to neurodata_types in the schema know what datasets they're supposed to validate, and what the valid shapes for those datasets are?

bendichter:
So I take it I should change this as well?

https://github.com/NeurodataWithoutBorders/pynwb/blob/d4de20f56d10f85b2921fb40909143fb64d9b5f7/src/pynwb/ecephys.py#L421-L429

bendichter:
@oruebel I ended up making a list of `TypeError`s (missing args, extra args, etc. This may not sound like a "type error" but it is as python defines them) and a list of `ValueError`s (right now just shape errors, but could include other errors in the future). `TypeError`s take precedence, just as they do in normal python.

bendichter:
@ajtritt Could you help me move this along?

bendichter:
@oruebel what do you think?

bendichter:
@ajtritt can you take a look at this for me?

tests/unit/form_tests/test_io_hdf5_h5tools.py
```diff
@@ -423,7 +423,7 @@ def test_link_resolve(self):
         for passband in ('theta', 'gamma'):
             electrical_series = ElectricalSeries(name=passband + '_phase',
                                                  source='ephys_analysis',
-                                                 data=[1., 2., 3.],
+                                                 data=[[1.], [2.], [3.]],
```
ajtritt: why does this need to be 2D?

tests/unit/form_tests/utils_test/test_core.py
```diff
@@ -63,6 +63,36 @@ def method(self, **kwargs):
...
             method(self, arg1=1234560)
 
+    def test_bad_shape(self):
+        @docval({'name': 'arg1', 'type': 'array_data', 'doc': 'this is a bad shape', 'shape': (None, 2)})
+        def method1(self, **kwargs):
+            pass
+        with self.assertRaises(ValueError):
+            method1(self, arg1=[[1]])
+
+        @docval({'name': 'arg1', 'type': 'array_data', 'doc': 'this is a bad shape', 'shape': (None, 2)})
```
ajtritt: this might be more appropriate as 3 different tests, since it's testing 3 different cases

bendichter:
@ajtritt thanks, I fixed the issues you brought up

ajtritt (approved):
Thanks @bendichter

Issue #626, opened 2018-09-20
dim scales: Enhance specification language and containers


oruebel:
## 1) Feature Request

**Problem:** Currently the schema does not support direct association of datasets with dimensions of another datasets

**Proposed Solution**

**Part 1: Enhance the specification language:** Enhance the ``dims`` key in the specification language to not just allow a string to label a dimension but to provide a dict of the form 

```
dims: 
   - - label: time
        scale: timestamps
      - label: electrodes
        scale: electrodes   
   - 
   ....
```

Here the ``scale`` key would refer to a dataset with a fixed name located within the same ``Group``. In this way, we do not need to add complex cross-referencing mechanisms to the schema and requiring that scales should be in the same location as the data they label seems appropriate.  The ``scale`` key should be allowed to be a list to enable multiple scales for a single axis. Also, by using a reference to a table as a dimension scale, one can also encode complex annotations and labels in a scale. 

**Part 2: Enhance the container classes** In the API, a user should be able to request a list of all the scales for a particular dataset. For primary data containers (e.g., TimeSeries) this could be a function on the container class (e.g., ``get_scales`` or ``dims``) that would return an OrderedDict, of dicts {'time': {'timestamps': timestamps, ... }, ...}. 

### Problem/Use Case

For visualization, analysis, and query it is useful to be able to programmatically inspect the dimensions of an array in a consistent fashion. In this way, we can, e.g., automatically label axes in plots or query the dimensions of an array without having to know the specifics of containers

## Checklist

- [X] Have you ensured the feature or change was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues/) ?
- [X] Have you included a brief and descriptive title?
- [X] Have you included a clear description of the problem you are trying to solve?
- [ ] Have you included a minimal code snippet that reproduces the issue you are encountering?

bendichter:
If we are not going to yoke label with scale (which I'm fine with) I'd actually rather this go in 'shape'. I propose that you specify this by

```
shape:
    - 2
    - electrodes
```
Note that the current syntax of 
```
shape:
    - 2
    - null
```
is still valid. "null" (or blank) will be a special string that means "no scale matching." For all other non-int strings, dimensions of different datasets with matching string shape parameters must match.

How broad of a scope do we want this to be enforced? I think it should be enforced within a Group but not for the entire file. That will allow us to build this into docval, and I am worried that a file-wide scope will cause problems.

oruebel:
[...]

Issue #594, opened 2018-08-20
add support for generic numeric dtype 


ajtritt:
### Problem/Use Case
There is no way to specify numeric data types as either floating point or integer types i.e. users must specify one or the other. In 1.x, this was accomplished with `dtype: number`, but this has since been dropped.

HDF5IO needs to be updated to except `number`, and more generally, how to handle generic dtypes.  This should start somewhere around here:

https://github.com/NeurodataWithoutBorders/pynwb/blob/f91e0693259f9fef5e13697f15d3b2507385acfb/src/pynwb/form/backends/hdf5/h5tools.py#L404

This is needed for https://github.com/NeurodataWithoutBorders/nwb-schema/issues/178

## Checklist

- [ ] Have you ensured the feature or change was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues/) ?
- [ ] Have you included a brief and descriptive title?
- [ ] Have you included a clear description of the problem you are trying to solve?
- [ ] Have you included a minimal code snippet that reproduces the issue you are encountering?

PR #628, opened 2018-09-20
Update CONTRIBUTING.rst

bendichter:
Proposing some changes to Contributing to make it a little friendlier and and to reflect some changes we have made in our workflow. I'm open to discussion on these changes.

## Motivation

## How to test the behavior?
```
```

## Checklist


docs/CONTRIBUTING.rst
```diff
@@ -16,11 +16,6 @@ Types of Contributions
 Did you find a bug? or Do you intend to add a new feature or change an existing one?
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-* **Identify the appropriate repository** for the change you are suggesting:
```
oruebel, line 19: I think the "**Identify the appropriate repository**" section should be kept. The part we should clarify here, however, is that changes to the schema should be implemented and tested in PyNWB first. The changes can then be migrated to the nwb-schema repo once formally accepted. However, issue tickets regarding the schema should still be filed on the nwb-schema repo.

oruebel, line 19: I think the "Identify the appropriate repository" section should be kept. The part we should clarify here, however, is that changes to the schema should be implemented and tested in PyNWB first. The changes can then be migrated to the nwb-schema repo once formally accepted. However, issue tickets regarding the schema should still be filed on the nwb-schema repo.

oruebel (commented):
Other than that we should keep a section on guidance for what repo to use for what, the changes look good.

docs/CONTRIBUTING.rst
bendichter: OK, I do like having them separate so that the matnwb folks have a feed of the stuff that pertains to them as well. If we want to continue to have this be a rule then we need to follow it ourselves.  #552 was about ROI spec, and #626 and #594 were about the schema language... and there are about a dozen that I created that should probably be in schema as well.

oruebel: I agree.

ajtritt: @oruebel @bendichter Is this conversation concluded? Can we merge this?

ajtritt (approved):

Issue #614, opened 2018-09-12
Avoid double write of spec


oruebel:
## 1) Bug

``YAMLSpecWriter.write_spec`` currently writes the spec files and then reads/writes them again using ``reorder_yaml``.

```
   def write_spec(self, spec_file_dict, path):
        with open(os.path.join(self.__outdir, path), 'w') as stream:
            self.__dump_spec(spec_file_dict, stream)
        self.reorder_yaml(os.path.join(self.__outdir, path))
```

Instead ``write_spec`` should sort and write the spec directly. 

## Checklist

- [X] Have you ensured the feature or change was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues/) ?
- [X] Have you included a brief and descriptive title?
- [X] Have you included a clear description of the problem you are trying to solve?
- [X] Have you included a minimal code snippet that reproduces the issue you are encountering?

PR #630, opened 2018-09-26
Avoid double write of spec

jayrbolton:
## Motivation

This is a shot at issue https://github.com/NeurodataWithoutBorders/pynwb/issues/614. I reorganize the YAMLSpecWriter#write_spec method to avoid writing out the spec file twice. Instead, it sorts and then writes the file once. Let me know if you'd like me to tweak anything further. I'm still getting a feel for the codebase.

(cc @oruebel)

P.S. I assume the use of `json.loads(json.dumps(x))` in this module is just a generic way to convert objects to plain dict/lists. I imagine this is not a very efficient approach, maybe there is another way to go about it. 
## How to test the behavior?

I didn't add new tests; looks like 7 existing tests hit this method. Tests pass with these changes.

## Checklist

### DIFF
--- a/src/pynwb/form/spec/write.py
+++ b/src/pynwb/form/spec/write.py
@@ -33,23 +33,19 @@ class YAMLSpecWriter(SpecWriter):
         self.__outdir = getargs('outdir', kwargs)
 
     def __dump_spec(self, specs, stream):
-        yaml.main.safe_dump(json.loads(json.dumps(specs)), stream, default_flow_style=False)
+        specs_plain_dict = json.loads(json.dumps(specs))
+        yaml.main.safe_dump(specs_plain_dict, stream, default_flow_style=False)
 
     def write_spec(self, spec_file_dict, path):
-        with open(os.path.join(self.__outdir, path), 'w') as stream:
-            self.__dump_spec(spec_file_dict, stream)
-        self.reorder_yaml(os.path.join(self.__outdir, path))
+        out_fullpath = os.path.join(self.__outdir, path)
+        spec_plain_dict = json.loads(json.dumps(spec_file_dict))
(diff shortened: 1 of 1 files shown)


src/pynwb/form/spec/write.py
```diff
@@ -33,24 +33,20 @@ def __init__(self, **kwargs):
...
+        spec_plain_dict = json.loads(json.dumps(spec_file_dict))
+        sorted_data = self.sort_keys(spec_plain_dict)
+        with open(out_fullpath, 'w') as fd_write:
+            yaml.dump(sorted_data, fd_write, Dumper=yaml.dumper.RoundTripDumper)
 
     def write_namespace(self, namespace, path):
         with open(os.path.join(self.__outdir, path), 'w') as stream:
             self.__dump_spec({'namespaces': [namespace]}, stream)
 
-    def reorder_yaml(self, path):
```
oruebel: I think we may want to keep the reorder_yaml function, even though it is not needed for the main write. I think it is helpful to have to have it. Maybe it can just change to 

```
def reorder_yaml(self, path):
    with open(path, 'rb') as f_in:
         data = yaml.load(f_in, Loader=yaml.loader.RoundTripLoader)
    write_spec(data, path)
```

oruebel (commented):
Other than possibly keeping the reorder_yaml function, this looks good to me. @ajtritt what do you think should we keep the reorder_yaml function or remove it?

src/pynwb/form/spec/write.py
jayrbolton: Sure thing

jayrbolton: I added `reorder_yaml` back in and tested it manually. Let me know if you think it needs a unit test.

oruebel (approved):

PR #637, opened 2018-10-03
pandas interfaces for tabular data

### LINKED ISSUE #631: feature: pandas interface for tables
pandas is a widely-adopted and performant Python package for working with (and viewing!) tabular data. It would be useful if DynamicTable and NWBTable had methods for casting to and from pandas DataFrames. 

I'm working on this [here](https://github.com/NileGraddis/pynwb/tree/feature/units-interface). If there are no objections, I'll keep going.

NileGraddis:
Resolves #631

## Motivation
Pandas is widely used and performant. Providing tools for converting NWB tabular data in and out of pandas will make pynwb easier to integrate with other analysis software and easier to introduce to new users. This functionality is optional and requires no schema changes.

These changes were proximately motivated by my, @sgratiy, @nicain, and @JFPerkins' recent work on using pynwb with Allen Institute data. We've found that when creating an nwb file it is often easiest to assign a bundle of tabular data in one shot from pandas and that when reading one it is often helpful to use pandas to display the data.

## How to test the behavior?
Please see the added integration tests in test_core and test_nwbfile and the added unit tests in test_core and test_epoch.


ajtritt:
Hey @NileGraddis, I'm currently working on making the existing tables into DynamicTables. The goal is to get rid of most of the compound type based tables. With that said, Epochs and ROITable will become DynamicTables. When I get those changes merged in, we should no longer need the multiple implementations of to_dataframe and from_dataframe. 

I also noticed that you add a to_dataframe and from_dataframe method to VectorData. Was there a reason for this?

oruebel:
@NileGraddis thanks for working on this issue. It looks like @ajtritt beat me too it, but I had the same questions, i.e, 1) we should avoid replicating the "same" functionality in multiple places (it sounds like this won't be an issue after the schema changes are done), and 2) it is not clear to me what the advantage of representing VectorData (i.e., a collection of variable-length vectors) as a dataframe is.

NileGraddis:
@ajtritt @oruebel 

1. I completely agree that we should avoid reimplementing functionality (and thank you for expanding the use of DynamicTable!). Why don't we delete the unnecessary implementations when your changes are complete? That way, we and other uses can benefit from these methods in the mean time.

2. I did not implement to_dataframe and from_dataframe for VectorData. In the github diff UI you can click the expand button to see code that was elided for space (you might need to do this multiple times). You can also see the changes [on my feature branch](https://github.com/NileGraddis/pynwb/blob/feature/units-interface/src/pynwb/core.py#L284).

3. I see that I've got some flake8 warnings. I'll fix these.

oruebel:
R2 1. Since Andrew is working on the refactor right now, my guess is that it might be simpler to add your functions on top of his changes afterwards, rather than the other way around. I'd chat with @ajtritt to see what will be easier. In either case I don't think this should delay things by more than a week or two. 

Re 2. Sorry for misreading this in the GitHub UI, my mistake.  

Re 3. It looks like also docs/gallery/general/extensions.py broke according to the the gallery36 tests.

NileGraddis:
re 1. At the Allen we're currently pushing to get a bunch of our data into NWB files using pynwb. One of the reasons we're doing this is so that we can support Kitware's visualization tool development by providing them NWB files containing real data and examples of use. This work will be aided by having pandas interoperability in place. I'd be willing to do any work required to get these methods working again when the refactor is complete.


re 3. This is failing because I expanded the validator on DynamicTable.add_row to catch extra arguments in addition to missing ones (this helps to prevent the situation where you release a bunch of data, then realize after the fact that not everything you thought you were writing to the files actually got written). In this case it is failing because ['description' was dropped](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/file.py#L497) in the new electrode table implementation. I'm leaning towards fixing the example (since having a description on an electrode seems silly), but could also just add description back to the electrodetable. What are your thoughts @ajtritt ?

src/pynwb/core.py
```diff
@@ -744,6 +803,11 @@ def add_row(self, **kwargs):
         val = getargs('val', kwargs)
         self.data.append(val)
 
+    def extract_slice(self, slc, **constructor_kwargs):
```
ajtritt: Is the intention of this to make TableColumn behave like a pandas DataFrame and return a new TableColumn with the subset?

src/pynwb/epoch.py
```diff
@@ -165,3 +168,82 @@ def __calculate_idx_count(self, start_time, stop_time, ts_data):
         count = stop_idx - start_idx
         idx_start = start_idx
         return (int(idx_start), int(count))
+
+
+    def to_dataframe(self):
```
ajtritt, line 172: I am in the process of making Epochs a DynamicTable, so this functionality should be inherited from the super class.

src/pynwb/file.py
```diff
@@ -346,6 +346,21 @@ def file_create_date(self):
     def session_start_time(self):
         return self.__session_start_time
 
+
+    def __check_epochs(self):
+        if self.epochs is None:
+            self.epochs = Epochs(self.source)
+
+    @docval(*get_docval(DynamicTable.add_column))
+    def add_epoch_metadata_column(self, **kwargs):
```
ajtritt, line 354: Thanks for adding this. When Epochs is a DynamicTable, this should work perfectly.

src/pynwb/core.py
```diff
@@ -906,6 +982,81 @@ def __getitem__(self, key):
 
         return ret
 
+    @docval(
+        {'name': 'region', 'type': (slice, list, tuple), 'doc': 'the indices of the table'},
+        {'name': 'name', 'type': str, 'doc': 'the name of the extracted subtable', 'default': None},
+        {'name': 'source', 'type': str, 'doc': 'a description of where the extracted subtable came from', 'default': None},
+        {'name': 'description', 'type': str, 'doc': 'a description of what is in the extracted subtable', 'default': None}
+        )
+    def extract_subtable(self, **kwargs):
```
ajtritt: Same question as I had about TableColumn.extract_slice. Are you trying to add pandas-DataFrame-like behavior?

ajtritt (changes requested):
@NileGraddis After more careful review, most of this looks fine. The only issues are with the subsetting methods and the reimplementation of to_dataframe/from_dataframe on Epochs. The latter should be taken care of when I finish the Epoch refactor.

Regarding the former. I'm concerned about returning TableColumn/DynamicTable subsets. DynamicTable and TableColumn are containers that represent objects from the NWB data model. Subsets of the data contained in these objects don't fully represent the object that their copied metadata (i.e. description, source, parent, etc) would suggest. I worry this might lead to unintended behavior and side effects. I prefer that we don't merge this part, and address this functionality with a more focused effort. 

One last little request... Could you add some docstrings to the methods that stay. `docval` will autogenerate argument stuff, but it would be nice to have a simple description.

src/pynwb/core.py
NileGraddis: The behavior of a slice operation returning an object (or view) of the same class is a bit more generic than pandas and numpy - it means that if I write some client code for analyzing/viewing/whatever the data in a DynamicTable my code also works on a slice from that table.

I don't need this functionality right now though and am happy to shelf it for later. Particularly in light of your above concerns.

src/pynwb/epoch.py
NileGraddis: Sweet! Do you mind if we leave this around until then? The change should just be deletion and having these methods right now will be useful for our current work at the Allen.

NileGraddis:
@ajtritt 

Thank you for the feedback. As of now I'm thinking that my todos are:
1. remove the table subslicing methods.
2. add docstrings

re: to/from_dataframe on epochs. As the API we want has these methods anyways, the work to remove them is pretty straightforward, and the Epochs -> DynamicTable work might take (quoting @oruebel 1-2 weeks) I'd prefer if we leave these in place for now and just delete them when Epochs is converted to a DynamicTable. This will make it easier for us to generate files during this time and make our client code more stable. If you are virulently opposed to this plan, let me know and I will remove them.

re: the example. You've linked to domain/ecephys, but this issue is also present in general/extensions (which is where I ran into it). I certainly don't mind rebasing, though.

ajtritt:
> re: the example. You've linked to domain/ecephys, but this issue is also present in general/extensions (which is where I ran into it). I certainly don't mind rebasing, though.

My bad. I'm still confused as to why this caused a failure then. NWBFile.add_electrodes is allowed to take in extra arguments. In this case, it should just ignore `description`

>I'd prefer if we leave these in place for now and just delete them when Epochs is converted to a DynamicTable.

That's fine. As a heads up, anything you write with this structure will not be be readable. /epochs will become a DynamicTable. As of now, the DynamicTable in epochs are working on is /epochs/metadata. So, after the Epochs refactor, additional metadata columns will be read from /epochs/<column_name> and anything you write before the refactor will put them in /epochs/metadata/<column_name>.

NileGraddis:
@ajtritt  I [added that to the validation on add_row](https://github.com/NileGraddis/pynwb/blob/feature/units-interface/src/pynwb/core.py#L901). I'm concerned about the case where we write/release a bunch of files then realize after the fact that actually not all of the data we thought we were writing made it in. So far we've run into this twice that I know of - @nicain hit this issue when writing a file and us just now with this example, which was silently failing to write description. This could be a warning instead of an error, I suppose.

ajtritt:
@NileGraddis 
> This could be a warning instead of an error, I suppose.

An error is good. Thanks for adding that.

NileGraddis:
@ajtritt 

re: files with Epochs backwards compatibility. Gotcha - We'll make sure to regenerate our files when that change is implemented.

I think that covers all the changes. Are we good to go?

ajtritt (approved):

PR #640, opened 2018-10-03
WIP: mk/datetime

ukos-git:
## Issue
https://github.com/NeurodataWithoutBorders/nwb-schema/issues/50#issuecomment-420761991

## Motivation

Previously, all date formats were allowed in NWB for storing datetime objects. This results in incompatibility, making the program more complicated and making **magic** functions like `dateutil.parse` necessary.

ISO 8601 date strings are a defined way for storing datetime object information as readable strings as UTC with timezone offset.

They are independent from dtype specifications of underlying modules like hdf5 or numpy which only accepts timezone naive utc formats. DateTime naive is not wanted in neuro sciences as specific information about the local time (is it morning or evening) is missing.

## How to test the behavior?
input `datetime.now()` (tz naive) to NWBFile Container class and write to hdf5. The time datasets get written as iso string. Loading a hdf5 file will result in correct datetime object in the Container class.

Simple testing can be performed by executing `docs/gallery/domain/ecephys.py`.

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+arrow==0.10.0
 certifi==2018.1.18
 chardet==3.0.4
 h5py==2.7.1
--- a/src/pynwb/data/nwb.file.yaml
+++ b/src/pynwb/data/nwb.file.yaml
@@ -15,12 +15,14 @@ groups:
     value: 2.0b
   datasets:
   - name: file_create_date
-    dtype: text
-    doc: 'Time file was created, UTC, and subsequent modifications to file. COMMENT:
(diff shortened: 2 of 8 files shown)


ukos-git:
This is still in WIP state but read/write to/from hdf5 is working. There is also a WIP:commit which adds some comments and will be dropped before merge.

[ ] unit tests
[ ] stress test
[ ] maybe introduce a generic location for dtype conversions after reading from hdf5 (directly in the ObjectMapper or define a subclass for datetime representations).

@t-b Can you take a look if this is going in the right direction?

ukos-git:
updated location: https://github.com/NeurodataWithoutBorders/pynwb/pull/641

PR #641, opened 2018-10-03
isoformat for datetime objects in hdf5

ukos-git:
## Issue
https://github.com/NeurodataWithoutBorders/nwb-schema/issues/50#issuecomment-420761991

## Motivation

Previously, all date formats were allowed in NWB for storing datetime objects. This results in incompatibility, making the program more complicated and making **magic** functions like `dateutil.parse` necessary.

ISO 8601 date strings are a defined way for storing datetime object information as readable strings as UTC with timezone offset.

They are independent from dtype specifications of underlying modules like hdf5 or numpy which only accepts timezone naive utc formats. DateTime naive is not wanted in neuro sciences as specific information about the local time (is it morning or evening) is missing.

## How to test the behavior?
input `datetime.now()` (tz naive) to NWBFile Container class and write to hdf5. The time datasets get written as iso string. Loading a hdf5 file will result in correct datetime object in the Container class.

Simple testing can be performed by executing `docs/gallery/domain/ecephys.py`.

## Checklist


ukos-git:
This is still in WIP state but read/write to/from hdf5 is working. There is also a WIP:commit which adds some comments and will be dropped before merge.

[ ] unit tests
[ ] stress test
[ ] maybe introduce a generic location for dtype conversions after reading from hdf5 (directly in the ObjectMapper or define a subclass for datetime representations).

@t-b Can you take a look if this is going in the right direction?

src/pynwb/data/nwb.file.yaml
```diff
@@ -15,12 +15,14 @@ groups:
...
-    doc: 'Time file was created, UTC, and subsequent modifications to file. COMMENT:
-      Date + time, Use ISO format (eg, ISO 8601) or a format that is easy to read
-      and unambiguous. File can be created after the experiment was run, so this may
-      differ from experiment start time. Each modification to file adds new entry
-      to array.'
+    dtype: isodatetime
+    doc: 'A record of the date the file was created and of subsequent modifications.
+      COMMENT:
+        - The date is stored in UTC with local timezone offset as ISO 8601
+          formatted string: 2018-09-28T14:43:54+02:00
```
t-b: @ukos-git Add a sentence that we can have optional milliseconds as well. And instead of a numeric timezone offset we can also have `Z` for UTC.

src/pynwb/data/nwb.file.yaml
```diff
@@ -34,10 +36,12 @@ groups:
...
   - name: session_start_time
-    dtype: text
-    doc: 'Time of experiment/session start, UTC.  COMMENT: Date + time, Use ISO format
-      (eg, ISO 8601) or an easy-to-read and unambiguous format. All times stored in
-      the file use this time as reference (ie, time zero).'
+    dtype: isodatetime
+    doc: 'Date and time of the experiment/session start.
+      COMMENT:
+        - The date is stored in UTC with local timezone offset as ISO 8601
+          formatted string: 2018-09-28T14:43:54+02:00
```
t-b: Same as comment above.

src/pynwb/file.py
```diff
@@ -247,16 +247,19 @@ def __init__(self, **kwargs):
         super(NWBFile, self).__init__(*pargs, **pkwargs)
         self.__session_description = getargs('session_description', kwargs)
         self.__identifier = getargs('identifier', kwargs)
+
         self.__session_start_time = getargs('session_start_time', kwargs)
-        if isinstance(self.__session_start_time, str):
-            self.__session_start_time = parse_date(self.__session_start_time)
+        if self.__session_start_time.tzinfo is None:
```
t-b, line 252: @ukos-git Do we wanna add a warning here so that the user is made aware of the missing timezone?

t-b:
@ukos-git Looks good from the general direction.

I've added some minor comments.
The testing commit I added gives for the icephys.py example the following output

```
$ python icephys.py
C:\Anaconda2\envs\py36\lib\site-packages\h5py\__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to
 `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
before write
None
2018-10-04T19:08:05.285137
tzutc()
2018-10-04T17:08:05.288137+00:00
after read
None
2018-10-04T19:08:05.285137
  File "icephys.py", line 132, in <module>
    print(nwbfile.file_create_date[0].tzinfo)
AttributeError: 'str' object has no attribute 'tzinfo'
(py36)
```

From reading the code I would have expected to see timezones for both before writing and both after reading, and of course no error.

t-b (changes requested):

t-b:
@ukos-git The NWB file is still missing the timezone info.

```
   DATASET "session_start_time" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_UTF8;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SCALAR
      DATA {
      (0): "2018-10-04 19:08:05.285137"
      }
   }
```

file_create_date has it. In addition the separator is ` ` instead of `T` for both.

See

```
  DATASET "file_create_date" {
     DATATYPE  H5T_STRING {
        STRSIZE H5T_VARIABLE;
        STRPAD H5T_STR_NULLTERM;
        CSET H5T_CSET_UTF8;
        CTYPE H5T_C_S1;
     }
     DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
     DATA {
     (0): "2018-10-04 17:08:05.288137+00:00"
     }
  }
```

(Output created with h5dump).

ukos-git:
>   File "icephys.py", line 132, in <module>
>     print(nwbfile.file_create_date[0].tzinfo)
> AttributeError: 'str' object has no attribute 'tzinfo'

ok there is something fundamentally wrong here. This should return datetime objects. Thank you for finding this.

> From reading the code I would have expected to see timezones for both before writing and both after reading, and of course no error.

From what I learnt about the pynwb structure:

The builder object gets "constructed" on "read" and "written" on write. So there is indeed a split into the two interfaces.

The container is then built from the builder by writing the read objects as docvals to the container class. For converting the str datetime objects from the builder to the docvals there are the two override functions in the NWBFileMap.

Therefore, In the end the container class is the same interface for read and write.

The timezone is then added within the container class. It is the only class that gets real datetime objects.

I think a graph which shows the dependencies for write/read interactions with the main classes and subclasses as well as thier interfaces would be highly needed.

t-b:
@ukos-git 

> I think a graph which shows the dependencies for write/read interactions with the main classes and subclasses as well as thier interfaces would be highly needed.

If you can do that in a finite and small amount of time. Go for it!

src/pynwb/file.py
ukos-git: Is there any log level we can write this to without writing to stdout?

ukos-git:
I can not reproduce the error. output on python 2.7, 3.6 and 3.7 is the following:
```console
$ python docs/gallery/domain/icephys.py
/home/matthias/pynwb/local/lib/python2.7/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
before write
tzlocal()
2018-10-05T15:18:21.445020+02:00
tzlocal()
2018-10-05T15:18:21.446525+02:00
after read
tzoffset(None, 7200)
2018-10-05T15:18:21.445020+02:00
tzoffset(None, 7200)
2018-10-05T15:18:21.446525+02:00
```

Maybe the problem is related to Anaconda2. Maybe anaconda uses an old pynwb lib? Debugging the traceback with `pudb` would be good:

```console
$ python -m pudb docs/gallery/domain/icephys.py
```

 @t-b Should I test again on Windows or can you try?

src/pynwb/file.py
t-b: https://docs.python.org/3.1/library/warnings.html

ukos-git:
I have added a function for adding timezone information. This way it is easier to add the warning in a uniforma way.

I'm not sure why everything is acting differently on Windows. This is probably connected with the other error you get with your modified icephys file. I will install python for Windows for testing. I wish there was a way for holding windows10 in a docker file :-)

my h5dump by the way looks like this:
```console
$ h5dump --dataset=/file_create_date --dataset=/session_start_time icephys_example.nwb
HDF5 "icephys_example.nwb" {
DATASET "/file_create_date" {
   DATATYPE  H5T_STRING {
      STRSIZE H5T_VARIABLE;
      STRPAD H5T_STR_NULLTERM;
      CSET H5T_CSET_UTF8;
      CTYPE H5T_C_S1;
   }
   DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
   DATA {
   (0): "2018-10-08T15:05:00.155631+02:00"
   }
}
DATASET "/session_start_time" {
   DATATYPE  H5T_STRING {
      STRSIZE H5T_VARIABLE;
      STRPAD H5T_STR_NULLTERM;
      CSET H5T_CSET_UTF8;
      CTYPE H5T_C_S1;
   }
   DATASPACE  SCALAR
   DATA {
   (0): "2018-10-08T15:05:00.152845+02:00"
   }
}
}
```

src/pynwb/form/build/map.py
```diff
@@ -568,6 +569,8 @@ def __convert_value(self, value, spec):
                         string_type = text_type
                     elif 'ascii' in spec.dtype:
                         string_type = binary_type
+                    elif 'isodatetime' in spec.dtype:
+                        string_type = datetime.isoformat
```
ukos-git, line 585: we could also set the accuracy here

ukos-git:
How should we deal with this test case error?

```console
FAIL: test_build (ui_write.test_nwbfile.TestNWBFileIO)
  File "/home/circleci/project/tests/integration/ui_write/base.py", line 49, in test_build
    self.assertDictEqual(result, self.builder)
AssertionError: 
-  'file_create_date': {'attributes': {}, 'data': ['2017-04-15T12:00:00+00:00']},
?                                                             ^   ------

+  'file_create_date': {'attributes': {}, 'data': ['2017-04-15 12:00:00']},
? 
```

t-b:
> Should I test again on Windows or can you try?

I can don that. Or even better, let CI handle that.

ukos-git:
@t-b circleci is not passing and I don't know why.
I removed the arrow package, as I suspected that this was the source of the problem (no dpkg euqivivalent) but the error remains.

t-b:
@ukos-git I've no clue. Does the tests pass locally with at least one python version? If yes, let's pretend it has nothing to do with this PR and let's finish up with PR.

oruebel:
@dorukozturk could you have a look at the CircleCI error in this PR?

ukos-git:
I don't understand the error. It is a missing directory issue after the successful build of the wheels package.

t-b:
@ukos-git Could you rebase onto newest dev branch now that #659 is merged?

ukos-git:
now waiting for #661

t-b (approved):

oruebel:
On a separate note, when implementing changes that affect other parts of our ecosystem, it will be useful if you could create corresponding issue tickets so that folks are aware of the changes. In this case, the addition of the isodatetime dtype in the schema language affects  both MatNWB and the nwb-schema docs. I just created the following two issues for this:

https://github.com/NeurodataWithoutBorders/nwb-schema/issues/195
https://github.com/NeurodataWithoutBorders/matnwb/issues/74

t-b:
@oruebel Thanks for the explanation and creating of the issues.

src/pynwb/data/nwb.file.yaml
```diff
@@ -15,12 +15,16 @@ groups:
...
+      COMMENT:
+        - The date is stored in UTC with local timezone offset as ISO 8601
+          formatted strings: 2018-09-28T14:43:54.123+02:00
+        - Dates stored in UTC end in "Z" with no timezone offset.
+        - Date accuracy is up to milliseconds.
+        - The file can be created after the experiment was run, so this may
+          differ from the experiment start time.
+        - Each modification to the nwb file adds a new entry to the array.'
     dims:
     - '*unlimited*'
```
oruebel, line 29: I think this should be something like:
```
dims:
- timestamps
```

src/pynwb/file.py
```diff
@@ -492,6 +493,18 @@ def set_electrode_table(self, **kwargs):
...
 
 
+def _add_missing_timezone(date):
+    """
+    Add local timezone information on a datetime object if it is missing.
+    """
+    if not isinstance(date, datetime):
+        raise ValueError("require datetime object")
+    if date.tzinfo is None:
+        warn("Date is missing timezone information. Updating to local timezone.")
```
oruebel, line 503: I think it would be good to say in the warning text what the local timezone.

oruebel (approved):
Looks in general good to me.

bendichter (approved):

Issue #638, opened 2018-10-03
fix links in documentation


ukos-git:
Hello.
I am new to pynwb and was reading about the software architecture on readthedocs.

I noticed some broken links and some typos for example in [formio description](https://pynwb.readthedocs.io/en/latest/overview_software_architecture.html#formio)

I have collected my changes in a PR. Hope this will be a quick one, as it does not change (or add) anything functional.

PR #642, opened 2018-10-03
small documentation cleanup

ukos-git:
## Motivation

The documentation was missing some links and some docval documentation items had typos. This  improves the sphinx documentation for higher readability.

I am aware that those commits do not change anything functional so I collected them to not waste precious developer's time. The Issue is #638

The documentation, though, needs a deeper review if it should be easier understandable.

## How to test the behavior?

sphinx-build has passed.

## Checklist

- [X ] Have you ensured the PR description clearly describes problem and the solution?


docs/source/overview_software_architecture.rst
```diff
@@ -88,7 +87,7 @@ Spec
    * :py:class:`pynwb.spec.NWBDatasetSpec` - specification for dataset (like and n-dimensional array). Specifies data type, dimensions, etc.
    * :py:class:`pynwb.spec.NWBLinkSpec` - specification for link (like a POSIX soft link)
    * :py:class:`pynwb.form.spec.spec.RefSpec` - specification for references (References are like links, but stored as data)
-   * :py:class:`pynwb.spec.NWBDtypeSpec` - specification for compound data types. Used to build complex data type specification, e.g., to define tables (used only in ``DatasetSpec``)
+   * :py:class:`pynwb.spec.NWBDtypeSpec` - specification for compound data types. Used to build complex data type specification, e.g., to define tables (used only in :py:class:`DatasetSpec <pynwb.spec.NWBDatasetSpec>`)
```
oruebel: Can you change this to "... (used only in :py:class:`DatasetSpec <pynwb.from.spec.spec.DatasetSpec>` and correspondingly :py:class:`DatasetSpec <pynwb.spec.NWBDatasetSpec>`)

docs/source/overview_software_architecture.rst
```diff
@@ -176,8 +176,8 @@ TypeMap
 BuildManager
 ^^^^^^^^^^^^
 
-* Responsible for memoizing ``Builders`` and ``Containers``
```
oruebel, line 179: "memoizing" is actually the proper term here I believe

src/pynwb/__init__.py
```diff
@@ -186,7 +186,7 @@ def validate(**kwargs):
 
 class NWBHDF5IO(HDF5IO):
 
-    @docval({'name': 'path', 'type': str, 'doc': 'the path to the HDF5 file to write to'},
+    @docval({'name': 'path', 'type': str, 'doc': 'the path to the HDF5 file'},
             {'name': 'mode', 'type': str,
```
oruebel, line 190: this should be corrected to "to write to or read from"

src/pynwb/form/backends/hdf5/h5tools.py
```diff
@@ -29,7 +29,7 @@
 
 class HDF5IO(FORMIO):
 
-    @docval({'name': 'path', 'type': str, 'doc': 'the path to the HDF5 file to write to'},
```
oruebel, line 32: this should be corrected to "to write to or read from"

oruebel (commented):
Thanks for helping out with documentation. See comments below for a few fixes.

docs/source/overview_software_architecture.rst
ukos-git: ok, i am no native speaker, my background for this is the latin word memorare, which is with an "r" and means to remember sth. But the english language is not very straight forward in its spelling so I could be mistaking.
[memoizing](https://www.dict.cc/?s=memoizing) imho would mean, something like making a memo out of it, whereas [memorizing](https://www.dict.cc/?s=memorizing) would mean to remember it.

src/pynwb/__init__.py
ukos-git: My intention was that the mode flag should describe the action and path should only describe the target.

docs/source/overview_software_architecture.rst
oruebel: It's a computer science thing 😉  https://en.wikipedia.org/wiki/Memoization

src/pynwb/__init__.py
oruebel: Sorry, I read the edit the wrong way, you had removed the "to write to" part not added it. Yes, this is fine.

docs/source/overview_software_architecture.rst
ukos-git: That is right, We should not mix the two of them here.
Also note, that you can use the tilde operator for shortcuts like this:

```rst
(used only in :py:class:`~pynwb.form.spec.spec.DatasetSpec` and
     correspondingly :py:class:`~pynwb.spec.NWBDatasetSpec`)
```
which gives

> (used only in **DatasetSpec** and correspondingly **NWBDatasetSpec**)

While I was at it, I also added the tilde to all of the list items above for overall better readability as the origin of those classes is not important here for the description. And I fixed the mixed line-wrapping here.

Btw.
Is there any `textwidth` setting in your coding guidelines? - In pynwb python files and also rst files it seems to be quite random (looking ugly)

docs/source/overview_software_architecture.rst
ukos-git: How we say in germany: "Again what learnt"
I added the external wikipedia link to the word in the docu.

oruebel (approved):

PR #648, opened 2018-10-05
epochs df roundtrip bugfix and test

NileGraddis:
This is a bugfix for Epochs.to_dataframe(). On roundtrip to a file the timeseries column is coming back as h5py region references, which broke the timeseries column. This pr fixes that issue and expands the test case to cover this issue.

thanks @nicain for catching this


nicain (dismissed):
LGTM. This PR will unblock me.

nicain:
@NileGraddis I sent you a PR onto your fork to fix the flake8; should pass right through, so go ahead and merge https://github.com/NileGraddis/pynwb/pull/1 and then I will merge here.

nicain (approved):

Issue #651, opened 2018-10-05
python3.7 install numpy 1.14.2


ukos-git:
## 1) Bug

Can not install requirements for python 3.7 using pip

### Steps to Reproduce
```
virtualenv --python=python3.7 env
pip install -r requirements.txt -r requirements-dev.txt
```
### Environment
Please describe your environment according to the following bullet points.
- **Python Executable:** python
- **Python Version:** Python 3.7.1rc1
- **Operating System:**Ubuntu 18.04
- **Pynwb Version:** 430ae547afb1f454c6e7df636914ea082fc36fc2

### Problem/Use Case
There seem to be CPython issues with [numpy 1.14.2](https://github.com/numpy/numpy/issues/10500#issuecomment-393949228). numpy and pandas work together.

Pandas package here is fairly old. Updating both, numpy, and pandas to recent versions fixes the problem. Note that current numpy 1.15 throws some warnings to the cli: #650 so 

I suggest an update of the following packages (as they depend on each other):

* numpy==1.14.2 --> numpy==1.14.6 (this is the last 1.14, although it should be fixed already in 1.14.3)
* pandas==0.19.2 --> pandas==0.23.4 (not sure which version is suitable but pandas is currently not a base functionality)

## Checklist

- [X] Have you ensured the feature or change was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues/) ?
- [X] Have you included a brief and descriptive title?
- [X] Have you included a clear description of the problem you are trying to solve?
- [X] Have you included a minimal code snippet that reproduces the issue you are encountering?

ukos-git:
sorry I just overlooked, that pynwb does not support 3.7, right? - I just tested and there are no problems on 3.6.

Probably closing this would be best, except if the package update is fine.

t-b:
@ukos-git Solved in #652. Please reopen if not.

PR #652, opened 2018-10-05
update requirements.txt for python3

ukos-git:
numpy 1.14.2 has problems installing in python3.7.
This fix lets pynwb install again on a clean venv.

see #651


dorukozturk:
See #659

dorukozturk:
Ok this started to make sense. Right now we are installing tox 3.5.1 on conda. Though tox released 3.5.2 which fixes this issue. So I will revert my PR from yesterday because the latest tox is working for us now. Once they get https://github.com/conda-forge/tox-feedstock/pulls in we should be good.

ukos-git:
@dorukozturk this issue is not quite related. It is targetting towards a wheels->cpython compiler issue https://github.com/numpy/numpy/issues/10500#issuecomment-393949228 

But it is only happening in python 3.**7** , which is not supported by pynwb but the changes will not hurt. We could either close this issue here due to the not supported python version or update the numpy package (and pandas, which goes together) to the newer subsubversion as suggested.

dorukozturk:
@ukos-git I was just pointing why the miniconda tests are failing. I haven't checked the context of this PR yet :)

ukos-git:
I understand. I was just clarifying.

ajtritt (approved):

ajtritt:
Thanks for tackling this @ukos-git

PR #661, opened 2018-10-09
Use the latest tox since the issue is solved as of 3.5.2

dorukozturk:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ references:
           name: Run the tests
           command: |
             . ../venv/bin/activate
-            pip install tox==3.4.0
+            pip install tox
             tox -e $TEST_TOX_ENV
           # Install is expected to be quick. Increase timeout in case there are some network issues.
           # While pip installing tox does not output by default. Circle thinks task is dead after 10 min.


bendichter:
should I be concerned that the minicondas are still failing? @ukos-git

ukos-git:
strange thing is that the minicondas are still failing in this MR.

ukos-git:
ok, i was mixing everything. My fault.

I thought that the miniconda thing was fixed here but the solution here is to revert to normal tox. This solution is actually a good one.

So we probably have to wait for conda to pull the latest tox. Waiting is probably the only solution.

bendichter:
@ukos-git 
> wait for conda to pull the latest tox

I don't quite understand what you mean by this. Can you elaborate?

ukos-git:
conda keeps the repositories in another collection than pip. This is probably why conda is a little bit behind with the tox package. The last push was https://github.com/conda-forge/feedstocks/commit/76420f3d2482dc1958bbbeb6a09c0cdc021f3b06

See here for all commits to tox
https://github.com/conda-forge/feedstocks/search?q=tox&type=Commits

dorukozturk:
@bendichter @ukos-git Seems like everything is good now. We should merge this and rebase other branches. Thanks for your patience 👍

ukos-git:
Is tox considered unstable or do we need any recent Features from it? Otherwise maybe fixing it at the next stable release would be good to avoid problems like this.

I am unsure what is the right thing here.

dorukozturk:
Actually it has been very stable. It is a very active project and this is the first time something like this is happening. I think either way works.

bendichter (approved):

ukos-git:
Ok. So lets keep it the way it is now.

PR #678, opened 2018-10-19
Remove enum34 dependency from requirements-dev.txt (fixes #677)

### LINKED ISSUE #677: Do not install enum34 (from requirements-dev.txt) in Python>3.4
## 1) Bug

`requirements-dev.txt` installs enum34 irrespective of the python version.

enum34 provides python 3.4-like enum functionality to earlier versions of python. If it is installed on python >3.4, it can actually break enums (https://bitbucket.org/stoneleaf/enum34/issues/19/enum34-isnt-compatible-with-python-36)

We don't seem to use any enums in pynwb, but flake8 depends on it. 

Here are 2 proposed solutions

1) If requirements-dev.txt is meant to freeze requirements, we should add an environment marker to the enum34 line to avoid installing it inappropriately:
   ```
   enum34==1.1.6 ; python_version < '3.4'
   ```
2) If we don't care about capturing the version of enum34, we could just remove this line, and let flake8 pull in an appropriate version of enum34. 

Solution (2) would also allow for the use of conda to install all the requirements (e.g. `conda create -c conda-forge -n pynwb python=3.6 --file requirements.txt --file requirements-dev.txt`). Currently that command fails because conda (correctly) detects a conflict between python==3.6 and enum34==1.1.6. Solution (1) won't work in this case since conda doesn't know how to interpret environment markers. (pip is not as picky, and will happily install enum34 on python 3.5 or 3.6). 
## Checklist

- [X] Have you ensured the feature or change was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues/) ?
- [] Have you included a minimal code snippet that reproduces the issue you are encountering?

tjd2002:
Fixes #677


oruebel (dismissed):

dorukozturk:
@tjd2002 Can you update the [docs](https://pynwb.readthedocs.io/en/latest/update_requirements.html#requirements-dev-doc-txt) so that the next developer knows why enum is missing from the requirements file? Some kind of a note or even mentioning the github issue there would be fantastic.

tjd2002:
Thanks for the pointer @dorukozturk, I didn't know about that doc. I have tried to update it to explain this special case.

tjd2002:
I don't want to overengineer things (especially since this is not currently breaking anything), but would it be appropriate to add a test that fails if `enum34` is accidentally installed in python>3.4?

dorukozturk:
I think we don't need to worry about it for now. How about we add it, if someone runs into this issue?

bendichter (approved):

PR #681, opened 2018-10-24
hdf5io accepts files

### LINKED ISSUE #680: Allow NWBHDF5IO and HDF5IO to accept h5py.file objects
These classes currently require a path and construct an `h5py.File` object at that path when their `open` method is called. It would be handy to have the option of passing in a pre-constructed `h5py.File` object (skipping the construction of a new one in `open`). This would not impact existing functionality, but would allow us to use h5py features for free without explicitly implementing support for them (for instance: using the core driver for fast testing).

NileGraddis:
Resolves #680 

Adds an optional argument to the HDF5IO and NWBHDF5IO constructors for an existing h5py.File object. This change:
- does not impact existing APIs and users
- allows us easy access to a bunch of h5py functionality


ajtritt:
Is there a reason you didn't go the route of just letting the user pass in kwargs that get passed into h5py.File? This way HDF5IO still controls opening/closing of the underlying HDF5 file, and h5py dependencies stay contained to pynwb.form.backends.hdf5

NileGraddis:
@ajtritt Yes. By passing in a constructed object you can do stuff like defining an `h5py.File` subclass with any behavior you desire and pass an instance in to the io object constructor. You could also pass in a mock (well, a mocked-out `h5py.File`) for testing.

Your concern about the `h5py` dependency leaking out of `pynwb.form.backends.hdf5` is warranted. However, I would point out that the leak is only to support the docval type requirement on `NWBHDF5IO` (and the test of course). Since `NWBHDF5IO` is a subclass of `HDF5IO` and `HDF5IO` has an h5py dependency, it seems not totally crazy for `NWBHDF5IO` to also have this dependency. Of course, we could just make the docval requirement on `NWBHDF5IO` generic.

With these changes `HDF5IO` does still control the opening and closing of the underlying file except when a user opts to pass in their own h5py.File. In that case, they presumably want to control the opening of the file themselves.

NileGraddis:
weird, the flake8 check is complaining about:
```
src/pynwb/core.py:389:-696: W605 invalid escape sequence '\_'
src/pynwb/core.py:389:-694: W605 invalid escape sequence '\_'
src/pynwb/core.py:389:-685: W605 invalid escape sequence '\_'
src/pynwb/core.py:389:-683: W605 invalid escape sequence '\_'
src/pynwb/core.py:389:-304: W605 invalid escape sequence '\_'
src/pynwb/core.py:389:-302: W605 invalid escape sequence '\_'
src/pynwb/core.py:389:-293: W605 invalid escape sequence '\_'
src/pynwb/core.py:389:-291: W605 invalid escape sequence '\_'
src/pynwb/core.py:389:-115: W605 invalid escape sequence '\_'
src/pynwb/core.py:389:-113: W605 invalid escape sequence '\_'
src/pynwb/core.py:389:-100: W605 invalid escape sequence '\_'
src/pynwb/core.py:389:-98: W605 invalid escape sequence '\_'
src/pynwb/form/build/map.py:891:13: F841 local variable 'v' is assigned to but never used
```
... but I did not touch those files. After looking around I notice that the check for this PR is running with [flake8 3.6](https://circleci.com/gh/NeurodataWithoutBorders/pynwb/10307?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) [rather than 3.5](https://circleci.com/gh/NeurodataWithoutBorders/pynwb/10266?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). I can replicate these errors locally if I upgrade. Cool.

dorukozturk:
@NileGraddis A small unrelated note. Can you squash your flake8 related commits when this PR is ready?

NileGraddis:
@dorukozturk sure
@ajtritt Does this look reasonable to you?

tests/integration/ui_write/test_nwbfile.py
```diff
@@ -269,3 +269,7 @@ def addContainer(self, nwbfile):
 
     def getContainer(self, nwbfile):
         return nwbfile.epochs
+
```
ajtritt: This creates a diamond inheritance pattern. Why is this necessary?

ajtritt (changes requested):
Please address the diamond inheritance in the test suite.

tests/integration/ui_write/base.py
```diff
@@ -188,3 +191,28 @@ def addContainer(self, nwbfile):
     def getContainer(self, nwbfile):
         ''' Should take an NWBFile object and return the Container'''
         return nwbfile.get_acquisition(self.container.name)
+
+
+class TestMapRoundTripWithExistingFileObject(TestMapRoundTrip):
```
ajtritt: Can't you just implement setUpContainer, addContainer, getContainer?

ajtritt: I appreciate the desire to reuse code, but this has made things much more complicated than necessary. There's a lot of extra code and logic in test_roundtrip solely for testing a single feature. I prefer that we keep the general methods upon which the entire integration test suite depends as simple as possible. 

Is there a reason you didn't just duplicate test, and modify it to pass in an h5py.File?

https://github.com/NeurodataWithoutBorders/pynwb/blob/5bcdc8316d81d099514a3df95ad1d884a2c49fd6/tests/integration/test_io.py#L65

ajtritt (dismissed):

NileGraddis:
@ajtritt I don't mind moving the test to io - that makes more sense I think. I had put the modifications in base so that it would be easy to switch any arbitrary test (or all of them) to run this case.

ajtritt:
@NileGraddis 

> I had put the modifications in base so that it would be easy to switch any arbitrary test (or all of them) to run this case.

If it becomes clear that we need to run the roundtrip tests in both modes, I'll eat my own words and plead with you to put that code back ;)

ajtritt (approved):

PR #693, opened 2018-10-30
add Image type and subtypes GrayScaleImage, RGBImage, RGBAImage

### LINKED ISSUE #129: request: `StaticImage`

### LINKED ISSUE #143: Missing neurodata_type Image
# Bug

The neurodata_type ``Image`` is used as part of ``Images`` but does not seem to be defined. The following code from nwb.base.yaml:

```
...
197   datasets:
198     - doc: Images stored in this NWBDataInterface
199       neurodata_type_inc: Image
200       quantity: '+'
...
203   neurodata_type_def: Images
...
```

We probably need to update the ``Images`` type to replace ``Image``. Also should ``Images`` be moved to nwb.image.yaml?

bendichter:
* add Image to core.base
* add GrayscaleImage, RGBImage, and RGBAImage to core.image
* add pynwb support for Image, Images, GrayscaleImage, RGBImage, and RGBAImage

## Motivation

fix https://github.com/NeurodataWithoutBorders/nwb-schema/issues/143, fix https://github.com/NeurodataWithoutBorders/nwb-schema/issues/129

## Checklist


src/pynwb/image.py
```diff
@@ -268,3 +268,60 @@ def __init__(self, **kwargs):
...
+            {'name': 'resolution', 'type': float, 'doc': 'pixels / cm', 'default': None},
+            {'name': 'description', 'type': str, 'doc': 'description of image', 'default': None})
+    def __init__(self, **kwargs):
+        name, source, resolution, data, description = popargs(
+            'name', 'source', 'resolution', 'data', 'description', kwargs)
+        super(Image, self).__init__(name=name, source=source)
+
+        self.resolution = resolution
+        self.data = data
+        self.description = description
```
oruebel: You could just pass resolution, data, description to the  super(Image, self) constructor as well to avoid the duplicate code for assigning them in the subclasses. This is the case for RGBAImage, RGBImage, and GrayscaleImage

src/pynwb/image.py
```diff
@@ -268,3 +268,60 @@ def __init__(self, **kwargs):
         self.distance = distance
         self.field_of_view = field_of_view
         self.orientation = orientation
+
+
+@register_class('GrayscaleImage', CORE_NAMESPACE)
+class GrayscaleImage(Image):
+    __nwbfields__ = ('data', 'resolution', 'description')
```
oruebel: For GrayscaleImage ``description`` is part of the __nwbfields__ and for RGBImage, RGBAImage, and Image it is not. I assume it should not be part of the fields. I assume ``description`` should be removed from ``__nwbfields_`` in GrayscaleImage

oruebel (dismissed):

src/pynwb/image.py
bendichter: good catch. I think it should be added

src/pynwb/image.py
bendichter: yeah good call

oruebel (dismissed):
Pending pass of the CI tests ;-)

oruebel:
I guess I was too soon with approving it. Looks like the CI fails. But other than the failed tests it looked good to me.

bendichter:
Do we want to enforce that `Image` is only to be used as an abstract class? i.e. an image must be declared as a `GrayscaleImage`, `RGBImage`, `RGBAImage`, or other subtype of `Image`?

src/pynwb/data/nwb.base.yaml
```diff
@@ -91,6 +91,24 @@ groups:
 - neurodata_type_def: NWBDataInterface
   neurodata_type_inc: NWBContainer
   doc: An abstract data type for differentiating experimental data from metadata
+- neurodata_type_def: Image
```
ajtritt, line 81: Why is Image a group and not a dataset? Making a group requires the additional dataset. This doubles the number of objects need to store an image

bendichter:
Yeah that was my original plan but I didn't know how to get Images to work
with it

ajtritt:
@bendichter I just submitted a PR to this branch. I thought seeing the diff would be easier than trying to explain it over text.

bendichter:
@ajtritt can you review this?

src/pynwb/image.py
```diff
@@ -251,3 +251,39 @@ def __init__(self, **kwargs):
...
+            {'name': 'data', 'type': ('array_data', 'data'), 'doc': 'data of image', 'shape': (None, None, 3)},
+            {'name': 'resolution', 'type': float, 'doc': 'pixels / cm', 'default': None},
+            {'name': 'description', 'type': str, 'doc': 'description of image', 'default': None})
+    def __init__(self, **kwargs):
+        super(RGBImage, self).__init__(**kwargs)
+
+
+@register_class('RGBAImage', CORE_NAMESPACE)
+class RGBAImage(Image):
+    __nwbfields__ = ('data', 'resolution', 'description')
```
ajtritt: remove this. These properties will be inherited from the base Image class.

src/pynwb/image.py
```diff
@@ -251,3 +251,39 @@ def __init__(self, **kwargs):
...
+            {'name': 'data', 'type': ('array_data', 'data'), 'doc': 'data of image', 'shape': (None, None)},
+            {'name': 'resolution', 'type': float, 'doc': 'pixels / cm', 'default': None},
+            {'name': 'description', 'type': str, 'doc': 'description of image', 'default': None})
+    def __init__(self, **kwargs):
+        super(GrayscaleImage, self).__init__(**kwargs)
+
+
+@register_class('RGBImage', CORE_NAMESPACE)
+class RGBImage(Image):
+    __nwbfields__ = ('data', 'resolution', 'description')
```
ajtritt: remove this. These properties will be inherited from the base Image class.

src/pynwb/image.py
```diff
@@ -251,3 +251,39 @@ def __init__(self, **kwargs):
         self.distance = distance
         self.field_of_view = field_of_view
         self.orientation = orientation
+
+
+@register_class('GrayscaleImage', CORE_NAMESPACE)
+class GrayscaleImage(Image):
+    __nwbfields__ = ('data', 'resolution', 'description')
```
ajtritt: remove this. These properties will be inherited from the base Image class.

ajtritt (changes requested):
Some extra \_\_nwbfields\_\_, but other than that it looks good.

PR #694, opened 2018-10-30
make Image basetype a dataset

ajtritt:
## Motivation

make Image base type a dataset

## Checklist

### DIFF
--- a/src/pynwb/data/nwb.base.yaml
+++ b/src/pynwb/data/nwb.base.yaml
@@ -78,21 +78,8 @@ datasets:
   - name: description
     dtype: text
     doc: Description of what this table region points to
-groups:
-- neurodata_type_def: NWBContainer
-  doc: The attributes specified here are included in all interfaces.
-  attributes:
-  - name: help
-    dtype: text
-    doc: Short description of what this type of NWBContainer contains.
-  - name: source
-    dtype: text
-    doc: Path to the origin of the data represented in this interface.
-- neurodata_type_def: NWBDataInterface
-  neurodata_type_inc: NWBContainer
-  doc: An abstract data type for differentiating experimental data from metadata
 - neurodata_type_def: Image
-  neurodata_type_inc: NWBDataInterface
+  neurodata_type_inc: NWBData
   doc: Image base type.
   attributes:
   - name: help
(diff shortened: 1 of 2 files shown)


bendichter:
@ajtritt I like this if we can make it work but right now there's an error:
`TypeError: incorrect type for 'neurodata_type_inc' (got 'NWBDatasetSpec', expected 'NWBGroupSpec or str')`

bendichter:
@ajtritt 

Can datasets have datasets? Or should it be:

```yaml
datasets:
- neurodata_type_def: GrayscaleImage
  neurodata_type_inc: Image
  doc: Grayscale image. Origin in top left.
  shape:
    - null
    - null
  dims:
    - 'y'
    - 'x'
- neurodata_type_def: RGBImage
  neurodata_type_inc: Image
  doc: Color image. Origin in top left.
  shape:
  - null
  - null
  - 3
  dims:
  - 'y'
  - 'x'
  - 'R,G,B'
- neurodata_type_def: RGBAImage
  neurodata_type_inc: Image
  doc: Color image with transparency. Origin in top left.
  shape:
  - null
  - null
  - 4
  dims:
  - 'y'
  - 'x'
  - 'R,G,B,A'
```

ajtritt:
@bendichter yes, you are correct. I sloppily copy-pasted... I just changed it though.

bendichter:
@ajtritt How do you interact with these types of datasets in `pynwb`? I don't see any such datasets in any of the schema namespace files other than abstract types in `base`

ajtritt:
@bendichter there is an Image base class in file.py

https://github.com/NeurodataWithoutBorders/pynwb/blob/b9a2ef1a7125607edc05f66859cc29f843aa4702/src/pynwb/file.py#L26

It’s not fully implemented right now, but it would just need to be implemented to support the necessary attributes on the dataset, and the necessary subclasses would need to be added.

bendichter:
why specify `Image` in `nwb.base.yaml` and then define the constructor in `file.py`? Wouldn't it make more sense to have the constructor in `base.py`?

ajtritt:
Sure, it probably makes more sense. But, there's no requirement about where specs are and where their API classes are.

PR #695, opened 2018-10-31
NWBContainer: Remove source attribute

t-b:
The source attribute is currently a free text entry designed for
provenance attribution.

Experienced showed that this was either mostly ignored or heavily
misused, e.g. sneaking in additional metadata.

Potential readers also don't benefit from the free text approach as it
is difficult to understand the contents without any rules on how to
interpret it.

So let's remove it.

All current users of source can either switch to description or
comments for NWBContainer.


t-b:
Sorry for the late PR, this was supposed to be finished earlier. And also sorry for the diff, this is as much fun reviewing than doing...

bendichter:
@t-b holy poop you actually did it! Thanks for tackling this, it is a great service to the project. I'll look through it now

oruebel:
@t-b thanks for the help!

nicain:
Many have tried (including myself!), but only @t-b has succeeded!!

bendichter (approved):

t-b:
Thanks all for your kind words!

ajtritt (approved):
f*&# yeah!

PR #700, opened 2018-11-01
fix error when omitting tags

### LINKED ISSUE #699: omitting tag in epochs causes error
## 1) Bug

```python
from pynwb import NWBFile
from datetime import datetime

nwbfile = NWBFile("a file with header data", "NB123A", datetime.now().astimezone())
nwbfile.create_epoch(start_time=0., stop_time=1.)
```

```python
/Users/bendichter/anaconda3/envs/dev_pynwb/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
  File "/Users/bendichter/dev/pynwb/test_electrodes.py", line 9, in <module>
    nwbfile.create_epoch(start_time=0., stop_time=1.)
  File "/Users/bendichter/dev/pynwb/src/pynwb/form/utils.py", line 375, in func_call
    return func(self, **parsed['args'])
  File "/Users/bendichter/dev/pynwb/src/pynwb/file.py", line 387, in create_epoch
    self.epoch_tags.update(kwargs.get('tags', list()))
TypeError: 'NoneType' object is not iterable
```

## Checklist

bendichter:
## Motivation

fix #699 
## Checklist


src/pynwb/file.py
```diff
@@ -384,7 +384,10 @@ def create_epoch(self, **kwargs):
         enclosure versus sleeping between explorations)
         """
         self.__check_epochs()
-        self.epoch_tags.update(kwargs.get('tags', list()))
+        if kwargs['tags'] is None:
+            self.epoch_tags.update(list())
```
ajtritt: I apologize in advance for being picky. This line does nothing. You could use fewer lines of code if you only update if kwargs['tags'] is not None

ajtritt (dismissed):

src/pynwb/file.py
bendichter: Oh ok I didn't know if the empty list was important. Thanks for the improvement!

PR #701, opened 2018-11-02
Add sweeptable for PatchClampSeries

### LINKED ISSUE #499: Adding sweep based information
# Current design

https://github.com/t-b/ipfx/blob/feature/sweep_table/sweeptables.md

# Old design

This is a POC design we've come up during the hackathon at the allen institute.

# Problem

In Icephys it is common to have sweeps (a group of PatchClampSeries belonging
together, were up to two TimeSeries are from one electrode, including
other TimeSeries not related to an electrode (aka TTL channels)).

Currently the concept of sweeps is not present in NWB so linking different
TimeSeries together is not possible.

# Goals

Allow to find the TimeSeries which are from one sweep without having to iterate over all present TimeSeries.

# Solution

Add some more tables

## Sweep Table (DynamicTable)

|Sweep number  | Vector of indizes into the acquisition table (vector column data) |
|------- | ----------------------------------------------- |
|        |                                                 |

## Acquisition Table (DynamicTable)

| Index | Sweep number | Soft link to electrode| Soft link to device | Stimulus (Soft link to TimeSeries) | neurodata_type of Stimulus TimeSeries | Acquisition (Soft link to TimeSeries) | neurodata_type of Acquisition TimeSeries |
 |---|--------|--------|---------------------|------------------------------------|------------------------------------|------------------------------------|------------------------------------|
 |        |        |                     |      |     |                                    |                                    |                                    |                                    |                                    |
 
The columns in the acquisition table could be reduced to the bare minimum: `Index, Stimulus (Soft link to TimeSeries), Acquisition (Soft link to TimeSeries)`.

Both tables live in `/general/intracellular_ephys`.
In the acquisition table entries which don't make sense are empty (e.g. no link to an electrode for TTL data).
The first column for both tables is an unique index.

t-b:
Done.

Can it be that cells in VectorData columns must have all the same number of entries? I could not append to a cell for changing that.

Nevertheless I've found a way to achieve what I wanted.

Closes #499.

### DIFF
--- a/docs/gallery/domain/icephys.py
+++ b/docs/gallery/domain/icephys.py
@@ -91,6 +91,12 @@ vcs = VoltageClampSeries(
 nwbfile.add_acquisition(vcs)
 
 ####################
+# If you are interested in all PatchClampSeries with a given sweep number, use get_series()
+# exposed via the :py:meth:`~pynwb.icephys.SweepTable.get_series` attribute.
+
+series = nwbfile.sweep_table.get_series(15)
+
+####################
 # .. _icephys_writing:
 #
 # Once you have finished adding all of your data to the :py:class:`~pynwb.file.NWBFile`,
--- a/docs/gallery/domain/icephys.py
+++ b/docs/gallery/domain/icephys.py
@@ -70,7 +70,7 @@ from pynwb.icephys import CurrentClampStimulusSeries
 
 ccss = CurrentClampStimulusSeries(
     name="ccss", data=[1, 2, 3, 4, 5], unit='A',
-    starting_time=123.6, rate=10e3, electrode=elec, gain=0.02)
+    starting_time=123.6, rate=10e3, electrode=elec, gain=0.02, sweep_number=15)
 
 nwbfile.add_stimulus(ccss)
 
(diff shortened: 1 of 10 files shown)


bendichter:
my guess is it has to do with `__nwbfields__`. I'll take a look

bendichter:
@t-b there you go. All fields that are to be saved to disk need to be listed in `__nwbfields__`. The `'child'` field is needed for any `neurodata_type` that is not a link. I don't think this is well documented, so that's something we'll need to fix.

t-b:
@bendichter 100e100 thanks. That would have taken a while to figure out.

t-b:
@bendichter Sorry for bugging again. Any clue why `__columns__` does not work in the SweepTable class?

I'm getting

```
id=id <class 'pynwb.core.ElementIdentifiers'>
Fields:
, colnames=(), columns=[]
```

ajtritt:
@t-b Try adding `"required": True` to the two dicts in `__columns__`. By default, DynamicTable assumes specified columns are optional, and therefore does not add them to table unless they are populated.

bendichter:
@t-b It seems like this is ready for review. Should we removed the WIP prefix in the title?

bendichter (commented):
It looks like your workflow is to add an extra `sweep_number` arg to the data constructors, and then with `_update_sweep_table` automatically go through and build the `Sweeps` table based on these values. Our workflow so far has been to have the user manually add data or references to data row-by-row (`trials`, `epochs`, `units`, and calcium imaging rois all follow this convention). I don't think there is anything inherently wrong with the way you are doing it, but it may be confusing to users to have these different workflows in pynwb. Is there a particular reason you went this route?

ajtritt (approved):
nice work. thanks @t-b

t-b:
@bendichter Sorry for the late reply, I've been at a conference.

>  Our workflow so far has been to have the user manually add data or references to data row-by-row (trials, epochs, units, and calcium imaging rois all follow this convention). I don't think there is anything inherently wrong with the way you are doing it, but it may be confusing to users to have these different workflows in pynwb. Is there a particular reason you went this route?

Well in my case the sweeptable stores the connection between a PatchClampSeries and the sweepnumber. And there is nothing more the user needs to add, so doing it implicitly is good way to keep the data consistent.

PR #707, opened 2018-11-03
Remove unused _ancestry class variable

t-b:
This is not written to NWB anymore.


bendichter:
Wait I actually like this. Is there currently no way to know from the nwb file how a class was derived? Would analysis tools benefit from access to this variable? Eg knowing that an extension derived from TimeSeries? If so it would probably be easier to do this automatically.

oruebel:
ancestry should certainly not be stored in the file. That is the role of the schema and having the ancestry separately in the file is problematic.

bendichter:
I guess the cached schema would accomplish this

bendichter:
I don't really understand what's "problematic" about this though

t-b:
From the cached spec in the file the ancestry can always be recovered.

ajtritt:
@bendichter If you want ancestry, we could have the register_class decorator add it as a class attribute (this is how the neurodata_type class attribute gets added). Hardcoding it as a comma delimited string is problematic for long term maintenance and consistency. It’s just a field that would need to be kept in sync with the schema

bendichter:
@ajtritt thanks for the explanation. @oruebel also pointed me to [the discussion in the change logs](https://nwb-schema.readthedocs.io/en/latest/format_release_notes.html#ancestry). Let's remove this inactive code. Instead, how would you all feel about implementing a function that is analogous to `isinstance` but for neurodata_types instead of python classes using the schema?

bendichter (approved):

ajtritt (approved):
looks good to me. Thanks @t-b

ajtritt:
>Instead, how would you all feel about implementing a function that is analogous to isinstance but for neurodata_types instead of python classes using the schema?

@bendichter that seems reasonable. I prefer we address that under a new issue though

t-b:
@ajtritt @bendichter "Merge pull request" <-- Click here

PR #709, opened 2018-11-06
Add NWBFile.timestamps_reference_time field.

### LINKED ISSUE #49: modify timestamps specification to allow storing of different time formats
Users would like to store timestamps in both absolute and relative time. To achieve this, the following modification is proposed:

**Proposal**
Add an additional top-level field called  _timestamps_offset_ - the time difference by which timestamps are offset from being relative to the file's global starting time i.e. `session_start_time`. This value is 0 by default i.e. timestamps are directly relative to `session_start_time`. If this value is not 0, then to obtain timestamps that are directly relative to `session_start_time`, one must subtract this offset.

**Advantages**
- This allows users to store timestamps in absolute POSIX time without having to set `session_start_time` to "00:00:00 1 January 1970" i.e. the unix epoch.  Saving timestamps as absolute POSIX timestamps can be done by setting `timestamps_offset` to the  POSIX timestamp equivalent of `session_start_time`. Saving timestamps as relative to `session_start_time` (the current specification) can be achieved by setting `timestamps_offset` to 0. 

- This also makes the specification easier to develop against, since this will explicitly specify the offset to obtain relative timestamps, eliminating the need for APIs to guess based on range. 

**Disadvantages**
- Users and APIs will need to be aware of the additional field. 

**Modifications**
- An alternative to this could be to allow each timestamps dataset to have it's own _timestamps_offset_ field.

tjd2002:
## Motivation

This introduces a new top-level NWBFile field called `Timestamps_reference_time`, as discussed at fix https://github.com/NeurodataWithoutBorders/nwb-schema/issues/49

This field allows users to specify a date and time corresponding to time zero for all timestamps in the nwb file. The `session_start_time` field currently does double-duty in this regard, representing both the start time of the current session, and time zero for timestamps.

## How to test the behavior?
Tests are included in this PR. Briefly, for a file using POSIX timestamps
```python
from datetime import datetime
from dateutil import tz
from pynwb import NWBFile

session_start_time        = datetime(2017, 10, 31,  9, 59,  0, tzinfo=tz.gettz('US/Pacific'))
timestamps_reference_time = datetime(1970,  1,  1,  0,  0,  0, tzinfo=tz.tzutc())

nwbfile = NWBFile('session description', 'TEST124', session_start_time,
                    timestamps_reference_time=timestamps_reference_time)
```

## Checklist

### Credits:
In collaboration with @AngevineMiller

### DIFF
--- a/src/pynwb/data/nwb.file.yaml
+++ b/src/pynwb/data/nwb.file.yaml
@@ -18,7 +18,7 @@ groups:
     dtype: isodatetime
     doc: 'A record of the date the file was created and of subsequent modifications.
       COMMENT:
-        - The date is stored in UTC with local timezone offset as ISO 8601
+        - The date is stored in UTC with local timezone offset as ISO 8601 extended format
           formatted strings: 2018-09-28T14:43:54.123+02:00
         - Dates stored in UTC end in "Z" with no timezone offset.
         - Date accuracy is up to milliseconds.
--- a/src/pynwb/data/nwb.file.yaml
+++ b/src/pynwb/data/nwb.file.yaml
@@ -44,6 +44,14 @@ groups:
         - The date is stored in UTC with local timezone offset as ISO 8601
           formatted string: 2018-09-28T14:43:54.123+02:00
         - Dates stored in UTC end in "Z" with no timezone offset.
+        - Date accuracy is up to milliseconds.'
+  - name: timestamps_reference_time
+    dtype: isodatetime
+    doc: 'Date and time corresponding to time zero of all timestamps.
+      COMMENT:
+        - The date is stored in UTC with local timezone offset as ISO 8601
+          formatted string: 2018-09-28T14:43:54.123+02:00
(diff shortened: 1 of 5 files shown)


tjd2002:
Per discussion with @oruebel and @ajtritt I have added the NWB 2.0 label.

Since this PR includes schema changes, what is the best way to ensure that nwb-schema and matnwb get these changes? 
- [ ] Create issue in nwb-schema
- [ ] Create issue in matnwb

tjd2002:
My use of `ObjectMapper.constructor_args` at https://github.com/NeurodataWithoutBorders/pynwb/blob/cd3d85bbd85c428d029bedc213135b3ffaa56a19/src/pynwb/io/file.py#L57-L61 to get the python datetime to roundtrip is completely copy-pasta (thanks @ukos-git!). It seems inelegant that we should have to repeat that conversion code for each datetime object. I would welcome feedback on that

ajtritt (dismissed):

bendichter (approved):

PR #717, opened 2018-11-08
bug/table foreign ref

ajtritt:
## Motivation

fix #713 

## Checklist

- [X] Have you checked our [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [x] Have you ensured the PR description clearly describes problem and the solution?
- [x] Is your contribution compliant with our coding style ? This can be checked running `flake8` from the source directory.
- [x] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/NeurodataWithoutBorders/pynwb/pulls) for the same change?
- [x] Have you included the relevant issue number using `#XXX` notation where `XXX` is the issue number ?

PR #719, opened 2018-11-08
Fix type of electrode_group in Units table

tjd2002:
## Motivation

NB this is a PR against another bugfix branch (#717), because it is actively working on related code.

The `electrode_group` field of the Units table had the type `array_type`, instead of `ElectrodeGroup`. ~Since ElectrodeGroup is a DynamicTableRegion~[EDIT: no it's not, I was thinking of ElectrodeTableRegion], we only need one of them to refer to multiple electrode channels.

However... do we need both? 

## Checklist

### DIFF
--- a/src/pynwb/misc.py
+++ b/src/pynwb/misc.py
@@ -6,6 +6,7 @@ from .form.utils import docval, getargs, popargs, call_docval_func
 from . import register_class, CORE_NAMESPACE
 from .base import TimeSeries, _default_conversion, _default_resolution
 from .core import NWBContainer, ElementIdentifiers, DynamicTable
+from .ecephys import ElectrodeGroup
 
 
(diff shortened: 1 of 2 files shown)


tjd2002:
Closing temporarily, so I can think about whether we even need electrode_group...

bendichter:
@tjd2002 This is an improvement, however each electrode in the electrodes table only has a single group, so if you use `ElectrodeGroup` objects, you won't be able to have different subsets with overlapping electrode memberships. If you want that, you'll have to use DynamicTable` reference objects that reference the `electrodes` table.

tjd2002:
Every electrode is already associated with an electrodeGroup (https://pynwb.readthedocs.io/en/latest/pynwb.file.html#pynwb.file.NWBFile.add_electrode), so I think this field is actually redundant.

PR #723, opened 2018-11-12
Enh/device description

ajtritt:
## Motivation

After #695, Device became a group with a help string. This PR adds a required description field. A Device ObjectMapper was also included for setting a default value for those files generated prior to this change.

This change was inspired by discussions with @ageorgou

## Checklist


oruebel (dismissed):

oruebel:
Looks like some of tests need to be updated, but other than that it looks good to me.

bendichter:
I prefer these types of additions to be optional. This change will require changes to dozen different conversion scripts and does not offer me much utility. I don't think these types of descriptive fields need to be required

ajtritt:
Then we should just remove Device. A group with a name and a constant help string doesn’t do much.

bendichter (dismissed):
making Device optional obviates the Device io file

bendichter:
The addition of `description` to the tests is now superfluous, but I don't mind leaving them

src/pynwb/data/nwb.file.yaml
```diff
@@ -235,6 +235,10 @@ groups:
           dtype: text
           doc: Value is 'A recording device e.g. amplifier'.
           value: A recording device e.g. amplifier
+        datasets:
```
bendichter, line 240: Why is this a dataset and not an attribute? I would not expect this to be changed once it is written, so shouldn't it be an attribute (https://github.com/NeurodataWithoutBorders/nwb-schema/issues/48)? Making this required would be a schema-breaking change, so I would like to make this optional.

bendichter (changes requested):

PR #738, opened 2018-11-18
remove docval ndim

bendichter:
## Motivation

remove remaining instances of docval 'ndim', which is deprecated in favor of 'shape'

## Checklist


t-b:
Could we teach the docval logic to complain about unknown settings? This would have caught the breakage here earlier.

bendichter:
@t-b Yeah good idea

oruebel (approved):

PR #742, opened 2018-11-20
NWBHDF5IO/HDF5IO: Use none as default opening mode

t-b:
The default opening mode for HDF5 files is, in accordance with h5py, "a"
which means "Read/write if exists, create otherwise (default)".

This is for numerous reasons a suboptimal choice:
- On windows this locks files which are maybe supposed to be opened read
  only. Thus other programs can not read these files.
- It opens the potential for data corruption if a file is meant to be
  read only and h5py and its C core crashes.
- It results in rather uninspiring error messages when
  trying to open a non-existing file for reading using

  with NWBHDF5IO("I_DONT_EXIST.nwb") as io:
          io.read()

  -> 5 pages of backtraces following

So let's use no mode as default and force the user to choose if she likes
to read or write.

### DIFF
--- a/src/pynwb/__init__.py
+++ b/src/pynwb/__init__.py
@@ -190,7 +190,7 @@ class NWBHDF5IO(HDF5IO):
 
     @docval({'name': 'path', 'type': str, 'doc': 'the path to the HDF5 file'},
             {'name': 'mode', 'type': str,
-             'doc': 'the mode to open the HDF5 file with, one of ("w", "r", "r+", "a", "w-")', 'default': 'a'},
+             'doc': 'the mode to open the HDF5 file with, one of ("w", "r", "r+", "a", "w-")'},
             {'name': 'load_namespaces', 'type': bool,
              'doc': 'whether or not to load cached namespaces from given path', 'default': False},
             {'name': 'manager', 'type': BuildManager, 'doc': 'the BuildManager to use for I/O', 'default': None},
@@ -226,7 +226,7 @@ class NWBHDF5IO(HDF5IO):
                 manager = get_manager(extensions=extensions)
             elif manager is None:
                 manager = get_manager()
-        super(NWBHDF5IO, self).__init__(path, manager, mode=mode, file=file_obj)
+        super(NWBHDF5IO, self).__init__(path, manager=manager, mode=mode, file=file_obj)
 
 
 from . import io as __io  # noqa: F401,E402
--- a/docs/gallery/general/advanced_hdf5_io.py
+++ b/docs/gallery/general/advanced_hdf5_io.py
@@ -174,7 +174,7 @@ io.close()
 #
 # Again, nothing has changed for read. All of the above advanced I/O features are handled transparently.
 
-io = NWBHDF5IO('advanced_io_example.nwb')
+io = NWBHDF5IO('advanced_io_example.nwb', 'r')
 nwbfile = io.read()
 
 ####################
--- a/docs/gallery/general/file.py
+++ b/docs/gallery/general/file.py
@@ -236,7 +236,7 @@ with NWBHDF5IO('basic_example.nwb', 'w') as io:
 # and use the :py:func:`~pynwb.form.backends.io.FORMIO.read` method to retrieve an
 # :py:class:`~pynwb.file.NWBFile` object.
 
-io = NWBHDF5IO('basic_example.nwb')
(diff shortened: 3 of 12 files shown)


src/pynwb/__init__.py
```diff
@@ -190,7 +190,7 @@ class NWBHDF5IO(HDF5IO):
 
     @docval({'name': 'path', 'type': str, 'doc': 'the path to the HDF5 file'},
             {'name': 'mode', 'type': str,
-             'doc': 'the mode to open the HDF5 file with, one of ("w", "r", "r+", "a", "w-")', 'default': 'a'},
+             'doc': 'the mode to open the HDF5 file with, one of ("w", "r", "r+", "a", "w-")', 'default': ''},
```
bendichter: wouldn't it be better to have no default value?

t-b: You mean no like in `None`?

bendichter:
```suggestion
             'doc': 'the mode to open the HDF5 file with, one of ("w", "r", "r+", "a", "w-")'},
```

bendichter: a `'default'` term is not required. This would make it a non-optional argument, which I believe is what you want

t-b: @bendichter Thanks for the clarifcation. Yes I want that!

t-b: If I leave out the `default` key I'll get something like

```
    @docval({'name': 'obj', 'type': (Container, Builder), 'doc': 'the object to get the ObjectMapper for'},
            returns='the ObjectMapper to use for mapping the given object', rtype='ObjectMapper')
    def get_map(self, **kwargs):
        """ Return the ObjectMapper object that should be used for the given container """
        obj = getargs('obj', kwargs)
        # get the container class, and namespace/data_type
        if isinstance(obj, Container):
            container_cls = obj.__class__
            namespace, data_type = self.get_container_ns_dt(obj)
            if namespace is None:
>               raise ValueError("class %s is not mapped to a data_type" % container_cls)
E               ValueError: class <class 'pynwb.file.NWBFile'> is not mapped to a data_type
```

I'll stick with None for now then.

bendichter: I'm confused with this error. Our convention for required args is to leave out the default key, so I'd like to make that work if possible. Let me check out your branch and see if I can make it work.

bendichter: OK I can recreate this error but I still haven't figured out what's going on here. Let me give this a bit more time, and as a plan B we can make the default arg `None`

oruebel: I have not looked at how docval handles this in detail, so my memory might be wrong, but here is what I think is going on.

I believe @bendichter is correct. Setting ``default=None`` I believe implies that None is a keyword argument with a default value of None. Since you didn't change how mode is being handled in the constructor, I think this means in this case that that self.mode in HDF5IO is being set to None and I think in turn h5py will end up choosing a mode, so effectively you still have an implicit default (i.e., whatever h5py does when you do h5py.File(..., mode=None)). I think this is also why the galleries are still passing tests, because there are quite a few instances in our tutorial galleries that do not set the mode, and all of those should fail tests if mode is a required argument. Note, when supplying no default arg in docval, then that means mode becomes a positional argument so one reason for the error you see could be that you inadvertently changed the order in which arguments are expected (which might explain the error you see).

t-b: @oruebel Thanks for the explanation. The 'keyword argument' vs positional might explain the error I'm seeing. I actually prefer having it a keyword argument and are now catching none in HDF5IO. As predicted some gallery tests broke.

oruebel: @t-b if I remember correctly, docval allows you to still set positional arguments via keyword assignemnt, i.e,. you should still be able to do ``mode='r'`` even when not specifying a default. In general, I think if ``mode`` is a required argument then we should do this in docval by leaving out the default key. The reason is, that in this way it shows up correctly as a required argument in the documentation and help, whereas if you set ``default=None`` then it looks like an optional argument in the documentation and then raise an error if a users actually uses the default.

t-b: @oruebel Makes sense. Done.

oruebel (approved):
Looks good to me. Thanks @t-b

bendichter:
Yeah, thanks @t-b! I'll merge as soon as tests pass

PR #750, opened 2018-12-04
add SpectralAnalysis

### LINKED ISSUE #227: storing phase and amplitude
Can we talk about how we store spectral band (e.g. theta phase and amplitude). There is an object called `FilteredEphys` in the `ecephys` module, which would seem to store this information:

> Ephys data from one or more channels that has been subjected to filtering. Examples of filtered data include Theta and Gamma (LFP has its own interface).

however on closer inspection it appears that this object is meant to store literally the filtered signal (the signal after a bandpass filter), not the phase or amplitude, because the object holds electrical_series objects, which are in units Volts, and there is no way of indicating if the signal is amplitude or phase. 

While it would be possible to get the phase and amplitude info from this filtered signal, this approach makes some assumptions about how the phase and amplitude are computed that are not true in all cases. Also, in my experience it is common for labs to store phase and amplitude directly as intermediate data.

This could of course be handled through extensions, but it seems like something that might be of general use to standardize across labs. I think this can be done in a non-schema-breaking way. The challenge is that these bands are computed differently lab-to-lab. For instance, we would need to account for the fact that labs tend to have different definitions of where the limits of the canonical band are. Actually, `ecephys.FeatureExtraction` is pretty close to what we need. It already has data of shape `[time x electrode x feature]`. Here is my proposal for how to extend it.

adding to ecephys:

```
LFPSpectralAnalysis(FeatureExtraction):
    band_limits: ([float float],...)
    metric: str
         recommended: phase, amplitude, power
    timeseries: link to raw signal
    band_name (optional): [str, str,...]
        recommended: alpha, beta, gamma, delta, theta, high gamma
```
The idea is for this structure to accommodate two different use-cases. The first is where we are looking at canonical bands (theta, gamma, etc.). In this case, `band_name` should be entered. The second is when a full spectrogram (e.g. the results of an STFT) is provided. In this case, the `band_names` can be omitted. The way I have proposed here, whether this is a phase or amplitude is indicated by the `metric` field. Since we don't yet have enumerated types, this must be a free field, which is not ideal. Alternatively we could have `LFPSpectralPhase` and `LFPSpectralAmplitude` inherit from `LFPSpectralAnalysis`.

You would use different objects for different types of spectral analysis (e.g. wavelet vs. Hilbert). Perhaps we could sub-class this with a few common ones, like STFT, and allow for explicit storage of variables e.g. window_len, window_function, window_overlap, etc. but this will be difficult to standardize across labs and might be better done with an extension.

Since this requires `electrodes` and is in the `ecephys` module, a separate structure would need to be made for storing the spectrogram of behavioral data, e.g. speech, but it would look very similar to this one.

@ajtritt I know you were thinking about this. Were you thinking of doing this another way?

bendichter:
## Motivation

fix https://github.com/NeurodataWithoutBorders/nwb-schema/issues/227

## How to test the behavior?
```python
timeseries = TimeSeries(name='dummy timeseries', description='desc',
                                data=np.ones((3, 3)),
                                timestamps=np.ones((3, )))
spec_anal = SpectralAnalysis(name='LFPSpectralAnalysis',
                                     description='my description',
                                     data=np.ones((3, 3, 3)),
                                     timestamps=np.ones((3,)),
                                     band_name=['alpha', 'beta', 'gamma'],
                                     band_limits=np.ones((3, 2)),
                                     timeseries=timeseries,
                                     metric='amplitude')
from pynwb import NWBHDF5IO, NWBFile
from datetime import datetime
nwbfile = NWBFile('a', 'b', datetime.now())
nwbfile.add_acquisition(spec_anal)
with NWBHDF5IO('test_write.nwb', 'w') as io:
    io.write(nwbfile)
```

## Checklist

### DIFF
--- a/src/pynwb/data/nwb.misc.yaml
+++ b/src/pynwb/data/nwb.misc.yaml
@@ -115,15 +115,6 @@ groups:
   neurodata_type_inc: TimeSeries
   doc: Holds spectral analysis of a timeseries. For instance of LFP or a speech signal
   datasets:
-  - name: band_limits
-    dtype: float
-    doc: Low and high limit of each band. If it is a Gaussian filter, use 2 SD on either side of the center
-    shape:
-    - null
-    - 2
-    dims:
-    - num_bands
-    - low, high
   - name: data
     dtype: float
     doc: The data goes here
--- a/src/pynwb/data/nwb.misc.yaml
+++ b/src/pynwb/data/nwb.misc.yaml
@@ -138,20 +138,42 @@ groups:
   - name: metric
     dtype: text
     doc: 'recommended: phase, amplitude, power'
-  - name: band_name
-    doc: 'recommended: alpha, beta, gamma, delta, theta, high gamma'
-    dtype: text
-    shape:
-    - null
-    dims:
-    - num_bands
-    quantity: '?'
   links:
   - name: source_timeseries
     doc: HDF5 link to TimesSeries that this data was calculated from. Metadata
       about electrodes and their position can be read from that ElectricalSeries so
       it's not necessary to mandate that information be stored here
     target_type: TimeSeries
+  groups:
+  - neurodata_type_inc: DynamicTable
+    name: bands
+    doc: A table for describing the bands that this series was generated from. There
+         should be one row in this table for each band
+    datasets:
+    - neurodata_type_inc: VectorData
+      name: band_name
+      dtype: text
+      doc: the name of the band e.g. theta
+      attributes:
+      - name: description
+        dtype: text
+        doc: value is 'the name of the band e.g. theta'
+        value: the name of the band e.g. theta
+    - neurodata_type_inc: VectorData
+      name: band_limits
+      dtype: float
+      shape:
+      - null
+      - 2
+      dims:
+      - num_bands
+      - low, high
+      doc: Low and high limit of each band. If it is a Gaussian filter, use 2 SD on either side of the center
+      attributes:
+      - name: description
+        dtype: text
+        doc: value is 'the mean of the sliding Gaussian'
+        value: the mean of the sliding Gaussian
 - neurodata_type_def: Units
   neurodata_type_inc: DynamicTable
   doc: Data about spiking units. Event times of observed units (e.g. cell, synapse, etc.)
--- a/src/pynwb/data/nwb.misc.yaml
+++ b/src/pynwb/data/nwb.misc.yaml
@@ -111,6 +111,47 @@ groups:
     - num_times
     shape:
     - null
+- neurodata_type_def: SpectralAnalysis
+  neurodata_type_inc: TimeSeries
+  doc: Holds spectral analysis of a timeseries. For instance of LFP or a speech signal
+  datasets:
+  - name: band_limits
+    dtype: float
+    doc: Low and high limit of each band. If it is a Gaussian filter, use 2 SD on either side of the center
+    shape:
+    - null
+    - 2
+    dims:
+    - num_bands
+    - low, high
+  - name: data
+    dtype: float
+    doc: The data goes here
+    shape:
+    - null
+    - null
+    - null
+    dims:
+    - num_times
+    - num_channels
+    - num_bands
+  - name: metric
+    dtype: text
+    doc: 'recommended: phase, amplitude, power'
+  - name: band_name
+    doc: 'recommended: alpha, beta, gamma, delta, theta, high gamma'
+    dtype: text
+    shape:
+    - null
+    dims:
+    - num_bands
+    quantity: '?'
+  links:
+  - name: source_timeseries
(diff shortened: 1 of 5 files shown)


ajtritt (commented):
The original implementation I suggested included a DynamicTable for storing information about bands. There were some pre-specified metadata fields (i.e. band_limits, band_name), but, being that it was a DynamicTable, users could specify other descriptors of a band. It also had the benefit of being able to use any functionality that came along with DynamicTable for manipulating this metadata. 

Is there any downside to using a DynamicTable to include band_limits and band_name?

bendichter:
@ajtritt yeah I can change it to use a DynamicTable for bands

bendichter:
@ajtritt It's added complexity, and I don't quite understand the motivation. Can you give me an example of what columns might be added to the bands `DynamicTable`?

bendichter:
@ajtritt how does this look?

src/pynwb/data/nwb.misc.yaml
```diff
@@ -111,6 +111,69 @@ groups:
     - num_times
     shape:
     - null
+- neurodata_type_def: SpectralAnalysis
+  neurodata_type_inc: TimeSeries
+  doc: Holds spectral analysis of a timeseries. For instance of LFP or a speech signal
+  datasets:
+  - name: band_limits
```
ajtritt: did you mean to leave this?

ajtritt (dismissed):

src/pynwb/data/nwb.misc.yaml
bendichter: ah, good catch

oruebel:
This question is simply because you mention LFP in this type, i.e., I have not looked at the details and so this may or may not be important. We have the LFP MultiContainerInterface 

https://github.com/NeurodataWithoutBorders/pynwb/blob/779fa2543788469d50fe36932aff6de79e5f11b5/src/pynwb/ecephys.py#L276-L277

and corresponding neurodata_type in the schem https://nwb-schema.readthedocs.io/en/latest/format.html#lfp . Does this mean that type should be updated to accept ``SpectralAnalysis`` in addition to (or instead of) ``ElectricalSeries``?

bendichter:
@oruebel that's an interesting idea. The way it's written now, you would infer that a spectral analysis object is a an LFP if the source timeseries is an electrical series that is within an LFP interface. I think I like your idea to include it in the LFP interface. This would require a second LFP interface, since the first on is in /acquisition, and SpectralAnalysis should be in a processing module. So would the idea be to have two LFP interfaces, one in /acquisition and one in a processing module?

oruebel:
@bendichter I may be confused, but I thought https://nwb-schema.readthedocs.io/en/latest/format.html#lfp was just a NWBDataInterface that could live in either acquistion or processing?

bendichter:
@oruebel Yes, it can. What I mean is that the ElectricalSeries raw LFP should live in /acquisition, and the SpectralAnalysis should live in /processing, so this would require two LFP objects, one in acquisition and one in processing. I'm just confirming that this is what you had in mind.

oruebel:
Ok, I see. The distinction between raw LFP and processed LFP was not clear to me. For now, I think this means that this PR can go through and we worry about the raw/processed LFP issue in a separate thread.

bendichter:
@oruebel all I'm saying is this will require two instances of the LFP object

oruebel:
> @oruebel all I'm saying is this will require two instances of the LFP object

I think that's fine.

bendichter:
@oruebel ok I have added this feature and the following test:

```python
        lfp = LFP()
        timeseries = TimeSeries(name='dummy timeseries', description='desc',
                                data=np.ones((3, 3)),
                                timestamps=np.ones((3,)))
        spec_anal = SpectralAnalysis(name='LFPSpectralAnalysis',
                                     description='my description',
                                     data=np.ones((3, 3, 3)),
                                     timestamps=np.ones((3,)),
                                     band_name=['alpha', 'beta', 'gamma'],
                                     band_limits=np.ones((3, 2)),
                                     timeseries=timeseries,
                                     metric='amplitude')
        lfp.add_spectral_analysis(spec_anal)
        self.assertEqual(lfp.spectral_analysis.get('LFPSpectralAnalysis'), spec_anal)
```

The downside is that `LFP` would no longer be subscriptable, i.e. `lfp['LFPSpectralAnalysis']` and `lfp['my_electrial_series']` would no longer work, and you would have to instead use `lfp.spectral_analysis.get('LFPSpectralAnalysis')` and `lfp.electrical_series.get('my_electrical_series')`

Are we all on-board with broadening `LFP` in this way to accept `SpectralAnalysis` in addition to the `ElectricalSeries` objects it currently holds? @ajtritt, would this cause issues with your query framework? @tjd2002, would this work for you?

oruebel:
>  added this feature and the following test:

I think this is fine. @ajtritt can you have a look.

ajtritt (approved):

PR #752, opened 2018-12-05
Add observation intervals (obs_intervals) as a column to misc.Units table

tjd2002:
## Motivation
When storing spike times for a clustered unit, it's also important to store the times during which the clustered cell was being observed. (For instance, a cell might not fire at all during a recording epoch, we still need to be able to represent this)

We propose to add an (optional) column to the units table called 'obs_intervals' that allows the user to store, for each unit, a list of the times that the unit was being observed, and a spike event *could* have happened. 

Note: we can't just store this list once for all units. For example, one unit may drift in and out of being clusterable independently of others. (or an artifact may prevent observation on some subset of channels)

We have had offline discussions about this feature with @bendichter @ajtritt and @oruebel, which led to the suggestion that we implement this as an optional column on the Units table. @ajtritt suggested using the VectorData facility, with a compound datatype.

This first commit tries to implement this, but currently fails when we try to actually store data, as we're not sure how to actually pass in the interval list to get it written successfully to the file.

cc @angevinemiller
## How to test the behavior?
```
```

## Checklist

-  [X] Is your contribution compliant with our coding style ? This can be checked running `flake8` from the source directory.
- [ ] Have you included the relevant issue number using `#XXX` notation where `XXX` is the issue number ? (will create an issue for discussion if needed)


bendichter:
Looks great, @tjd2002. Needs testing though.

oruebel:
> but currently fails when we try to actually store data

@tjd2002 just to clarify, are you asking for help on this PR?

tjd2002:
> @tjd2002 just to clarify, are you asking for help on this PR?

Yes, @ajtritt asked me to create a PR where he could help us. @angevinemiller is about to post an example.

@bendichter: Agreed it needs testing! We'll definitely provide that before asking for a merge.

angevineMiller:
Here is a minimal working example _without_ adding obs_intervals to the Units table:
```python

import numpy as np
import pynwb
from datetime import datetime
from dateutil import tz
nwbf = pynwb.NWBFile(session_description='Test without obs_intervals', 
                    identifier='Test1',
                    session_start_time=datetime(2006, 1, 1, 12, 0, 0, tzinfo=tz.gettz('US/Pacific')))

# Do not add any obs_intervals.
nwbf.add_unit(spike_times=[1,2,3,4,5])

with pynwb.NWBHDF5IO('output1.nwb', mode='w') as iow:
    iow.write(nwbf)
```

It also works if we add a 1D obs_intervals, but this is not what we want, since a single unit can have multiple observation intervals:
```python
nwbf.add_unit(spike_times=[1,2,3,4,5], obs_intervals=[0, 9])
```

However, it fails on write after adding obs_intervals as lists (or as numpy arrays). The only thing we change is this:
```python
# Add obs_intervals
nwbf.add_unit(spike_times=[1,2,3,4,5], obs_intervals=[[0, 4], [5, 9]])
```

Here is a snippet from the stack trace:
```python
...
...
...
~/Src/pynwb/src/pynwb/form/backends/hdf5/h5tools.py in write_dataset(self, **kwargs)
    676             # If the compound data type contains only regular data (i.e., no references) then we can write it as usual
    677             else:
--> 678                 dset = self.__list_fill__(parent, name, data, options)
...
...
...
OSError: Can't prepare for writing data (src and dest data spaces have different sizes)
```

angevineMiller:
Also, if we add multiple units:
```python
nwbf.add_unit(spike_times=[1,2,3,4,5], obs_intervals=[[0, 4], [5, 10]])
nwbf.add_unit(spike_times=[11,12,13,14,15], obs_intervals=[[10, 14], [15, 20]])
```
We get a slightly different, but seemingly very similar, error on write:
```python
TypeError: Can't broadcast (4, 2) -> (4,)
```

ajtritt:
Can you try adding the intervals as tuples or a list of tuples?

angevineMiller:
@ajtritt Thanks for the suggestion.  Adding the intervals as a list or tuple of tuples worked on write. I'm working on getting it to work on read now.

tjd2002:
I think we are ready for review on this now.

bendichter (approved):
Looks great! Thanks for the contribution! One final request would be to document this feature in the more descriptive tutorial sections of the documentation. I think it's particularly important because this is such a widespread need but not a common practice to store, so I think other labs would benefit a lot from a description of how to use it. Looking at the docs, the section about units in the ecephys section is pretty lacking, so this is a bigger task. The API docs look good, so I don't want to block a merge. Great job, thanks for the contribution!

bendichter:
Should anyone here present know of any reason that this PR should not be merged, speak now or forever hold your peace.

tjd2002:
We are going to talk to @ajtritt  in a couple hours, I can ask him then to take a quick look.

bendichter:
@tjd2002 ok perfect, thanks

ajtritt (approved):

PR #758, opened 2018-12-12
security update: urllib3==1.23

bendichter:
## Motivation

GitHub is warning about a security issue with urllib3<1.23 so I am updating the requirements


t-b:
Could you include some more info in the commit message about the bug we are trying to avoid here?

dorukozturk (approved):

Issue #669, opened 2018-10-11
Convert between pixel mask and image mask


ajtritt:
## Feature Request
Users are no longer required to provide both *pixel_mask* and *image_mask*--they can provide either or both.  To make it easier to work with files that use different mask types, supply PlaneSegmentation with a method that can convert from *pixel_mask* to *image_mask* and back. 

Motivated by @bendichter [here](https://github.com/NeurodataWithoutBorders/pynwb/pull/665#issuecomment-429113306)

## Checklist

- [X] Have you ensured the feature or change was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues/) ?
- [X] Have you included a brief and descriptive title?
- [X] Have you included a clear description of the problem you are trying to solve?
- [X] Have you included a minimal code snippet that reproduces the issue you are encountering?

mamelara:
Hello!

Looking to be a new contributor to pyNWB and wanted to work on this issue. I am a bit confused as to the difference between pixel_mask and img_mask. To me, the only difference is that one is basically a list of lists and the other is a list of tuples (this is from my very beginner understanding of the two types.) Could you please elaborate on what is expected from this conversion? Thanks!

bendichter:
Hey @mamelara, welcome to the team! pixel_mask and img_mask are two different ways of specifying regions of an image. A pixel mask is a list where each element of the list describes a single pixel: (x, y, value). The size of a pixel mask is [any x 3]. For an image mask, the mask is the same size as the original image, and only the values in the region are non-zero. We would like to be able to convert from one type to another. Does that help?

ageorgou:
I just thought I'd mention, since I don't see it here, that it may be useful to look at how [the old NWB  API](https://github.com/NeurodataWithoutBorders/api-python/blob/master/nwb/nwb_utils.py#L278) created a pixel mask from an image mask and vice versa. Some changes may be needed,  though, at least because the pixel mask now also contains the weights (which were a separate argument in the old API).

PR #766, opened 2018-12-18
Conversion between pixel_mask and image_mask

mamelara:
## Motivation

Addresses Issue #669 

## How to test the behavior?

Unit tests are included. To run the tests, use tox.
## Checklist

### DIFF
--- a/src/pynwb/ophys.py
+++ b/src/pynwb/ophys.py
@@ -304,6 +304,7 @@ class PlaneSegmentation(DynamicTable):
     def create_roi_table_region(self, **kwargs):
         return call_docval_func(self.create_region, kwargs)
 
+
 @register_class('ImageSegmentation', CORE_NAMESPACE)
 class ImageSegmentation(MultiContainerInterface):
     """
--- a/src/pynwb/ophys.py
+++ b/src/pynwb/ophys.py
@@ -279,6 +279,9 @@ class PlaneSegmentation(DynamicTable):
 
     @staticmethod
     def pixel_to_image(pixel_mask):
+        """
+        Converts a 2D pixel_mask of a ROI into an image_mask.
+        """
         image_matrix = np.zeros(np.shape(pixel_mask))
         npmask = np.asarray(pixel_mask)
         x_coords = npmask[:, 0].astype(np.int)
@@ -289,6 +292,9 @@ class PlaneSegmentation(DynamicTable):
 
     @staticmethod
     def image_to_pixel(image_mask):
+        """
+        Converts an image_mask of a ROI into a pixel_mask
+        """
         pixel_mask = []
         it = np.nditer(image_mask, flags=['multi_index'])
         while not it.finished:
--- a/src/pynwb/ophys.py
+++ b/src/pynwb/ophys.py
@@ -277,16 +277,18 @@ class PlaneSegmentation(DynamicTable):
             rkwargs['voxel_mask'] = voxel_mask
         return super(PlaneSegmentation, self).add_row(**rkwargs)
 
-    def pixel_to_image(self, pixel_mask):
+    @staticmethod
+    def pixel_to_image(pixel_mask):
         image_matrix = np.zeros(np.shape(pixel_mask))
         npmask = np.asarray(pixel_mask)
-        x_coords = tuple(map(int, npmask[:, 0]))
-        y_coords = tuple(map(int, npmask[:, 1]))
+        x_coords = npmask[:, 0].astype(np.int)
+        y_coords = npmask[:, 1].astype(np.int)
         weights = npmask[:, -1]
         image_matrix[y_coords, x_coords] = weights
         return image_matrix
 
-    def image_to_pixel(self, image_mask):
+    @staticmethod
+    def image_to_pixel(image_mask):
         pixel_mask = []
         it = np.nditer(image_mask, flags=['multi_index'])
         while not it.finished:
--- a/src/pynwb/ophys.py
+++ b/src/pynwb/ophys.py
@@ -277,25 +277,25 @@ class PlaneSegmentation(DynamicTable):
             rkwargs['voxel_mask'] = voxel_mask
         return super(PlaneSegmentation, self).add_row(**rkwargs)
 
-    def pixel_to_image(self, pixel_mask, width, height):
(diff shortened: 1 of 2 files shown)


mamelara:
This is still WIP, but just wanted to see if I'm going in the right direction with this. I also need to take into account that an image_mask can have multiple image matrices in it.

src/pynwb/ophys.py
```diff
@@ -271,6 +271,27 @@ def add_roi(self, **kwargs):
             rkwargs['voxel_mask'] = voxel_mask
         return super(PlaneSegmentation, self).add_row(**rkwargs)
 
+    def pixel_to_image(self, pixel_mask, width, height):
+        image_matrix = np.zeros((width, height))
+        for mask in pixel_mask:
```
oruebel: I think something like this might be a bit faster than setting all the elements one-by-one in the for loop:

```
npmask = np.asarray(pixel_mask)
image_matrix[npmask[:,0], npmask[:,1]] =  npmask[:,-1]
return image_matrix
```

src/pynwb/ophys.py
```diff
@@ -271,6 +271,27 @@ def add_roi(self, **kwargs):
...
+            col = mask[0]
+            row = mask[1]
+            weight = mask[-1]
+            if weight > 0:
+                image_matrix[row][col] = weight
+        return image_matrix
+
+    def image_to_pixel(self, image_mask):
+        """Convert a image_mask to a pixel_mask that contains all the elements of the form (x, y, weight)"""
+        width, height = np.shape(image_mask)
```
oruebel: Here ``numpy nditer`` might help simplify things:

```
pixel_mask = []
it = np.nditer(image_mask, flags=['multi_index'])
while not it.finished:
    pixel_mask.append( [it.multi_index[0], it.multi_index[1], it[0][()]] )
    it.iternext()
return pixel_mask



```

mamelara:
Did some of the suggestions given. Only failing test is creating pixel_masks out of multiple images as given in this example in `test_init`:

```python
img_mask = [[[1.0 for x in range(w)] for y in range(h)], [[2.0 for x in range(w)] for y in range(h)]]
pix_mask = [[1, 2, 1.0], [3, 4, 1.0], [5, 6, 1.0],
                    [7, 8, 2.0], [9, 10, 2.0]]
``` 

For clarification, does the pixel_mask just treat the two w x h arrays as just one giant coordinate? Meaning that, pix_mask [7, 8, 2.0] refers to the x-y coordinates of the second image?

src/pynwb/ophys.py
```diff
@@ -272,6 +272,27 @@ def add_roi(self, **kwargs):
             rkwargs['voxel_mask'] = voxel_mask
         return super(PlaneSegmentation, self).add_row(**rkwargs)
 
+    def pixel_to_image(self, pixel_mask):
```
bendichter:
```suggestion
        @staticmethod
    def pixel_to_image(pixel_mask):
        image_matrix = np.zeros(np.shape(pixel_mask))
        npmask = np.asarray(pixel_mask)
        x_coords = npmask[:, 0].astype(np.int)
        y_coords = npmask[:, 1].astype(np.int)
        weights = npmask[:, -1]
        image_matrix[y_coords, x_coords] = weights
        return image_matrix

    @staticmethod
    def image_to_pixel(image_mask):
```

src/pynwb/ophys.py
```diff
@@ -272,6 +272,27 @@ def add_roi(self, **kwargs):
             rkwargs['voxel_mask'] = voxel_mask
         return super(PlaneSegmentation, self).add_row(**rkwargs)
```
bendichter, line 279:
```suggestion

    @staticmethod
```

src/pynwb/ophys.py
bendichter:
```suggestion
    def pixel_to_image(pixel_mask):
```

src/pynwb/ophys.py
```diff
@@ -272,6 +272,27 @@ def add_roi(self, **kwargs):
...
 
+    def pixel_to_image(self, pixel_mask):
+        image_matrix = np.zeros(np.shape(pixel_mask))
+        npmask = np.asarray(pixel_mask)
+        x_coords = tuple(map(int, npmask[:, 0]))
+        y_coords = tuple(map(int, npmask[:, 1]))
+        weights = npmask[:, -1]
+        image_matrix[y_coords, x_coords] = weights
+        return image_matrix
+
```
bendichter, line 292:
```suggestion

    @staticmethod
```

src/pynwb/ophys.py
```diff
@@ -272,6 +272,27 @@ def add_roi(self, **kwargs):
...
+    def image_to_pixel(self, image_mask):
```
bendichter:
```suggestion
    def image_to_pixel(image_mask):
```

src/pynwb/ophys.py
```diff
@@ -272,6 +272,27 @@ def add_roi(self, **kwargs):
             rkwargs['voxel_mask'] = voxel_mask
         return super(PlaneSegmentation, self).add_row(**rkwargs)
 
+    def pixel_to_image(self, pixel_mask):
+        image_matrix = np.zeros(np.shape(pixel_mask))
+        npmask = np.asarray(pixel_mask)
+        x_coords = tuple(map(int, npmask[:, 0]))
```
bendichter:
```suggestion
        x_coords = npmask[:, 0].astype(np.int)
```

src/pynwb/ophys.py
```diff
@@ -272,6 +272,27 @@ def add_roi(self, **kwargs):
...
+        y_coords = tuple(map(int, npmask[:, 1]))
```
bendichter:
```suggestion
        y_coords = npmask[:, 1].astype(np.int)
```

oruebel:
@mamelara what is the status of this PR? Looks like there were a few minor changes requested, could you add those so we can close and merge?

mamelara:
Rebased to current dev and added suggested changes

oruebel (dismissed):
It would be good to add a brief docstring to the pixel_to_image and image_to_pixel methods to describe what they do, but other than that I think this looks good.

oruebel (approved):

PR #778, opened 2019-01-09
add py37

bendichter:
## Motivation

Test python 37 support

## How to test the behavior?
```
Show here how to reproduce the new behavior (can be a bug fix or a new feature)
```

## Checklist

- [ ] Have you checked our [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [ ] Have you ensured the PR description clearly describes problem and the solution?
- [ ] Is your contribution compliant with our coding style ? This can be checked running `flake8` from the source directory.
- [ ] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/NeurodataWithoutBorders/pynwb/pulls) for the same change?
- [ ] Have you included the relevant issue number using `#XXX` notation where `XXX` is the issue number ?

PR #779, opened 2019-01-09
Python 3.7

dorukozturk:
## Motivation

Fixes #778 
Add python 3.7 to our test suite

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -284,17 +284,23 @@ workflows:
           <<: *no_filters
       - python36:
           <<: *no_filters
+      - python37:
+          <<: *no_filters
       - miniconda27:
           <<: *no_filters
       - miniconda35:
           <<: *no_filters
       - miniconda36:
           <<: *no_filters
+      - miniconda37:
+          <<: *no_filters
       - gallery27:
           <<: *no_filters
       - gallery35:
           <<: *no_filters
       - gallery36:
+          <<: *no_filters
+      - gallery37:
           <<: *no_filters
       - deploy-dev:
           requires:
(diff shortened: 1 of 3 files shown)


bendichter:
I don't see the tests in the test status box. How do you make them appear there?

dorukozturk:
@bendichter I think my force push confused circleci. I will restart the jobs and we should hopefully see it here.

t-b:
@dorukozturk I think the gcc errors are due to the fact that h5py 2.7 is not available in compiled form for python 3.7. At least this is one of the obstacles here on windows.

dorukozturk:
@t-b oh ok. Now it makes sense. Looking here they don't support 3.7 yet. https://pypi.org/project/h5py/2.7.1/. It is also the case with the latest h5py which is 2.9.0. 
What should we do?

t-b:
@dorukozturk I would disable the failing python 3.7 tests and document in the commit message why. Then we can later on enable them.

dorukozturk:
Updating h5py and ruamel.yaml solved the issues. I think this is ready for review.

oruebel (approved):

bendichter:
Wow, looks like you went above and beyond in getting this through. Thanks, @dorukozturk! So let me see if I understand correctly. pynwb works for python 3.7, but requires the user to manually build the hdf5 and h5py dependencies? Does this mean that it can't be conda installed entirely from scratch in python 3.7 yet?

dorukozturk:
Sure, I am happy to help. Right now pynwb should work with python3.7 without user doing anything (they can pip install and conda install pynwb). The new workflow is not using the cache for some reason and I am trying to figure that out. I have to do a follow up PR that solves 2 minor issues.

dorukozturk:
Ok everything is good now :D. I fixed the cache problem.

PR #782, opened 2019-01-11
convert datasets and attributes to minimum specified precision

### LINKED ISSUE #732: Bits per pixel not respected
When creating an `ImageSeries`, we always seem to end up with 8-bit data, even if that's not enough to store all values, which then corrupts our data. This happens even when explicitly setting the `bits_per_pixel` to a higher value.
Is that option supposed to be respected for I/O, or is it just metadata? Either way, is there a way to use more than 8 bits per pixel?

Steps to reproduce:

```python
from datetime import datetime

import numpy as np
from pynwb.image import ImageSeries

nwbfile = NWBFile('image series test', 'NWB123', datetime.now())

# Add a single-frame image that needs more than 8 bits
img = np.array([[255, 128], [0, 10]], dtype=np.int16)
print(img.dtype)
img_series = ImageSeries('image_series', data=[img],
                         timestamps=[0.0], bits_per_pixel=16)
nwbfile.add_acquisition(img_series)

# Write the file
with NWBHDF5IO('test_image.nwb', 'w') as io:
    io.write(nwbfile)

# Check again: the image is now 8bit
with NWBHDF5IO('test_image.nwb', 'r') as io:
    nwbfile = io.read(nwbfile)
    img = nwbfile.get_acquisition('image_series')
    print(img.data.dtype)
```

for which the output is:
```
<snipped timezone warning>
int16
int8
```

This is with the current `dev` (033d2462) and, in case it makes a difference:
```
numpy==1.13.0
h5py==2.8.0
```

Thanks in advance!

ajtritt:
## Motivation

fix #732 

## Checklist


oruebel (dismissed):

oruebel:
Looks like the validator is failing for 'isodatetime' in the gallery tests

ageorgou:
The latest commit in this branch seems to behave wrong when a dataset with a signed spec is given unsigned values. For example:

```python
from datetime import datetime

import numpy as np
from pynwb.image import ImageSeries

nwbfile = NWBFile('image series test', 'NWB123', datetime.now())

# Add an image
img = np.array([[255, 128], [0, 10]], dtype=np.uint16)
print(img.dtype)
img_series = ImageSeries('image_series', data=[img],
                         timestamps=[0.0], bits_per_pixel=16)
nwbfile.add_acquisition(img_series)

# Write the file
with NWBHDF5IO('test_image.nwb', 'w') as io:
    io.write(nwbfile)

# Check again: the image is now 8bit
with NWBHDF5IO('test_image.nwb', 'r') as io:
    nwbfile = io.read(nwbfile)
    img = nwbfile.get_acquisition('image_series')
    print(img.data.dtype)
```

This will write the field as `int8` (as in the spec) rather than `uint16` (as given), and the values > 127 overflow. This didn't occur in 9cf891d, which behaves correctly (at least for the above example), so it does look like it's the changes in 1c5b533.

oruebel:
@ln-vidrio @nclack note that this adds "numeric" as an allowable dtype in the schema. Once this is merged can you please pull and update the schema in MatNWB.

oruebel (approved):

ajtritt:
@ageorgou 

The issue you raised was happening because you were trying to pass in an unsigned int where a signed int was specified. To honor the specification, the uint was being cast to a signed int. I removed this behavior in this PR.

The issue you raised re-exposed the larger problem of having specific numeric types specified where users may want to store multiple types e.g ImageSeries.data was specified as an int8, but you want to store a uint. I addressed this by restoring the `numeric` option for dtype. I set the dtype on ImageSeries.data (among others) to `numeric`, so you should be able to pass in unsigned or signed integers, as well as floating points

Fore future reference, the only deviation you can make from the spec is passing in a higher precision value than what is specified. For example, if a field is specified as int16, you can pass an int16, int32, or int64. Passing a uint or float (of any precision) will cause an error when you call `NWBHDF5IO.write`

PR #784, opened 2019-01-14
Fix ogen, add RT tests

### LINKED ISSUE #783: OptogeneticStimulusSite RT error
[deleted]

bendichter:
## Motivation
The ogen module had a few inconsistencies. Some of the types are wrong and/or mismatched between the schema and pynwb, and there were previously no round trip tests. I created one, but it's not passing and I don't know why.

The changes I am suggesting including changing `excitation_wavelength` to be a `float` instead of a `str`, which would make it consistent with the ophys module, but is a breaking change. I think this should be allowed, since the original type was clearly a mistake, but I understand that breaking changes are a big ask at this point so I'd be willing to compromise on that point.

fix #783 

## Checklist

### DIFF
--- a/src/pynwb/data/nwb.ogen.yaml
+++ b/src/pynwb/data/nwb.ogen.yaml
@@ -31,7 +31,7 @@ groups:
     target_type: OptogeneticStimulusSite
 - neurodata_type_def: OptogeneticStimulusSite
   neurodata_type_inc: NWBContainer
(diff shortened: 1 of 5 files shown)


ajtritt (approved):
looks good, nice work!

oruebel (approved):
Thanks for the fix!

PR #809, opened 2019-01-31
[WIP] make HDMF a submodule of pynwb

ajtritt:
## Motivation

The `pynwb.form` subpackage has been renamed and moved to the it's own [repository](https://github.com/hdmf-dev/hdmf).

The main reasons for this was to make it easier to reuse the code infrastructure upon which PyNWB is built. We are also renaming the package from FORM (Flexible Object Relational Modeling) to HDMF (Hierarchical Data Modeling Framework). The original name is a misnomer, since there are no explicit capabilities for relational data modeling. *HDMF* was chosen since the primary functionality is for describing, reading, and writing hierarchical object data.


bendichter:
I expected you to make hdmf pip installable and list it in setup.py. Is there a reason you chose to make it a submodule instead?

requirements.txt
```diff
@@ -9,3 +9,4 @@ ruamel.yaml==0.15.85
 six==1.11.0
 urllib3==1.23
 pandas==0.23.4
+hdmf
```
jcfr: Waiting the package is available pypi, I suggest to remove the submodule and simply specify the following:

```
[...]
urllib3==1.23
pandas==0.23.4
git+https://github.com/hdmf-dev/hdmf@dev#egg=hdmf
```

If a specific commit need to be checkout, the line like the following should be used:


```
git+https://github.com/hdmf-dev/hdmf@237721cba907086b4cc52befe2ab616683ffa2c1#egg=hdmf
```

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "src/hdmf"]
+	path = src/hdmf
+	url = https://github.com/hdmf-dev/hdmf.git
```
jcfr: This is not needed. Checkout of the source will be handled in the `requirements.txt`

jcfr (changes requested):
While we can indicate that hdmf repository need to be checked out in requirements.txt, there is no way to update the setup.py do so the same thing. 

This means that until the hdmf wheel is available, user will have to explicitly install hdmf using something like 
```
pip install git+https://github.com/hdmf-dev/hdmf@dev#egg=hdmf
```

Note that having git installed is also required in that case.

I think finalizing the wheel upload would be the way forward.

requirements.txt
bendichter: why can't we just publish hdmf on pypi?

oruebel:
As the title suggests, this is still work in progress. One main reason for having HDMF as a submodule right now is to make the process of extracting HDMF more tractable. There is a lot to untangle, in particular on the CI, test, docs, and deployment side. Making HDMF a submodule at first is to make it easier to first separate the codes, then work on the CI, tests, docs, etc  @ajtritt has already create the hdmf pypi name https://pypi.org/project/hdmf/  and is working through the process of getting everything ready.

jcfr:
> Making HDMF a submodule at first is to make it easier to first separate the codes

To support this, the hdmf should be removed from requirements.txt and instead the setup.py should be updated. I will push a commit in few minutes.

jcfr:
The approach I proposed in this last commit only support the creation of valid wheels or regular install but doesnt:

```
pip install .
python test.py
```

but doesn't support the editable install (`pip install -e .` or `python setup.py develop`)

jcfr:
To  support editable install, the following workaround could be done:

```
pip install -e src/hdmf
pip install -e .
```

jcfr (dismissed):

ajtritt:
Thanks @jcfr for all your work, though I'm still not entirely sure what has been done here. As I think about this more, making HDMF a dependency in requirements.txt from the get go might be easier. 

My main concern about having them separated is users submitting issues on PyNWB that are actually issues with HDMF, but that's something that can be dealt with. I'm not worried about making the process of separating them more tractable--I would rather users have to deal with a single change than multiple changes. 

I can just focus on getting HDMF available on PyPI and then we can make it a dependency in requirements.txt.

jcfr:
> I'm still not entirely sure what has been done here

In short, it makes sure hdmf package (folder with an __init__.py) are
bundled in the same wheel.

> users submitting issues on PyNWB that are actually issues with HDMF,

If that helps, I think there is a way to move issue between project in
github.

t-b:
@ajtritt It think this can be closed right?

ajtritt:
no longer going this route

PR #814, opened 2019-02-05
fix conversion from ndarrays to dataframes

### LINKED ISSUE #810: DynamicTable.to_dataframe cannot handle new DynamicTableRegion objects
## 1) Bug

Running `to_dataframe` on a `DynamicTable` that contains a `DynamicTableRegion` no longer works due to the changes to `DynamicTableRegion`. This is a common practice on the Units table, which uses `DynamicTableRegion` to store electrodes (as illustrated below) so I expect many users will be affected by this bug.

### Steps to Reproduce
```python
from pynwb import NWBFile

from datetime import datetime
from dateutil.tz import tzlocal

nwbfile = NWBFile('aa', 'aa', datetime(2017, 4, 15, 12, tzinfo=tzlocal()))

device = nwbfile.create_device(name='trodes_rig123')
electrode_name = 'tetrode1'
description = "an example tetrode"
location = "somewhere in the hippocampus"
electrode_group = nwbfile.create_electrode_group(electrode_name,
                                                 description=description,
                                                 location=location,
                                                 device=device)
for idx in [1, 2, 3, 4]:
    nwbfile.add_electrode(idx,
                          x=1.0, y=2.0, z=3.0,
                          imp=float(-idx),
                          location='CA1', filtering='none',
                          group=electrode_group)

nwbfile.add_unit(id=1, electrodes=[1])
nwbfile.add_unit(id=2, electrodes=[1])

nwbfile.units.to_dataframe()
```

```python
  File "/Users/bendichter/dev/pynwb/test_units_to_dataframe.py", line 26, in <module>
    nwbfile.units.to_dataframe()
  File "/Users/bendichter/dev/pynwb/src/pynwb/core.py", line 1243, in to_dataframe
    data[name] = col[:]
  File "/Users/bendichter/dev/pynwb/src/pynwb/core.py", line 404, in __getitem__
    ret.append(self.__getitem_helper(i))
  File "/Users/bendichter/dev/pynwb/src/pynwb/core.py", line 397, in __getitem_helper
    return self.target[start:end]
  File "/Users/bendichter/dev/pynwb/src/pynwb/core.py", line 1360, in __getitem__
    raise ValueError("unrecognized argument: '%s'" % key)
ValueError: unrecognized argument: 'slice(0, 1, None)'
```

## Checklist

bendichter:
*convert nd arrays to list of arrays when converting to df
*add test

## Motivation

fix #810 

## How to test the behavior?
```python
from pynwb.core import DynamicTable, VectorData
import numpy as np 
import pandas as pd

col = VectorData(name='name', description='desc', data=np.array([[1,1,1],[2,2,2],[3,3,3]])) 
DynamicTable('test', 'desc', np.arange(3, dtype='int'), (col, )).to_dataframe()
```

## Checklist

### DIFF
--- a/src/pynwb/core.py
+++ b/src/pynwb/core.py
@@ -1240,10 +1240,10 @@ class DynamicTable(NWBDataInterface):
         data = {}
         for name in self.colnames:
             col = self.__df_cols[self.__colids[name]]
-            if col.data.ndim == 1:
-                data[name] = col[:]
-            else:
+            if isinstance(col.data, np.ndarray) and col.data.ndim > 1:
                 data[name] = [x for x in col[:]]
+            else:
+                data[name] = col[:]
 
         return pd.DataFrame(data, index=pd.Index(name=self.id.name, data=self.id.data))
 
--- a/src/pynwb/core.py
+++ b/src/pynwb/core.py
@@ -1240,7 +1240,10 @@ class DynamicTable(NWBDataInterface):
         data = {}
         for name in self.colnames:
             col = self.__df_cols[self.__colids[name]]
-            data[name] = col[:]
+            if col.data.ndim == 1:
+                data[name] = col[:]
+            else:
+                data[name] = [x for x in col[:]]
 
         return pd.DataFrame(data, index=pd.Index(name=self.id.name, data=self.id.data))
 
--- a/tests/unit/pynwb_tests/test_core.py
+++ b/tests/unit/pynwb_tests/test_core.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 
+import numpy as np
 import pandas as pd
 import unittest2 as unittest
 from dateutil.tz import tzlocal
@@ -206,6 +207,10 @@ class TestDynamicTable(unittest.TestCase):
 
         dynamic_table_region = DynamicTableRegion('dtr', [0, 1], 'desc', table=table)
         assert dynamic_table_region[slice(0, 1)]
+
+    def test_nd_array_to_df(self):
+        col = VectorData(name='name', description='desc', data=np.array([[1, 1, 1], [2, 2, 2], [3, 3, 3]]))
+        DynamicTable('test', 'desc', np.arange(3, dtype='int'), (col, )).to_dataframe()
 
 
 class TestNWBTable(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


src/pynwb/core.py
```diff
@@ -1240,7 +1240,10 @@ def to_dataframe(self):
         data = {}
         for name in self.colnames:
             col = self.__df_cols[self.__colids[name]]
-            data[name] = col[:]
+            if isinstance(col.data, np.ndarray) and col.data.ndim > 1:
+                data[name] = [x for x in col[:]]
```
oruebel, line 1244: I think this is the same as ``data[name] = col.tolist()`` correct? If so, I think using ``tolist`` is probably prefered

https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.ndarray.tolist.html

tests/unit/pynwb_tests/test_core.py
```diff
@@ -207,6 +208,10 @@ def test_indexed_dynamic_table_region(self):
         dynamic_table_region = DynamicTableRegion('dtr', [0, 1], 'desc', table=table)
         assert dynamic_table_region[slice(0, 1)]
 
+    def test_nd_array_to_df(self):
+        col = VectorData(name='name', description='desc', data=np.array([[1, 1, 1], [2, 2, 2], [3, 3, 3]]))
+        DynamicTable('test', 'desc', np.arange(3, dtype='int'), (col, )).to_dataframe()
```
ajtritt: Can you add assert statement checks to make sure the dataframe is converted appropriately?

ajtritt (changes requested):

src/pynwb/core.py
ajtritt: `tolist` makes a list of lists.... I'm not sure if that's preferable to a list of numpy arrays. @bendichter ?

ajtritt (changes requested):

bendichter:
Yeah I prefer a list of arrays

src/pynwb/core.py
oruebel: Ok, I see. Thanks for the clarification, that makes sense.

oruebel (approved):

ajtritt (approved):

PR #821, opened 2019-02-18
add spike_event_series link to ElectrodeGroup

bendichter:
## Motivation

Part of plan to make spike waveforms accessible from Units table. See https://github.com/NeurodataWithoutBorders/nwb-schema/issues/248

This is not working and I can't figure out why. Help @ajtritt?

### DIFF
--- a/src/pynwb/ecephys.py
+++ b/src/pynwb/ecephys.py
@@ -20,24 +20,24 @@ class ElectrodeGroup(NWBContainer):
                      'description',
                      'location',
                      'device',
-                     {'name': 'spike_event_series', 'doc': 'doc', 'child': False})
+                     {'name': 'electrical_series', 'doc': 'doc', 'child': False})
 
     @docval({'name': 'name', 'type': str, 'doc': 'the name of this electrode'},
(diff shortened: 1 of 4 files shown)


src/pynwb/ecephys.py
```diff
@@ -19,20 +19,25 @@ class ElectrodeGroup(NWBContainer):
     __nwbfields__ = ('name',
                      'description',
                      'location',
-                     'device')
+                     'device',
+                     {'name': 'spike_event_series', 'doc': 'doc', 'child': False})
```
ajtritt: This should be an _ElectricalSeries_, and _doc_ should be something that makes sense. The docstring will get rendered on the API docs, so this is what users will see if they use the API docs for reference.

ajtritt (dismissed):

src/pynwb/ecephys.py
bendichter: Yeah I can change the doc, I just wanted to draw your attention to the error so we can fix it. Why should this be an ElectricalSeries?

PR #830, opened 2019-02-20
add methods to TimeSeries of aligning to events

bendichter:
## Motivation

Addresses a common need to align `TimeSeries` data to trials

## How to test the behavior?
```python
import numpy.testing
import datetime
from pynwb import TimeSeries, NWBFile

nwbfile = NWBFile(identifier='foo', session_start_time=datetime.datetime.now(), session_description='bar')
nwbfile.add_trial(start_time=5., stop_time=10.)
nwbfile.add_trial(start_time=15., stop_time=20.)
ts = TimeSeries(name='a', data=np.arange(1000), timestamps=np.arange(1000), unit='m')
nwbfile.add_acquisition(ts)
numpy.testing.assert_equal(ts.align_by_trials(), [[6, 7, 8, 9, 10], [16, 17, 18, 19, 20]])
```

## Checklist

### DIFF
--- a/src/pynwb/base.py
+++ b/src/pynwb/base.py
@@ -1,11 +1,13 @@
 from warnings import warn
+from bisect import bisect
+import numpy as np
 
 from collections import Iterable
 
 from .form.utils import docval, getargs, popargs, fmt_docval_args, call_docval_func
 from .form.data_utils import AbstractDataChunkIterator, DataIO
 
-from . import register_class, CORE_NAMESPACE
+from . import register_class, CORE_NAMESPACE, NWBFile
 from .core import NWBDataInterface, MultiContainerInterface, NWBData
 
 _default_conversion = 1.0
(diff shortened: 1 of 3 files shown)


bendichter:
@VBaratham

ajtritt:
I think this already exists here. 

https://github.com/NeurodataWithoutBorders/pynwb/blob/a9429c93f29763b9ebe9022b099afcffbc6be493/src/pynwb/epoch.py#L65

There is a way to implement this without requiring users explicitly call a method on each TimeSeries. I am not sure what the API should like though. Is there an issue this PR addresses that would have requirements specced out?

bendichter:
@ajtritt Thanks for pointing me to this code- I was not aware of it. It looks like we used very similar approaches for getting the data (though it looks like you may have handled the roundtrip case better than me).

This came out of a conversation with @VBaratham and addresses his request for the ability to trialize `TimeSeries` objects. He's far from the first user to request this feature. I prefer this as a method of `TimeSeries` because we can provide a convenience function for automatically aligning to `trials`, since that will be the most common operation. If we make this a method of a `TimeIntervals` object, it's unclear what `TimeSeries` the user wants to align. It also just makes more intuitive sense to me that trialization is an action on the `TimeSeries` object, so it should be a method of that class. In my experience, we generally want to trialize specific time series objects, and users are unlikely to require trialization of every `TimeSeries` objects in the file all at once.

We also need a lot of flexibility in timing:
1) Ending a trial either using 'stop_time' or a duration from 'start_time'
2) Aligning to different events in the trial, such as stimulus onset, or behavior onset.
3) Expanding the window before the start of the trial and/or after the end
4) Aligning to non-interval events, such as annotations
5) Aligning to intervals other than `trials`

All of these are features that would be used regularly in analysis.

From what I can tell, the code you pointed to is not exposed to the user in the API, and does not offer any of the above flexibility.

ajtritt:
moving this to #832

PR #843, opened 2019-03-10
fix slice indexing of DynamicTableRegion and add test

### LINKED ISSUE #842: issue indexing dynamic table region with slice
## 1) Bug

There is currently a bug when using a `slice` to access a region of a `DynamicTableRegion`. The slice should index `self.data` which should then be used to index `self.target` (`DynamicTable`). Instead, the `slice` indexes `[0, 1, 2, ...]`.

(This was my mistake. I did something goofy and it looks like I didn't have a good understanding of slice indexing at the time. The fix is pretty straightforward.)

If you are reporting a bug please provide the following:
### Steps to Reproduce
```python
from pynwb.core import DynamicTableRegion
dtr = DynamicTableRegion('name', [0,1,1], 'desc', nwb.electrodes)
dtr[:2]
```

## Checklist

bendichter:
## Motivation

fix #842 

## Checklist

### DIFF
--- a/tests/unit/pynwb_tests/test_core.py
+++ b/tests/unit/pynwb_tests/test_core.py
@@ -206,8 +206,9 @@ class TestDynamicTable(unittest.TestCase):
     def test_indexed_dynamic_table_region(self):
         table = self.with_columns_and_data()
 
-        dynamic_table_region = DynamicTableRegion('dtr', [0, 1], 'desc', table=table)
-        assert dynamic_table_region[slice(0, 1)]
+        dynamic_table_region = DynamicTableRegion('dtr', [0, 1, 1], 'desc', table=table)
+        fetch_ids = [x[1] for x in dynamic_table_region[:3]]
+        assert fetch_ids == [1, 2, 2]
 
     def test_nd_array_to_df(self):
         data = np.array([[1, 1, 1], [2, 2, 2], [3, 3, 3]])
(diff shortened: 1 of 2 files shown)


bendichter:
The error is due to https://github.com/AllenInstitute/AllenSDK/issues/426.  It looks like they are aware of the problem and working to resolve it quickly.

tests/unit/pynwb_tests/test_core.py
```diff
@@ -206,8 +206,9 @@ def test_extra_columns(self):
     def test_indexed_dynamic_table_region(self):
         table = self.with_columns_and_data()
 
-        dynamic_table_region = DynamicTableRegion('dtr', [0, 1], 'desc', table=table)
-        assert dynamic_table_region[slice(0, 1)]
+        dynamic_table_region = DynamicTableRegion('dtr', [0, 1, 1], 'desc', table=table)
+        fetch_ids = [x[1] for x in dynamic_table_region[:3]]
+        assert fetch_ids == [1, 2, 2]
```
ajtritt: Is there a reason you're not using [assertListEqual](https://docs.python.org/3.7/library/unittest.html#unittest.TestCase.assertListEqual) or [assert_array_equal](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.testing.assert_array_equal.html)?

bendichter: no, I'm happy to use those instead. What is the advantage?

ajtritt: The TestCase assert methods give more informative error message in the event that the test fails.

bendichter: ah ok that makes sense

ajtritt (approved):

PR #847, opened 2019-03-13
fix/mpi_io

### LINKED ISSUE #362: parallel io
## Feature Request

The basic necessity is to add arguments to `NWBHDF5IO` and similar utilities to facilitate opening an NWB file with MPI and ensure that data can be read/written in parallel.

In addition, we would like to implement parallel reading/writing tools for NWB. We have developed similar tools in [neuroh5](https://bitbucket.org/ivan_raikov/neuroh5/src).

bendichter:
## Motivation

Add MPI functionality to `NWBHDF5IO`
fix #362

## How to test the behavior?
Creating an automatic test for this will take some work, since it requires HDF5 to be compiled against MPI. For now, you can test with the following code:
```python
from mpi4py import MPI                                                                                                
import h5py                                                                                                           
                                                                                                                      
from dateutil import tz                                                                                               
                                                                                                                      
from pynwb.form.backends.hdf5.h5tools import HDF5IO                                                                   
from pynwb import NWBHDF5IO, NWBFile, TimeSeries                                                                      
from datetime import datetime                                                                                         
                                                                                                                      
start_time = datetime(2018, 4, 25, 2, 30, 3, tzinfo=tz.gettz('US/Pacific'))                                           
                                                                                                                      
fname = 'test_parallel_pynwb.nwb'                                                                                     
                                                                                                                      
rank = MPI.COMM_WORLD.rank  # The process ID (integer 0-3 for 4-process run)                                          
                                                                                                                      
if not rank:                                                                                                          
    nwbfile = NWBFile('aa', 'aa', datetime.now())                                                                     
    nwbfile.add_acquisition(TimeSeries('ts_name', description='desc', data=[0, 0, 0, 0], rate=100., unit='m'))        
    with NWBHDF5IO(fname, 'w') as io:                                                                                 
        io.write(nwbfile)                                                                                             
                                                                                                                      
MPI.COMM_WORLD.Barrier()                                                                                              
                                                                                                                      
with NWBHDF5IO(fname, 'a', comm=MPI.COMM_WORLD) as io:                                                                
        nwbfile = io.read()
        print(rank)                                                                                           
        nwbfile.acquisition['ts_name'].data[rank] = rank
```
Run the above with `mpiexec -n 4 python test_parallel_script.py`
The final nwb file should have `nwbfile.acquisition['ts_name'].data[:] == [0, 1, 2, 3]`

## Checklist


bendichter:
@VBaratham's work. I'm just creating the PR

oruebel:
@ajtritt fyi, this has changes in  HDMF

VBaratham:
Just thinking aloud... but I had some second thoughts about this. I think there needs to be some more logic to make sure h5 parallel is used "properly" (eg. collective operations are performed on each rank, etc.) Or should this be the user's responsibility? I don't think so, because the user doesn't know what hdf5 operations are performed for a given nwb call.

What do others think?

bendichter:
@VBaratham I totally agree that we should have functions to allow users to easily leverage parallel computing without knowing the details of MPI, and we should also support low level mpi like the test above for users who want to write their own MPI scripts. I think the best first step is this low-level support, then we can decide how we want to help users automatically use MPI

oruebel:
For data read, allowing a user to pass in comm and using the mpio driver should be fine. 

Ultimately, I agree that the HDF5IO backend should directly support parallel write, however, the story there is a bit more complicated. E.g., 

1. In the above example you do ``data[rank] = rank`` to write different values for each rank to different parts of the array. When you do this on the backend, this means that you have to somehow communicate both the values as well as the hyperslap selection to define where each rank should write its data to. In principle you could do this right now by having each rank supply a ``DataChunkIterator`` for your timeseries ``data``, each of which would supply part of the data relevant for the rank. From an interface perspective, I think this would make sense and would build on capabilities we already have. However, I'm not sure if this would actually work with mpio on the backend right now.
2. Because certain operations have to be done collectively we need to make sure that all of this handled correctly in ``HDF5IO``. I don't think this has been tested. 
3. For performance you want to support optimizations, e.g., MPI collective I/O for datasets (see http://docs.h5py.org/en/stable/whatsnew/2.6.html#add-mpi-collective-i-o-support). These sort of things are not implemented yet.

Ultimately, I think all of this is doable and a lot of pieces are in place to make this work, but I think it will take some work and testing to make this work fully. In practice, I'd expect that for write, there are typically a few select data arrays that you actually want to write in parallel. For that, the strategy from the example above, where you setup the NWB:N file on one rank and then open the file with MPI on all ranks to write the data for the few arrays that require parallel I/O,  is I think a reasonable approach. In contrast to the above example, however, when you create the file from a single rank, you probably do not want actually populate the whole array with data, but simply just create the Dataset and then populate it with the actual data in parallel later. To do this you could us the ``H5DataIO`` wrapper to supply the backend with the instructions on how to setup the dataset to avoid the overhead for initializing the dataset when you set up the file.

bendichter:
That plan makes sense to me. Does H5DataIO have the capability to initialize a dataset shape without setting the values?

oruebel:
> Does H5DataIO have the capability to initialize a dataset shape without setting the values?

With ``H5DataIO`` alone, the answer right now is, No. Although, I don't think it should be too hard to modify ``H5DataIO`` and ``HDF5IO`` to support this. However, you can do a slightly different trick and simply use an empty ``DataChunkIterator``. 

Lets say you want to create an empty dataset with a ``shape=(100000, 100)`` what you can do is set 

```
data = DataChunkIterator(data=None, 
                        maxshape=(100000,100), 
                        dtype=np.dtype('float'))
```

In this way, the backend will create the dataset but since the ``DataChunkIterator`` never produces any data it won't initialize it.  In this case  both ``shape`` and ``maxshape`` of the dataset would be set to ``(100000, 100)``. To control this further you can combine it with ``H5DataIO`` and do something like:

```
data = H5DataIO(DataChunkIterator(data=None,   maxshape=(100000, 100), dtype=np.dtype('float')) , 
                                  chunks=(10,10), 
                                  maxshape=(None, None))
``` 

in which case your dataset would be initialized with a shape of ``(100000, 100)`` and maxshape of ``(None, None)`` and your own custom chunking.  Honestly, I think this should cover most practical cases where you want to create an empty dataset.  

One caveat with using ``DataChunkIterator`` this way is that ``H5DataIO.__chunked_iter_fill__`` will always enable chunking, i.e., one thing you really can't do with this right now is create an empty unchunked dataset with a fixed shape.

bendichter:
@oruebel thanks for the detailed description! It seems like maybe it would be best to create a guide for using MPI. I'll test out this chunking stuff with the test code above

oruebel:
I agree that we need to do an MPI tutorial at some point

bendichter:
since we have moved the backend to the hdmf repo, I am closing this PR in favor of

https://github.com/hdmf-dev/hdmf/pull/14

bendichter:
@VBaratham can you take out the changes to src/pynwb/form/backends/hdf5/h5tools.py? That file has been moved to another package. I have submitted a PR for the changes [here](https://github.com/hdmf-dev/hdmf/pull/14/files)

VBaratham:
@bendichter Done!

bendichter:
@VBaratham ok great, could you also merge the latest dev? 

You are developing this in your own private fork and have exclusive privileges in that fork. If you would like us to help with these types of things in the future, it might be easier to develop in a branch of this repo. That would allow me to merge the latest dev for you, for instance.

t-b:
Just from the sidelines:

> You are developing this in your own private fork and have exclusive privileges in that fork. If you would like us to help with these types of things in the future, it might be easier to develop in a branch of this repo. That would allow me to merge the latest dev for you, for instance.

I'm not a fan of this development strategy. The person developing the branch just needs to rebase against latest origin/dev and that's it. These additional merge commits also make for a really messy history.

VBaratham:
@bendichter Sure, merged. I'm happy to do it either way, but I didn't think I could create branches on this respository

bendichter:
You might be right I'm not sure exactly how we set up the permissions

oruebel:
> You might be right I'm not sure exactly how we set up the permissions

Typically, for core developments that we expect to be merged quickly we often use branches. But for larger developments and external contributions forks are usually a bit cleaner.

bendichter:
@VBaratham I'd like to merge this branch, but I am running into a little logistical issue. Every time the dev branch changes, we need to merge those changes and run the CI tests before merging this PR into dev. When changes are created via a branch, I can run this first merge myself, but since this is a fork and I do not have edit permissions, you must update. 

The issue is just one of syncing up our activities. You update, but by the time I circle back and want to merge this, some other change has been made to dev which require another update and test which I have to ask you to perform since, again, I don't have permissions. 

I'd like to get these changes in. I think the easiest way will be for you to give me edit permissions for your fork of this repo (which I will use solely for merging)

VBaratham:
Hi Ben,
I just added you to my fork, thanks for taking care of this! Let me know if you need anything else from me

oruebel (approved):

PR #850, opened 2019-03-13
Add HDMF to PyNWB

ajtritt:
## Motivation

HDMF has been moved into its own package.


test.py
```diff
@@ -14,7 +14,7 @@
 import traceback
 import unittest2 as unittest
 
-flags = {'form': 1, 'pynwb': 2, 'integration': 3, 'example': 4}
+flags = {'pynwb': 2, 'integration': 3, 'example': 4}
```
oruebel, line 17: Very minor, nitpicky thing. Should this change to ``flags = {'pynwb': 1, 'integration': 2, 'example': 3}`` just to have the numbering of options start at 1. Just ignore if this causes trouble.

oruebel (approved):
The changes all look good to me. I think this can be merged as is.

test.py
ajtritt: I didn't see this until after already merged... I'll get it on the next fix

oruebel: No problem. Its such a minor thing anyways.

PR #856, opened 2019-03-21
Bug/docs

ajtritt:
## Motivation

Docs have not been building on Read the Docs

## Checklist


ajtritt:
Looks like we had the sphinx version pinned, which was causing the problem. I updated the version and now the docs are building!

https://readthedocs.org/projects/pynwb/builds/8819242/

oruebel (approved):

oruebel:
> Looks like we had the sphinx version pinned, which was causing the problem. I updated the version and now the docs are building!

Great! We had to pin Sphinx a while back as a workaround to avoid a bug with the then latest Sphinx version (or maybe it was a combination of Sphinx+RTD, I don't remember the details). Good that the current version of Sphinx is working now.

PR #857, opened 2019-03-24
remove json

bendichter:
## Motivation

Removing json from docs since we are deprecating support for json spec files

## Checklist

### DIFF
--- a/docs/source/extensions.rst
+++ b/docs/source/extensions.rst
@@ -352,7 +352,7 @@ We can now create the sources of our documentation as follows:
                  --external_description my_extension_source/docs/mylab_description.rst \  (Optional)
                  --external_release_notes my_extension_source/docs/mylab_release_notes.rst \  (Optional)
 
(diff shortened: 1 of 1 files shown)


oruebel:
@ajtritt @bendichter when I updated the docutils I only intended to remove JSON as part of the rendering of the docs to Sphinx. This was mainly a decision to simplify the docutils and to have a consistent visual for the docs. Since we are  1) currently not using any YAML-specific features in the specification language and 2) since JSON is valid YAML, I don't see a pressing reason why we would need to say only YAML would be allows to save specfication files. HDMF, PyNWB (and in turn the docutils) currently still support JSON as input files. Allowing JSON will be useful when we need the specification for web-based data exchange since JSON is more common there. My vote is to still support JSON as input (and output) for HDMF and PyNWB. Only for rendering the docs in Sphinx my vote is to just use YAML because it simplifies the docs and looks more human-readable in the docs.

bendichter:
Oh ok my mistake. I saw removal of json and wanted to make sure we were consistent.

PR #859, opened 2019-03-27
Enh/nwbio

### LINKED ISSUE #858: NWBIO
## Feature Request
I would like an `NWBIO` function that takes an NWB of any backend, automatically determines the backend, and calls the proper `NWB<backend>IO` class (currently only `NWBHDF5IO` is available. This would allow us to create backend-agnostic code for reading from a file path, i.e.

```python
with NWBIO('path/to/file', 'r') as io:
    io.read()
````

This came up as a need in our collaboration with CaImAn.

## Checklist

### PR DESCRIPTION
**Pull request by bendichter**
## Motivation

add function NWBIO, fix #858 

## How to test the behavior?
```python
from pynwb import NWBIO
fpath = 'path_to/YutaMouse41-150903.nwb'
with NWBIO(fpath, 'r') as io:
    print(io.read())
```

## Checklist

### DIFF
--- a/docs/source/api_docs.rst
+++ b/docs/source/api_docs.rst
@@ -20,7 +20,7 @@ API Documentation
    Miscellaneous <pynwb.misc>
    Epoch <pynwb.epoch>
    PyNWB <pynwb>
-   Flexible Object-Relational Mapping <pynwb.form>
+   Flexible Object-Relational Mapping <hdmf>
 
 
 
--- a/docs/source/building_api.rst
+++ b/docs/source/building_api.rst
@@ -85,7 +85,7 @@ If your specification extension contains a table definition i.e. a dataset with 
 the :py:class:`~pynwb.core.NWBTable` class to represent this specification. Since :py:class:`~pynwb.core.NWBTable`
 subclasses :py:class:`~pynwb.core.NWBData` you can still use ``__nwbfields__``. In addition, you can use the
 ``__columns__`` class property to specify the columns of the table. ``__columns__`` should be a list or a tuple of
-:py:func:`~pynwb.form.utils.docval`-like dictionaries.
+:py:func:`~hdmf.utils.docval`-like dictionaries.
 
 The following example demonstrates how to define a table with the columns ``foo`` and ``bar`` that are of type
 str and int, respectively. We also register the class as the reppresentation of the *neurodata_type* "MyTable"
--- a/docs/source/extensions.rst
+++ b/docs/source/extensions.rst
@@ -268,9 +268,9 @@ ObjectMapper : Customizing the mapping between NWBContainer and the Spec
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 If your :py:class:`~pynwb.core.NWBContainer` extension requires custom mapping of the :py:class:`~pynwb.core.NWBContainer`
-class for reading and writing, you will need to implement and register a custom :py:class:`~pynwb.form.build.map.ObjectMapper`.
+class for reading and writing, you will need to implement and register a custom :py:class:`~hdmf.build.map.ObjectMapper`.
 
-:py:class:`~pynwb.form.build.map.ObjectMapper` extensions are registered with the decorator :py:func:`~pynwb.register_map`.
+:py:class:`~hdmf.build.map.ObjectMapper` extensions are registered with the decorator :py:func:`~pynwb.register_map`.
 
 .. code-block:: python
 
--- a/docs/source/make_roundtrip_test.rst
+++ b/docs/source/make_roundtrip_test.rst
@@ -14,7 +14,7 @@ with the following command::
     $ python test.py -i
 
 The roundtrip test will generate a new NWB file with the name ``test_<CLASS_NAME>.nwb`` where ``CLASS_NAME`` is
-the class name of the :py:class:`~pynwb.form.container.Container` class you are roundtripping. The test
+the class name of the :py:class:`~hdmf.container.Container` class you are roundtripping. The test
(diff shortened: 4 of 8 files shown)


bendichter:
I'm seeing if the import works across OS's, then I'll write some tests

bendichter:
I'm having trouble finding a version of python-magic that works across OS.

* https://pypi.org/project/python-magic/ does not include libmagic
* https://pypi.org/project/python-magic-win64/#files only works on Windows
* https://pypi.org/project/python-magic-bin/0.4.14/#files only works on OSX and Windows
* https://anaconda.org/conda-forge/python-magic only works on OSX and Linux

oruebel:
Which specific use-case are you trying to solve here? For write I think this is not applicable. For read I can see why you may want to have something like this, although even then, I think the user should know what kind of backend is being used, rather than hiding it from the user.. 

I see a couple of problems with the current implementation in this PR:
* As far as I know using Python-magic only works for file-types. However, file-type alone will likely not be sufficient to determine the correct backend. E.g., both Zarr and Exdir support a files+folder approach but with different mappings. Also, if you have a cloud-based or database backend you may not even have a physical file that you can check for type.
* Different backends may have different additional input parameters for reading and writing and if the selection of the backend is hidden in a function, the user does not know which parameters to use. 
* You may not be able to uniquely determine a suitable backend, e.g., imagine the case where you have a hybird HDF5+Database backend. In this case you cannot tell which of the available backends you may want to use. 
* Pushing the logic to determine the backend into a separate class means that a backend developer no longer just needs to implement HDMFIO but also needs to update this function/class.

I would propose a slightly different solution. 
1. Have a function to ``get_backend_class`` which would return the suggested backend class to be used. The function would return a list of all backends that can open given NWB:N file. In most cases the returned list will have a length of 1 but it may have 0 or multiple ones as well. In this way you are solving the problem for the user in 99% of all cases but the user still knows which backend is being used and the user has a chance to mitigate ambiguous cases. 
2. The logic to determine if a backend can open a particular NWB:N file should be part of the backend itself. E.g., you could have a static function ``supports_read_of(...)`` (or something like this) in HDMFIO that the backend developer can implement to determine if the backends supports reading data from the given file. 
3. The ``get_backend_class`` function would then simply iterate over all available backends and return a list of all backends that apply. I.e., if a new backend is being added the function will not need to be updated as long as you have a way of finding all classes inherit from HDMFIO.

bendichter:
> Which specific use-case are you trying to solve here?

This is work to support the collaborative project with `caiman`. I'm trying to work with them to read two-photon imaging data from an NWB file. Their user-facing syntax is `caiman.load('path/to/file')`, which can take several different filetypes: .avi, .hdf5, etc.. The load function then uses the extensions to infer what the filetype is and loads the data appropriately. Since HDF5 is the only currently supported backend, this could easily be implemented with

```python
if extension == '.nwb':
    with NWBHDF5IO('path/to/file', 'r') as io:
        nwb = io.read()
        data = nwb.acquisition['TwoPhotonSeries1'].data[:]
```

however this assumes that the backend is HDF5, so any backend we implement in the future would not be supported. A main point of using pynwb as opposed to h5py would be to support different backends, so including an NWBHDF5IO call. It seems to me that other projects using nwb should be able to rely on pynwb to infer the backend and read that data appropriately into the api. I proposed a different layout- something like:

```python
io = NWBHDF5IO('path/to/file', 'r'):
nwb = io.read()
caiman.load_nwb(nwb.acquisition['TwoPhotonSeries1'])
```
where the caiman library offers a function that takes in a `pynwb.TimeSeries` object and the backend-specific code is handled beforehand, but they would rather handle data loading inside of their load function. This does make things much easier and more uniform for the user, so from a usability perspective I agree with them. 

In order to get data to read from any .nwb file given just the filepath, we'd need some sort of automatic filetype detection function. What exactly this function should look like is a bit unclear at this point since we only currently support hdf5, but I wanted to at least create a function as a placeholder that can determine whether or not a .nwb file is hdf5, and supply that function for external repos. I think it would be useful to have some read option that is backend-agnostic that we can supply to external repos. 

I see that there are some advantages to breaking this into its own `get_backend_class` function in hdmf. I'll take a look at doing it that way.

bendichter:
> Which specific use-case are you trying to solve here?

This is work to support the collaborative project with `caiman`. I'm trying to work with them to read two-photon imaging data from an NWB file. Their user-facing syntax is `caiman.load('path/to/file')`, which can take several different filetypes: .avi, .hdf5, etc.. The load function then uses the extensions to infer what the filetype is and loads the data appropriately. Since HDF5 is the only currently supported backend, this could easily be implemented with

```python
if extension == '.nwb':
    with NWBHDF5IO('path/to/file', 'r') as io:
        nwb = io.read()
        data = nwb.acquisition['TwoPhotonSeries1'].data[:]
```

however this assumes that the backend is HDF5, so any backend we implement in the future would not be supported. A main point of using pynwb as opposed to h5py would be to support different backends, so including an NWBHDF5IO call. It seems to me that other projects using nwb should be able to rely on pynwb to infer the backend and read that data appropriately into the api. I proposed a different layout- something like:

```python
io = NWBHDF5IO('path/to/file', 'r'):
nwb = io.read()
caiman.load_nwb(nwb.acquisition['TwoPhotonSeries1'])
```
where the caiman library offers a function that takes in a `pynwb.TimeSeries` object and the backend-specific code is handled beforehand, but they would rather handle data loading inside of their load function. This does make things much easier and more uniform for the user, so from a usability perspective I agree with them. 

In order to get data to read from any .nwb file given just the filepath, we'd need some sort of automatic filetype detection function. What exactly this function should look like is a bit unclear at this point since we only currently support hdf5, but I wanted to at least create a function as a placeholder that can determine whether or not a .nwb file is hdf5, and supply that function for external repos. I think it would be useful to have some read option that is backend-agnostic that we can supply to external repos. 

I see that there are some advantages to breaking this into its own `get_backend_class` function in hdmf. I'll take a look at doing it that way.

PR #867, opened 2019-03-31
Fix data read when printing

### LINKED ISSUE #812: calling __str__ on a loaded `units` table takes a long time (~10 secs)
I suspect `__str__` is loading data it does not need to load

bendichter:
## Motivation
fix #812

Some print statements were taking too long. The problem was that sometimes print would attempt to convert a list of `NWBBaseType`s into an array, which would involve loading data from HDF5. I added a check to make sure the list contains `NWBBaseType`s just go directly to converting to string.

## How to test the behavior?
```python
from pynwb import NWBHDF5IO

fpath = '/Users/bendichter/Desktop/Soltesz/data/DG_PP_spiketrain_12142018.nwb'

with NWBHDF5IO(fpath, 'r') as io:
    nwb = io.read()
    print(nwb.units)
```

## Checklist


oruebel (dismissed):

oruebel:
Looks like that code is currently not being tested, but otherwise the check seems Ok.

oruebel (dismissed):

oruebel (approved):

PR #872, opened 2019-04-02
change __nwbfields__ to __fields__

### LINKED ISSUE #871: __nwbfields__ to __fields__
(from convo with @ajtritt)

As the next step of the migration of HDMF, we should rename `__nwbfields__` to something non-specific like `__fields__`. This would allow us to generalize the code to dynamically create nested classes from spec files (https://github.com/hdmf-dev/hdmf/pull/30). 

## Checklist

bendichter:
fix #871

In order to get the whitespace to line up after changing from `__nwbfields__` to `__fields__` I ran a whitespace optimizer, which ended up catching a lot of other whitespace issues as well. I looked over the changes and I think they really do improve the readability of the project, particularly with improving the consistency of indenting of YAML files. I know we don't usually accept whitespace PRs but I'm hoping we can make an exception here.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,41 +124,41 @@ jobs:
     docker:
       - image: circleci/python:2.7.14-jessie
     environment:
-      - TEST_TOX_ENV: "py27"
-      - COVERAGE_TOX_ENV: "coverage-py27"
-      - BUILD_TOX_ENV: "build-py27"
-      - TEST_WHEELINSTALL_ENV: "wheelinstall-py27"
+     - TEST_TOX_ENV: "py27"
+     - COVERAGE_TOX_ENV: "coverage-py27"
+     - BUILD_TOX_ENV: "build-py27"
+     - TEST_WHEELINSTALL_ENV: "wheelinstall-py27"
     <<: *ci-steps
 
   python35:
     docker:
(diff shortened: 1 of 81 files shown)


t-b:
@bendichter Is it possible to have this as part of CI? You don't want the next person changing the YAML files to "mess" it up again.

bendichter:
@t-b Is there a way to get flake8 to cover YAML files? In my googling all I can find is the YAML config files for flake8

t-b:
@bendichter What you need is a pretty printer for YAML, maybe it is easiest to just use ruamel for that. See https://stackoverflow.com/a/31005595 for an example.

oruebel:
@bendichter for the format specification YAML files this is problematic because it looks like it changes the indent and other whitspaces which means if you use the spec API to update the files then the write from PyNWB/HDMF for the YAML files is probably not consistent with these changes, i.e., any time someone would make a simple change it would result in updates to the whole YAML. To avoid this the write for the YAML would need to be consistent with the changes here. For now, I would suggest to not do the updates to the specification YAML files.

bendichter:
@oruebel As far as I know we don't use the spec API to generate core YAML files, but I do see that this would make their indentation inconsistent with any files that are generated. Let me figure out how to undo that in git...

PR #873, opened 2019-04-03
Add isinstance check in to_dataframe for h5py Dataset

nicain:
## Motivation

to_dataframe on a DynamicTable had a corner case when reading from a file, as opposed to an im-memory object.  Isinstance is used to detect if a column consists of data with each row being a numpy array, and removing the outer dimension to build the column.  This PR catches the corner case.

test_pandas_roundtrip should be enhanced with a fixture that uses a file on tempdir, but Ill leave that to be organized with other test fixtures in the test directory

## Checklist

### DIFF
--- a/src/pynwb/core.py
+++ b/src/pynwb/core.py
@@ -1,4 +1,4 @@
-from h5py import RegionReference
+from h5py import RegionReference, Dataset
 import numpy as np
 import pandas as pd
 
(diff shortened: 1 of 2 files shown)


bendichter (dismissed):

bendichter:
thanks for the catch @nicain, this looks good to me. Let me see if I can throw together a test.

oruebel (approved):

PR #874, opened 2019-04-04
epoch start and stop time precision in spec: float to double

nicain:
## Motivation

Epoch start and stop time precision is defined in docval as float, which is double precision in python (see https://github.com/python/cpython/blob/e42b705188271da108de42b55d9344642170aa2b/Include/floatobject.h#L5).  This commit modifies the spec to conform.  Note: current behavior is that user-provided precision will be silently coerced to the precision defined in the spec, presumably by hdmf.

### DIFF
--- a/src/pynwb/data/nwb.epoch.yaml
+++ b/src/pynwb/data/nwb.epoch.yaml
@@ -11,7 +11,7 @@ groups:
   datasets:
   - neurodata_type_inc: VectorData
     name: start_time
-    dtype: float
+    dtype: double
     doc: Start time of epoch, in seconds
     attributes:
     - name: description
@@ -20,7 +20,7 @@ groups:
       value: start time of the epoch, in seconds
   - neurodata_type_inc: VectorData
     name: stop_time
-    dtype: float
+    dtype: double
     doc: Stop time of epoch, in seconds
     attributes:
     - name: description


t-b:
I thought the widhts/precisions in the spec are always minimum? So that this kind of change would not be required?

oruebel:
> I thought the widhts/precisions in the spec are always minimum?

Correct. I don't think this needs to be fixed in the schema but may need to be changed in PyNWB to accept np.float32 and np.float64.

bendichter:
@oruebel yes, that would make it more consistent with the schema. Should we do that with all entries of `float` in docvals? If so, maybe it would be better to define a data type group called 'float' that would accept (float, np.float32, and np.float64)?

ajtritt:
>  If so, maybe it would be better to define a data type group called 'float' that would accept (float, np.float32, and np.float64)?

Great idea @bendichter!

bendichter:
Looking closer at the code, it looks like this is already supported by 

https://github.com/hdmf-dev/hdmf/blob/495dacc2b1e6a9d5f4f4867684a3bc35f90327d2/src/hdmf/utils.py#L89-L93

I think this would also work as a type macro. Would it be better to define it that way? 

Either way, does it make sense to change all of the `float` instances in docval statements to `'float'`?

bendichter:
I did a little investigating and @nicain is absolutely right. `float` values are downgraded to `np.float32` (see https://github.com/hdmf-dev/hdmf/issues/32). 

docval ended up being a red flag. You can supply `float`, `np.float32`, and `np.float64` to an arg that is defined as `float` in docval.

bendichter:
@nicain The change you made, while it does fix your problem, is a schema-breaking change, which would make previously 2.0-valid NWB data invalid, so we would prefer another route. Do you mind if I close this PR in favor of https://github.com/hdmf-dev/hdmf/pull/33?

PR #894, opened 2019-04-22
fix python3 patch revision number for travis

nicain:
Previous attempt at adding python3.7 coverage on travis failed because the python3.7 patch revision requested was not available on the Travis CI system.  This commit should fix this problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
     - os: osx
       language: generic
       env:
-        - TRAVIS_PYTHON_VERSION=3.7.3
+        - TRAVIS_PYTHON_VERSION=3.7.0
         - TEST_TOX_ENV=py37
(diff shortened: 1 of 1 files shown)


nicain:
@dorukozturk Any idea why travis is not getting swept in automatically for this branch build (I don't see it the list of checks that was run for this PR)?

dorukozturk:
Yes, because we disabled travis for branches :). It only runs after a merge to master. The argument was long test times on macos. So unfortunately you will be able to test if it is working or not only after a merge to master.

dorukozturk (approved):

dorukozturk:
@nicain Got to do this one more time :) 
```
The following versions contain `3.7.0' in the name:
  3.7.0b3
  miniconda-3.7.0
  miniconda3-3.7.0
```
It looks like 3.7.0b3 is what they have. So changing the version to 3.7.0b3 should work.

PR #902, opened 2019-04-26
modular timestamps storage

### LINKED ISSUE #901: Having trouble linking timestamps across files
I'm trying to take advantage of modular data storage to break up some large files, following [this tutorial](https://pynwb.readthedocs.io/en/stable/tutorials/general/linking_data.html#sphx-glr-tutorials-general-linking-data-py). This works great, except I keep getting errors when trying to link timestamps as well as data.

Reproduce: https://gist.github.com/NileGraddis/7c1b10442f8b053d6ad5e8d6e968c552
pynwb version: 1.0.2
platform: ubuntu 18.04.2

Thanks in advance!

NileGraddis:
Fix #901 

Implements a fix for modular (cross-file) storage of timeseries timestamps and a roundtrip test of this functionality.

### DIFF
--- a/src/pynwb/io/base.py
+++ b/src/pynwb/io/base.py
@@ -59,9 +59,15 @@ class TimeSeriesMap(NWBContainerMapper):
         if tstamps_builder is None:
             return None
         if isinstance(tstamps_builder, LinkBuilder):
-            if tstamps_builder.builder.parent is None:
-                return
-            target = tstamps_builder.builder
-            return manager.construct(target.parent)
+            # if the parent of our target is available, return the parent object
+            # Otherwise, return the dataset in the target builder
+            #
+            # NOTE: it is not available when data is externally linked
+            # and we haven't explicitly read that file
+            if tstamps_builder.builder.parent is not None:
+                target = tstamps_builder.builder
+                return manager.construct(target.parent)
+            else:
+                return tstamps_builder.builder.data
         else:
             return tstamps_builder.data
--- a/src/pynwb/io/base.py
+++ b/src/pynwb/io/base.py
@@ -59,6 +59,8 @@ class TimeSeriesMap(NWBContainerMapper):
         if tstamps_builder is None:
             return None
         if isinstance(tstamps_builder, LinkBuilder):
+            if tstamps_builder.builder.parent is None:
+                return
             target = tstamps_builder.builder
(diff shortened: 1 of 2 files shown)


NileGraddis:
getting these weird failures to remove open files on windows. I will have to squash ...

bendichter:
Yeah appveyor has been acting funky for a while now

NileGraddis:
@ajtritt @bendichter Codecov is complaining, but I have added an integration test so I think this is ready to go.

The appveyor weirdness points to a problem with cross-file linking generally. Namely: the referencing container keeps a handle to the referenced file. This makes it tricky to ensure that the h5 file is actually closed. Even del-ing the container (and running gc.collect) doesn't give up the file handle.

oruebel:
@NileGraddis I believe codecove may count only coverage from unit tests

@ajtritt can you have a look. I think this PR is fine.

src/pynwb/io/base.py
```diff
@@ -59,6 +59,8 @@ def timestamps_carg(self, builder, manager):
         if tstamps_builder is None:
             return None
         if isinstance(tstamps_builder, LinkBuilder):
+            if tstamps_builder.builder.parent is None:
```
ajtritt: Why does this conditional get satisfied?

NileGraddis: I think that this info comes ultimately from `__read_group` and `__read_dataset` on `HDF5IO`.  `__read_group` does not explicitly endow the `LinkBuilder` with a parent and `__read_dataset` does not set one on the builder it returns.

This seems like OK behavior, but I'm not familiar enough with the intended design to tell if it is a bug. It is turning up problems in timestamps custom constructor arg but not elsewhere (though the parent is still None).

ajtritt: The only builder that does not have a parent should be a top-level builder i.e. the NWBFile group builder.

NileGraddis: I suppose it's not totally clear to me why this is true for linked datasets - their parent lives in potentially another file, and might not be itself linked (this seems like a special case of the general expectation that links break the tree structure). What needs to be done with that parent?

I obviously know a lot less about the intended behavior here than you do. Could you explain more about why this behavior seems wrong and what I ought to do in order to fix it?

ajtritt (approved):

PR #906, opened 2019-05-02
add parallel I/O tutorial as a code block because this won't pass CI

### LINKED ISSUE #900: create tutorial on parallel IO
depends on #847

bendichter:
## Motivation

fix #900 

Code has been tested on a server with MPI enabled. I'm not sure how to build this into the CI. @jcfr, how hard would this be?

image

## Checklist

### DIFF
--- a/docs/gallery/general/advanced_hdf5_io.py
+++ b/docs/gallery/general/advanced_hdf5_io.py
@@ -263,7 +263,6 @@ for k, v in nwbfile.acquisition.items():
 #
 #   # read from dataset in parallel
 #   with NWBHDF5IO(fname, 'r', comm=MPI.COMM_WORLD) as io:
-#       print(io.read().acquisition['ts_name'].data[:])
 #       print(io.read().acquisition['ts_name'].data[rank])
 
 ####################
--- a/docs/gallery/general/advanced_hdf5_io.py
+++ b/docs/gallery/general/advanced_hdf5_io.py
@@ -221,8 +221,8 @@ for k, v in nwbfile.acquisition.items():
 # -----------
 #
 # The HDF5 storage backend supports parallel I/O using the Message Passing Interface (MPI).
-# Using this feature requires that you install `hdf5` and `h5py` against an MPI driver, and you
-# install `mpi4py`. The basic installation of pynwb will not work. Setup can be tricky, and
+# Using this feature requires that you install ``hdf5`` and ``h5py`` against an MPI driver, and you
+# install ``mpi4py``. The basic installation of pynwb will not work. Setup can be tricky, and
 # is outside the scope of this tutorial (for now), and the following assumes that you have
 # HDF5 installed in a MPI configuration. Here we:
 #
@@ -250,7 +250,8 @@ for k, v in nwbfile.acquisition.items():
 #       nwbfile = NWBFile('aa', 'aa', start_time)
 #       data = DataChunkIterator(data=None, maxshape=(4,), dtype=np.dtype('int'))
 #
-#       nwbfile.add_acquisition(TimeSeries('ts_name', description='desc', data=data, rate=100., unit='m'))
+#       nwbfile.add_acquisition(TimeSeries('ts_name', description='desc', data=data,
+#                                          rate=100., unit='m'))
 #       with NWBHDF5IO(fname, 'w') as io:
 #           io.write(nwbfile)
 #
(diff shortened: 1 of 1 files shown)


docs/gallery/general/advanced_hdf5_io.py
```diff
@@ -216,6 +216,56 @@
...
+#   from dateutil import tz
+#   from pynwb import NWBHDF5IO, NWBFile, TimeSeries
+#   from datetime import datetime
+#   from hdmf.data_utils import DataChunkIterator
+#
+#   start_time = datetime(2018, 4, 25, 2, 30, 3, tzinfo=tz.gettz('US/Pacific'))
+#   fname = 'test_parallel_pynwb.nwb'
+#   rank = MPI.COMM_WORLD.rank  # The process ID (integer 0-3 for 4-process run)
+#
+#   # instantiate dataset, but do not write any data
```
oruebel: Instead of ``instantiate dataset, but do not write any data`` maybe say something like.
``Create file on one rank. Here we only instantiate the dataset we want to write in parallel but we do not write any data``

docs/gallery/general/advanced_hdf5_io.py
```diff
@@ -216,6 +216,56 @@
...
+#   if not rank:
```
oruebel: Instead ``if not rank`` I think it may be more intuitive to say ``if rank == 0``

docs/gallery/general/advanced_hdf5_io.py
```diff
@@ -216,6 +216,56 @@
...
+#       nwbfile = NWBFile('aa', 'aa', start_time)
+#       data = DataChunkIterator(data=None, maxshape=(4,), dtype=np.dtype('int'))
+#
```
oruebel, line 256: I think it might be worth adding a comment somewhere along the lines of: "To specify details about chunking, compression and other HDF5-specific I/O options, we can wrap ``data`` via ``H5DataIO``"

docs/gallery/general/advanced_hdf5_io.py
```diff
@@ -216,6 +216,56 @@
...
+#   # write to dataset in parallel
+#   with NWBHDF5IO(fname, 'a', comm=MPI.COMM_WORLD) as io:
+#       nwbfile = io.read()
+#       print(rank)
+#       nwbfile.acquisition['ts_name'].data[rank] = rank
+#
+#   # read from dataset in parallel
+#   with NWBHDF5IO(fname, 'r', comm=MPI.COMM_WORLD) as io:
+#       print(io.read().acquisition['ts_name'].data[:])
```
oruebel: Similar to the parallel write section, I think we should do ``nwbfile = io.read()`` first. Otherwise we end up parsing the file twice here.

docs/gallery/general/advanced_hdf5_io.py
```diff
@@ -216,6 +216,56 @@
...
+# install ``mpi4py``. The basic installation of pynwb will not work. Setup can be tricky, and
+# is outside the scope of this tutorial (for now), and the following assumes that you have
+# HDF5 installed in a MPI configuration. Here we:
+#
+# 1. Instantiate a dataset for parallel write
+#
+# 2. Write to that file in parallel using MPI
+#
+# 3. Read from the file in parallel using MPI.
+#
```
oruebel, line 237: I think it would help to expand the description of the operations a bit:

1. **Instantiate a dataset for parallel write:** We create TimeSeries with 4 timestamps that we will write in parallel
2. **Write to that file in parallel using MPI:** Here we assume 4 MPI ranks  while each rank writes the data for a different timestamp. 
3. **Read from the file in parallel using MPI:** Here each of the 4 MPI ranks reads one timestep from the file

docs/gallery/general/advanced_hdf5_io.py
oruebel: I think we should remove ``print(io.read().acquisition['ts_name'].data[:])`` asking each rank to read the full dataset is not really good practice, because if you really want all ranks to have all the, then you would usually read the data on rank 0 and broadcast with MPI as broadcasting over the interconnect is usually faster than hitting the I/O subsystem with full reads from each rank.

oruebel (dismissed):
Looks good. I added a few minor suggestions to clarify a few things but nothing major.

docs/gallery/general/advanced_hdf5_io.py
bendichter:
```suggestion
#   # Create file on one rank. Here we only instantiate the dataset we want to write in parallel but we do not write any data
```

docs/gallery/general/advanced_hdf5_io.py
bendichter:
```suggestion
#   if rank == 0:
```

docs/gallery/general/advanced_hdf5_io.py
bendichter:
```suggestion
```

oruebel (approved):

oruebel:
Thanks for adding this tutorial! 👍

PR #922, opened 2019-05-09
Clean up tests to remove errors and warnings

rly:
Just cleaning up while I modify other tests. A few deprecated warnings still remain.

### DIFF
--- a/tests/unit/pynwb_tests/test_base.py
+++ b/tests/unit/pynwb_tests/test_base.py
@@ -15,14 +15,14 @@ class TestProcessingModule(unittest.TestCase):
         self.assertEqual(self.pm.name, 'test_procmod')
         self.assertEqual(self.pm.description, 'a fake processing module')
 
-    def test_add_container(self):
+    def test_add_data_interface(self):
         ts = TimeSeries('test_ts', [0, 1, 2, 3, 4, 5],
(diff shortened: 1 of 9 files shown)


t-b:
@rly Very nice. Also the commits are pretty neat organized.

tests/unit/pynwb_tests/test_base.py
```diff
@@ -18,21 +18,21 @@ def test_init(self):
     def test_add_container(self):
         ts = TimeSeries('test_ts', [0, 1, 2, 3, 4, 5],
                         'grams', timestamps=[0.0, 0.1, 0.2, 0.3, 0.4, 0.5])
-        self.pm.add_container(ts)
+        self.pm.add_data_interface(ts)
```
oruebel, line 21: Should this test be renamed. The test is called ``test_add_container`` but it now tests ``add_data_interface`` instead. The same applies to the test_get_container case below.

rly:
As an aside, at what point do we remove deprecated functionality like `add_container`, `get_container`, and `ClusterWaveforms`?

oruebel (approved):

oruebel:
> As an aside, at what point do we remove deprecated functionality like `add_container`, `get_container`, and `ClusterWaveforms`?

In terms of versioning, that would be for 1.x rather than a 1.0.x. For now, I would make an issue and assign it to the NWB 2.x milestone.

Issue #921, opened 2019-05-09
Incompatible pinned requirements


tjd2002:
Both requirements.txt and requirements-dev.txt include 
requests=2.20.0
urllib=1.25.2

This leads to the following conflict:

```
$ pip install -r requirements.txt -r requirements-dev.txt
[...]
ERROR: requests 2.20.0 has requirement urllib3<1.25,>=1.21.1, but you'll have urllib3 1.25.2 which is incompatible.
```

Pip actually completes the install, but we should choose compatible pinned versions. Also, should add a call to `pip check` to CI so that we capture problems like this.

bendichter:
Thanks for the report, Tom. urllib3 was updated for security reasons. the requests library should support the new version of urllib3. They are aware of the issue and planning on rolling out an upgrade for urllib3 1.25 soon (https://github.com/kennethreitz/requests/issues/5065). I think the best move is to wait for them.

t-b:
How about we say that `-dev` is an addition to the normal requirements file only? In that way there is no duplication.

bendichter:
@t-b seems reasonable. @jcfr, thoughts?

tjd2002:
I think we should add `pip check` to the tests to catch this situation in future.  pip is quite happy to install conflicting packages (it reports the error, but then returns successfully).

tjd2002:
@t-b sure--I believe that's what is currently done with the requirements-doc.txt file

bendichter:
@tjd2002 yeah `pip check` sounds like a good idea

tjd2002:
@bendichter upon reading your link more closely: it looks like the security fix was backported to urllib3==1.24.3, so I think that's what we should use until requests updates. I'll push a PR for that.

I'll also add pip check in a separate PR

tjd2002:
Actually @t-b, adding in `-r requirements.txt` to the requirements-dev.txt file makes it so you can't pass that file into conda to create a new environment (as I like to do instead of using pip). 

Also, the instructions for generating the requirements files currently recommend a nice method of just using pip freeze to capture all the pinned requirements.

Neither pip nor conda complains if there are exact duplicate lines across these files, so I think it's harmless to leave it with duplicate entries across these two files (at least for requirements-dev.txt. requirements-doc.txt does not have pinned versions, and looks like it has been manually managed, so I won't touch that for now).

jcfr:
### suggested next step

[...]

PR #925, opened 2019-05-10
Catch dependency conflicts in testing (and update docs)

tjd2002:
Per discussion at #921.

It is currently possible for our pinned requirements list to specify a set of packages that are in conflict with each other, and for tests to succeed. (This is because pip install does not fail in this case, see: pypa/pip#775 pypa/pip#988 )

Indeed, right now we have such a conflict (until #923 is merged). So adding this test should cause CI to (correctly) fail.

The `pip check` command checks for this condition, and fails if there are conflicts. This PR adds pip check to the tox testenv, and also recommends running it when initially writing the requirements files.

I am not sure if I put the command in the right place within tox, but I did test that it detects the conflict.


tjd2002:
Now that #923 is merged, CircleCI is passing as expected 🎉, but Appveyor fails for odd reasons:

Appveyor:py27 fails because virtualenv installs an old version of pip version (8.1.1) that doesn't support `pip check`. Updating pip within appveyor.yml doesn't help as pip seems to be provided by virtualenv.

Appveyor:py36 fails for a strange reason: it is still installing the conflicting urllib3 version (1.25.2) even though 1.24.3 is listed in requirements.txt and requirements-dev.txt.

Appveyor:py37 uses pip 10.0.1, and passes, so this is not a generic Windows problem. 

I am going to try updating **virtualenv** itself within appveyor.yml to see if perhaps both of these problems will go away.

tjd2002:
Hurray! Updating virtualenv within appveyor.yml gives us a more recent pip within the virtualenvs, and also fixes the strange dependency behavior I saw in Appveyor:py36.

dorukozturk (approved):

PR #930, opened 2019-05-13
remove num_samples from timeseries

### LINKED ISSUE #929: remove num_samples
the num_samples field on TimeSeries is not in the spec:
https://github.com/NeurodataWithoutBorders/pynwb/blob/f236738e46b541162d2ff3df936f71a6919ec648/src/pynwb/data/nwb.base.yaml#L104-L233

but it is set on TimeSeries:
https://github.com/NeurodataWithoutBorders/pynwb/blob/f236738e46b541162d2ff3df936f71a6919ec648/src/pynwb/base.py#L140
with a bunch of custom logic.

per conversation with @ajtritt I will remove this field.

NileGraddis:
fix #929


bendichter:
When would you come up against an unreadable warning?

NileGraddis:
@bendichter  Top-level attributes may define `__len__` by dispatch to an attribute of their own, which might not have a `__len__`.

EDIT: example: an NWBData whose __data doesn't define `__len__`, which I think is the case in one of the tests.

ajtritt (approved):

PR #937, opened 2019-05-15
add option to exclude columns from DynamicTable.to_dataframe()

### LINKED ISSUE #936: TimeIntervals.to_dataframe is a bit slow
This is due to the timeseries column. As long as we don't read that column things are real snappy: 
```
def fast_to_df(time_intervals):
    import pandas as pd
    table = pd.DataFrame({
        col.name: col.data for col in self.nwbfile.epochs.columns 
        if col.name not in set(['timeseries', 'timeseries_index'])
    }, index=pd.Index(name='id', data=self.nwbfile.epochs.id.data))
    table.index = table.index.astype(int)

    return table[sorted(table.columns)]
```
This works for everything that I want to do with time intervals but it is obviously dropping information.  One option would be to have on TimeIntervals:
```
def to_dataframe(self, drop_timeseries=False):
```
or 
```
def to_dataframe(self, timeseries_format='object'):
```
where timeseries_format can be:
- object: default behavior
- name : just report the name of the series
- None : drop the series

NileGraddis:
Fix #936

### DIFF
--- a/src/pynwb/core.py
+++ b/src/pynwb/core.py
@@ -1251,13 +1251,16 @@ class DynamicTable(NWBDataInterface):
             return self[key]
         return default
 
-    def to_dataframe(self):
+    def to_dataframe(self, exclude=set([])):
         '''Produce a pandas DataFrame containing this table's data.
         '''
 
         data = {}
         for name in self.colnames:
+            if name in exclude:
+                continue
             col = self.__df_cols[self.__colids[name]]
+
             if isinstance(col.data, (Dataset, np.ndarray)) and col.data.ndim > 1:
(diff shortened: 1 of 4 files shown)


oruebel (commented):
The changes look good. Can you please fix the flake8 errors before you merge:

```
tests/integration/ui_write/test_nwbfile.py:277:1: W293 blank line contains whitespace
tests/integration/ui_write/test_nwbfile.py:278:5: E303 too many blank lines (2)
tests/integration/ui_write/test_nwbfile.py:288:33: E127 continuation line over-indented for visual indent
```

PR #944, opened 2019-05-16
Allow to validate files against the cached spec

### LINKED ISSUE #935: Make validation tool default to using the cached specification if present
From private conversion with @ajtritt:

- [ ] The validation tool at https://github.com/NeurodataWithoutBorders/pynwb/blob/85ef17dc5d820deeaf1c40e5ed22e22336b691ec/src/pynwb/__init__.py#L174 does currently not use the cached specification but the one which ships with pynwb.

- [ ] From reading the code I would also say that present extensions in the file are not read and used for validation.

t-b:
With these changes the cached specification in NWB files can be used for validation.

@rly Is there a good way to enforce a specific hdfm version?

Close #935.


oruebel (dismissed):

oruebel (dismissed):

src/pynwb/validate.py
```diff
@@ -2,10 +2,15 @@
...
 from argparse import ArgumentParser
 
-from pynwb import validate, load_namespaces, get_manager, NWBHDF5IO
+from hdmf.spec import NamespaceCatalog
+from hdmf.build import BuildManager
+from hdmf.build import TypeMap as TypeMap
+from hdmf.backends.hdf5 import HDF5IO
+
+from pynwb import validate, available_namespaces, NWBHDF5IO
+from .spec import NWBDatasetSpec, NWBGroupSpec, NWBNamespace
```
ukos-git:
```console
ModuleNotFoundError: No module named '__main__.spec'; '__main__' is not a package
```
this should probably be `pynwb.spec`

t-b: @ukos-git How did you call the Validation script? I never had that with `python -m pynwb.validate ..`.

ukos-git: `python validate.py example.nwb`

t-b: I'll look into it.

t-b:
@rly Ready from my POV.

oruebel (approved):

PR #953, opened 2019-05-22
Remove SpecFile type from schema and the container class

oruebel:
## Motivation

HDMF/PyNWB perform the caching of the schema on the backend. This is because caching the schema will likely be different for different backends. However, the schema had not been updated to remove the unused ``SpecFile`` type and ``/general/specifications`` group. This lead to problems with MatNWB which relies on the schema. This PR updates the schema accordingly.

See also https://github.com/NeurodataWithoutBorders/matnwb/issues/124 for the related MatNWB issue.

See also https://github.com/NeurodataWithoutBorders/nwb-schema/pull/263 for the update of the storage docs to describe the caching of the spec.

## Checklist


bendichter (dismissed):

src/pynwb/data/nwb.namespace.yaml
```diff
@@ -46,4 +46,4 @@ namespaces:
   - source: nwb.retinotopy.yaml
     title: Retinotopy
     doc: This source module contains neurodata_type for retinotopy data.
-  version: 2.0.1
+  version: 2.0.2
```
t-b, line 53: @oruebel Is that the version written to the NWB file?

Using latest master I get

```sh
$ h5dump icephys_example.nwb  | grep -A30 version | head -n11
   ATTRIBUTE "nwb_version" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_UTF8;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SCALAR
      DATA {
      (0): "2.0b"
      }

thomas@thomas-win7-x64 MINGW64 /e/projekte/pynwb (master)
$ git nn
8f472d6c (/#936 fast epoch to df (#937), 2019-05-16)
```

oruebel: Good catch. Looks like the ``nwb_version`` attribute is currently hard-coded in the specification of nwb-file. I've updated the schema accordingly for now, but it should really be populated from the version indicated in the namespace yaml. Can you file a corresponding issue for this. A separate (but related) issue is that we should also store the version of the namespace with the namespace attribute for the types.

t-b: Will do.

bendichter (approved):

ajtritt (approved):

PR #959, opened 2019-05-28
fix #958 intervals group io error

bendichter:
* rename from time_intervals to intervals
* add test for equality for DynamicTable objects using pandas
* add test for adding custom intervals object
* adjust string printing check

## How to test the behavior?
```python
nwbfile = NWBFile('aa', 'aa', datetime.now().astimezone())
df_words = pd.DataFrame({'start_time': [.1, 2.], 'stop_time': [.8, 2.3], 'label': ['hello', 'there']})
words = TimeIntervals.from_dataframe(df_words, name='words')
nwbfile.add_time_intervals(words)
nwbfile.intervals['words'] == words
```

## Checklist

- [x] Have you checked our [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [x] Have you ensured the PR description clearly describes problem and the solution?
- [x] Is your contribution compliant with our coding style ? This can be checked running `flake8` from the source directory.
- [x] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/NeurodataWithoutBorders/pynwb/pulls) for the same change?
- [x] Have you included the relevant issue number using `#XXX` notation where `XXX` is the issue number ?

PR #961, opened 2019-05-28
fix #941: clean up API

### LINKED ISSUE #941: Improve docs and api names for processing modules and other containers
## 2) Feature Request

"What is a processing module?"
"What is a data interface?"

These are terms that some users at the hackathon were confused and intimidated by. Either we should explain them more clearly and earlier in the tutorials or rename them to something more approachable to neuroscientists. For example, representing processing modules as folders is pretty intuitive and showing a pictoral representation of an `NWBFile` being a container (or folder) with properties and nested containers would be helpful. 

Also I think renaming `nwbfile.module['my_module_name'].add_data_interface(my_elec_series)` to `nwbfile.processing['my_module_name'].add(my_elec_series)` would be a lot easier for new users to understand.

Relatedly, in the documentation, I think showing multiple options for creating and adding a new field to a container is confusing, e.g. `addProcessingModule` and `createProcessingModule`. Same for getting data, e.g. `nwbfile.module[name]` and `nwbfile.getProcessingModule(name)` (it is also confusing that the naming in this example breaks the convention of `nwbfile.field` == `nwbfile.getField`)

bendichter:
* change ProcessingModule.add_data_interface() to .add()
* change ProcessingModule.get_data_interface() to .get()
* change nwbfile.modules to nwbfile.processing

## Motivation

syntax is now: `nwbfile.processing['my_module_name'].add(my_elec_series)`

## Checklist

### DIFF
--- a/docs/gallery/general/file.py
+++ b/docs/gallery/general/file.py
@@ -273,7 +273,7 @@ nwbfile.add_processing_module(ecephys_module)
 # These are: 'behavior', 'ecephys', 'icephys', 'ophys', 'ogen', 'retinotopy', and 'misc'. You may also create
 # a processing module with a custom name. Once these processing modules are added, access them with
 
-nwbfile.modules
+nwbfile.processing
 
 ####################
 # which returns a `dict`:
@@ -291,7 +291,7 @@ nwbfile.modules
 #
 # :py:class:`~pynwb.core.NWBDataInterface` objects can be added to the behavior :ref:`ProcessingModule <basic_procmod>`.
 
-nwbfile.modules['behavior'].add_data_interface(position)
+nwbfile.processing['behavior'].add(position)
 
 ####################
 # .. _basic_epochs:
@@ -419,7 +419,7 @@ with NWBHDF5IO('example_file_path.nwb', 'w') as io:
 
 io = NWBHDF5IO('example_file_path.nwb', mode='a')
 nwbfile = io.read()
-position = nwbfile.modules['behavior'].data_interfaces['Position']
+position = nwbfile.processings['behavior'].data_interfaces['Position']
 
 ####################
 # Next, add a new :py:class:`~pynwb.behavior.SpatialSeries`.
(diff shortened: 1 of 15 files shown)


bendichter:
now that I'm thinking about it maybe we should smoothly deprecate the old way of calling things

rly:
> we should smoothly deprecate the old way of calling things

Yes, I was actually going to comment that for #959 as well.

docs/gallery/general/file.py
```diff
@@ -419,7 +419,7 @@
 
 io = NWBHDF5IO('example_file_path.nwb', mode='a')
 nwbfile = io.read()
-position = nwbfile.modules['behavior'].data_interfaces['Position']
+position = nwbfile.processings['behavior'].data_interfaces['Position']
```
rly: this should be `.processing`

src/pynwb/file.py
```diff
@@ -109,7 +109,7 @@ class NWBFile(MultiContainerInterface):
             'get': 'get_stimulus_template'
         },
         {
-            'attr': 'modules',
```
rly, line 112: Does `nwb.modules['module_name']` still work?

src/pynwb/base.py
```diff
@@ -54,7 +54,7 @@ def add_container(self, **kwargs):
         '''
         container = getargs('container', kwargs)
         warn(PendingDeprecationWarning('add_container will be replaced by add_data_interface'))
```
rly: Change warning to `add_container will be replaced by add`

bendichter:
```suggestion
        warn(PendingDeprecationWarning('add_container will be replaced by add'))
```

docs/gallery/general/file.py
bendichter:
```suggestion
position = nwbfile.processing['behavior'].data_interfaces['Position']
```

src/pynwb/file.py
bendichter: Yes it still works. I just added a test for it

rly (approved):

PR #822, opened 2019-02-18
add UnitSeries

bendichter:
## Motivation

Part of plan to make spike waveforms accessible from Units table. See https://github.com/NeurodataWithoutBorders/nwb-schema/issues/248

## How to test the behavior?
```
from pynwb.misc import Units, UnitSeries

units = Units()
us = UnitSeries('test_unit_series', np.array([0, 0, 1, 1], dtype=int), timestamps=[.1, .2, .3, .4], units=units, description='description')
```

## Checklist

- [x] Have you checked our [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [x] Have you ensured the PR description clearly describes problem and the solution?
- [x] Is your contribution compliant with our coding style ? This can be checked running `flake8` from the source directory.
- [x] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/NeurodataWithoutBorders/pynwb/pulls) for the same change?
- [x] Have you included the relevant issue number using `#XXX` notation where `XXX` is the issue number ?

PR #962, opened 2019-05-28
Fix type hierarchy for CorrectedImageStack, ImagingRetinotopy; improve docs

### LINKED ISSUE #271: Make CorrectedImageStack and ImagingRetinotopy inherit from NWBDataInterface
Looking at the hierarchy, it looks like `CorrectedImageStack` and `ImagingRetinotopy` are data types, rather than metadata types, so they should inherit from `NWBContainer>NWBDataInterface` rather than just `NWBContainer`. This is reflected in pynwb (see below) but not the YAML files.

pynwb:
```python
class CorrectedImageStack(NWBDataInterface):
```
and 
```python
class ImagingRetinotopy(NWBDataInterface):
```

rly:
Fixes https://github.com/NeurodataWithoutBorders/nwb-schema/issues/271 and improve documentation

## Checklist


oruebel (changes requested):
You should also update the patch version of the namespace, i.e., 2.0.x to 2.0.(x+1) since your are changing the schema. 

In https://github.com/NeurodataWithoutBorders/nwb-schema/issues/271 you should then also update the release of the nwb-schema docs. 

Luckily, NWBDataInterface does not any fields so this change should not affect compatability.

rly:
Regarding updated author list, see also updates to nwb-schema: https://github.com/NeurodataWithoutBorders/nwb-schema/pull/265 and https://github.com/NeurodataWithoutBorders/nwb-schema/pull/266

oruebel (dismissed):
I forgot #953 also changes the version, i.e., depending on the order these get merged we may need to update the version again. 

Also t-b discovered in #953  another issue with the version in that we need to update the value of the ``nwb_version`` attribute in nwb.file.yaml as well. While this is a separate issue that we should fix, we should in the meantime make sure to update both until we have fixed the problem with having ``nwb_version`` as attribute in the schema.

rly:
Can we put all of these schema changes under 2.0.2 or should they be kept separate?

rly:
#822 and #825 seem ready to merge with an update to the schema as well.

oruebel:
> Can we put all of these schema changes under 2.0.2 or should they be kept separate?

Putting #953 and this PR under the 2.0.2 umbrella seems fine. #822 and #825 add new functionality and should be part of a separate version jump. 

> #822 and #825 seem ready to merge with an update to the schema as well.

I remember we had some discussions about those PRs, which is why we did not merge those yet, but I don't recall what the issue was. @ajtritt do you recall if #822 and #825 were good to merge now?

ajtritt (approved):

PR #968, opened 2019-05-30
WIP: Support IPNWB nwb files better

t-b:
- [ ] Fix "expected `text` got `ascii`" validation for old NWBv2 files, requires 

Branch: https://github.com/t-b/hdmf-common-schema/tree/use-text-as-encoding-for-dynamic-table-column-names.
Issue: https://github.com/hdmf-dev/hdmf-common-schema/issues/4


t-b:
@ajtritt The commit c66d03f reverses what you did earlier on. Do you remember why you did that? Is there a way to tell the validator that ascii is a valid dtype if text is expected?

oruebel:
>  Is there a way to tell the validator that ascii is a valid dtype if text is expected?

I think that would be in HDMF in the ``check_type`` function of ``validator.py`` https://github.com/hdmf-dev/hdmf/blob/5a864ef997f000d3a8137736ddfc73de33683e3f/src/hdmf/validate/validator.py#L44

oruebel:
What is the status of this PR?

t-b:
@oruebel Currently on the backlog. You can close it for now if you like.

t-b:
Fixed via https://github.com/hdmf-dev/hdmf-common-schema/pull/5.

PR #971, opened 2019-06-02
WIP: Documentation: Add script and code for extracting the specification from NWB files

t-b:


docs/source/extracting-specification.py
```diff
@@ -0,0 +1,60 @@
...
+        spec = nwb[nwb.attrs['.specloc']]
+        for ns in spec.values():
+            for version in ns.values():
+                for sub in version.values():
+                    contents = json.loads(sub[()])
+
+                    def getName(o):
+                        return os.path.basename(o.name)
+
+                    filename = "{}_{}_{}_{}.json".format(getName(spec), getName(ns),
```
oruebel: I think it would be good to use the SpecWrite from HDMF to do the write:

https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/spec/write.py#L28

docs/source/extracting-specification.py
```diff
@@ -0,0 +1,60 @@
...
+
+    filename = 'specification-extraction.nwb'
+
+    with NWBHDF5IO(filename, 'w') as nwbio:
+        nwbio.write(nwbfile, cache_spec=True)
+
+    return filename
+
+
+def exportSpecification(filename):
```
oruebel: Rather than making this part of the docs, I think it would be useful to have a function either in the HDF5IO backend (or maybe the [YAMLSpecReader ](https://github.com/hdmf-dev/hdmf/blob/8ddc8574123853215a4d6c43d27bd2aec908017d/src/hdmf/spec/namespace.py#L184) to load the specification(s)), which I guess would return either a ``NamespaceCatalog`` in case you load all namespaces of a ``SpecNamespace`` if you load a particular namespace from the file.

This would have the advantage of making the functionality more reusable, part of the regular tests, and also greatly simplify the tutorial.

docs/source/extracting-specification.py
```diff
@@ -0,0 +1,60 @@
...
+                    contents = json.loads(sub[()])
+
+                    def getName(o):
+                        return os.path.basename(o.name)
+
+                    filename = "{}_{}_{}_{}.json".format(getName(spec), getName(ns),
+                                                         getName(version), getName(sub))
+                    with io.open(filename, mode="w", newline="\n") as f:
+                        json.dump(contents, fp=f, sort_keys=True,
+                                  ensure_ascii=False, indent=2)
```
ukos-git: can we remove `indent` and add `separators=(',',':')` to have compact layout?

docs/source/extracting-specification.py
t-b: @oruebel Hmmpf. I'm not very inclined to add something I use at most once year into hdmf with tests.

How about adding this script, and the forgotten one in `docs/code` to the example tests? I did that in 3b152bbb (run_example_tests(): Test the scripts in docs/code as well, 2019-06-11).

oruebel:
What is the status of this PR?

t-b:
@oruebel Currently on the backlog. You can close it for now if you like.

PR #978, opened 2019-06-18
Reorder keys in YAML files, add dims next to shape, add line spacing

### LINKED ISSUE #274: Reorder YAML keys to be consistent and more human-readable
The ordering of keys for many of the types in the YAML files could be improved to be consistent across definitions/files and more human readable. 

For example, if `neurodata_type_def` is present, it should always be first; else, `name` should be first. `dims` should be next to `shape`. `default_value` should be next to `value`. Description of the size/shape/quantity of the dataset should go before any attributes, groups, datasets, or links contained within it. 

See also https://github.com/NeurodataWithoutBorders/nwb-schema/issues/139 where the order: ```['neurodata_type_def', 'neurodata_type_inc', 'name', 'dtype', 'doc']``` was decided on.

I propose this more full ordering below. As was discussed previously, this ordering should not be enforced, but we should try to follow these guidelines when modifying the core and in the script that generates extensions. 

I also propose having a blank line before each new type definition `neurodata_type_def`

```
neurodata_type_def
neurodata_type_inc
name
default_name
dtype
target_type
dims
shape
value
default_value
doc
required
quantity
attributes
datasets
groups
links
```

rly:
Fixes https://github.com/NeurodataWithoutBorders/nwb-schema/issues/274 . This does not functionally change the content of the schema.

1. Ordering of keys in YAML files is now:
```yaml
neurodata_type_def
neurodata_type_inc
name
default_name
dtype
target_type
dims
shape
value
default_value
doc
required
quantity
attributes
datasets
groups
links
```
except where `neurodata_type_def` is not used, in which case `name` goes before `neurodata_type_inc`.

2. `ElementIdentifiers`, `Image`, `DynamicTable.colnames`, `DynamicTable.id`, `NWBFile.keywords`, `Units.waveform_mean`, and `Units.waveform_sd` now have the attribute `dims` that should accompany `shape`.

3. Added line break before `neurodata_type_def` to make them more visible/readable.


bendichter:
Looks good to me. Can you change the output order accordingly in HDMF? 

https://github.com/hdmf-dev/hdmf/blob/8b1b595d94e641d7105ad12b7a97ec2420265c4a/src/hdmf/spec/write.py#L72

bendichter (dismissed):

ajtritt (approved):

PR #980, opened 2019-06-20
Fix broken tests to work with latest hdmf

rly:
IO tests broke after the latest updates to HDMF where:
- an UnsupportedOperation is thrown if an IO object is opened in w- mode and the file already exists
- you cannot change the parent of a Container after it has been set, so an electrode table region cannot be re-used.

### DIFF
--- a/tests/integration/test_io.py
+++ b/tests/integration/test_io.py
@@ -1,12 +1,13 @@
 import unittest2 as unittest
-import six
 from datetime import datetime
 from dateutil.tz import tzlocal, tzutc
 import os
+import re
 from h5py import File
 
 from pynwb import NWBFile, TimeSeries, get_manager, NWBHDF5IO, validate
 
+from hdmf.backends.io import UnsupportedOperation
 from hdmf.backends.hdf5 import HDF5IO, H5DataIO
 from hdmf.data_utils import DataChunkIterator
 from hdmf.build import GroupBuilder, DatasetBuilder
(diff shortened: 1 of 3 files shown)


rly:
Since this relies on importing `UnsupportedOperation` from HDMF, which exists only in the latest code, this PR should be merged after the next release of HDMF.

ajtritt:
[Issues that I need to review](https://trello.com/c/nrzC7XLX/1-issues-that-i-need-to-review)

tests/integration/test_io.py
```diff
@@ -273,8 +265,8 @@ def test_append(self):
 
         with NWBHDF5IO(self.path, mode='a') as io:
             nwb = io.read()
-            elec = nwb.processing['test_proc_mod']['LFP'].electrical_series['test_device'].electrodes
-            ts2 = ElectricalSeries(name='timeseries2', data=[4., 5., 6.], rate=1.0, electrodes=elec)
+            link_electrodes = nwb.processing['test_proc_mod']['LFP'].electrical_series['test_es'].electrodes
+            ts2 = ElectricalSeries(name='timeseries2', data=[4., 5., 6.], rate=1.0, electrodes=link_electrodes)
             nwb.add_acquisition(ts2)
             io.write(nwb)
```
ajtritt, line 274: Can you add a line that checks (after reading, in the context below) that the two electrodes are the same object?  Something like:
```
self.assertIs(nwb.processing['test_proc_mod']['LFP'].electrical_series['test_es'].electrodes, 
              nwb.acquisition['timeseries2'].electrodes)
```

It would be nice to make sure the linked objects get read back in correctly.

ajtritt (changes requested):
Everything looks good. But, while you're on this code, can you update the test to make sure external links get read correctly

ajtritt (approved):

PR #995, opened 2019-07-02
Improve unit test runner output to make debugging easier

rly:
This changes test.py to output colored, tabular test results when in verbose mode (python test.py -v) which should make figuring out the source of a warning or error easier.

Mirrors update to test.py in https://github.com/hdmf-dev/hdmf/pull/89

screenshot


rly:
Also moves unit/pynwb_tests/* to unit/*

ajtritt (dismissed):

ajtritt (approved):

PR #1008, opened 2019-07-09
change dims for Units waveform

bendichter:
add an optional 3D shape for waveform_mean and waveform_sd including a third dimension for electrodes

## Motivation

You can easily put multi-electrode waveforms in the waveform_mean and waveform_sd columns, but it previously was not clear what the order of those dimensions was. This adds an optional 3D shape to waveform_mean and waveform_sd that puts electrodes in the last dimension

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/src/pynwb/data/nwb.misc.yaml
+++ b/src/pynwb/data/nwb.misc.yaml
@@ -253,21 +253,33 @@ groups:
     neurodata_type_inc: VectorData
     dtype: float
     dims:
-    - num_units
-    - num_samples
+    - - num_units
+      - num_samples
+    - - num_units
+      - num_samples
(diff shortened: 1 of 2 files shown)


rly:
Could you give an example use case for the new electrodes dimension?

bendichter:
It has come up several times that neurons are recorded from more than one electrode. Most recently with neuropixel data

rly:
Ah, thanks. Right. A unit recorded on multiple electrodes has a mean and sd waveform on each electrode, and it makes sense to store those.

rly (approved):

bendichter:
@kristinquick this has been merged into dev

PR #1014, opened 2019-07-17
Update requirements

rly:
I'm preparing for a release. I updated the requirements.txt and requirements-dev.txt to the latest versions. I also removed unnecessary requirements so that there is less to manage, e.g. pycodestyle, pyflakes, and mccabe are required by flake8 and flake8 is the only package  that requires them, so just require flake8 and let pip handle the dependencies.


rly:
Also, I am not sure what certifi, idna, requests, and urllib3 are for. AFAIK, pynwb does not make any http calls... Can they be removed?

t-b:
@rly One of the gallery examples (Allen brain observatory) does download its data. Maybe its from there? c136c829 (urllib2 in setup.py, 2017-08-08) hints a bit in that direction as @nicain did that.

rly:
Thanks for the tip @t-b . It looks like the allensdk package wraps all of the data download machinery, and the use of urllib in setup.py was originally for downloading the schema, which we now keep a local copy of. So it looks like all four of those packages (certifi, idna, requests, and urllib3) can be removed.

rly:
To be clear, I did  a search of the dev branch for any calls to those packages and could not find any.

ajtritt (approved):
thanks for cleaning this up!

PR #1032, opened 2019-08-05
Remove help attribute from schema and pynwb

rly:
Addresses https://github.com/NeurodataWithoutBorders/nwb-schema/issues/270

This PR removes the `help` attribute from all neurodata_types in the schema and PyNWB. Now that the schema is cached, the `help` attribute is redundant, confusing, clutters the file and documentation, and adds substantial boilerplate to writing extensions. In addition, it was often used inconsistently.

Since extra attributes are allowed during validation, this change does not break backwards compatibility on read. Users and extensions can still include a `help` attribute if they would like. 

Minor:
- For some types, the `description` attribute, which was set to a fixed value and served the same purpose as `help` was removed (I suspect these were meant to be given the attribute name `help` instead of `description`).
- For some classes, the docval key `help` was replaced with `doc`, which seemed more appropriate.
- Documentation and tests were updated to reflect these changes


bendichter (approved):

bendichter:
-600 lines of code! woohoo!

oruebel (approved):

PR #1038, opened 2019-08-06
Raise warning if name/notes supplied when adding Container to scratch

rly:
Raise warning if name/notes supplied when adding Container to scratch (they are ignored), and add tests.

@ajtritt `table_description` appears not to be used in `add_scratch`?


tests/integration/ui_write/test_scratch.py
```diff
@@ -21,34 +20,65 @@ def getContainer(self, nwbfile):
...
         data = pd.DataFrame(data={'col1': [1, 2, 3, 4], 'col2': ['a', 'b', 'c', 'd']})
-        ret = self.roundtripScratch(data, 'DataFrame')
+        warn_msg = 'Notes argument is ignored when adding a pandas DataFrame to scratch'
+        with self.assertWarnsRegex(UserWarning, warn_msg):
+            ret = self.roundtripScratch(data, 'DataFrame')
         assert_array_equal(data.values, ret.values)
         assert_array_equal(data.index.values, ret.index.values)
+        self.validate()
+
+    def test_scratch_container(self, validate=True, **kwargs):
```
ajtritt: Can you make this a camel case method, and then have `test_*` method that calls the camel case method? That way `test_scratch_container_with_name` and `test_scratch_container_with_notes` don't depend on another test

src/pynwb/file.py
```diff
@@ -618,15 +618,21 @@ def add_stimulus_template(self, timeseries):
...
     def add_scratch(self, **kwargs):
         '''Add data to the scratch space'''
-        data = getargs('data', kwargs)
+        data, name, notes, table_description = getargs('data', 'name', 'notes', 'table_description', kwargs)
         if isinstance(data, (np.ndarray, pd.DataFrame, list, tuple)):
-            name, notes, table_description = getargs('name', 'notes', 'table_description', kwargs)
             if name is None:
                 raise ValueError('please provide a name for scratch data')
             if isinstance(data, pd.DataFrame):
                 data = DynamicTable.from_dataframe(df=data, name=name)
```
ajtritt: `table_description` should be passed in here. See [DynamicTable.from_dataframe](https://pynwb.readthedocs.io/en/latest/pynwb.core.html?highlight=DynamicTable#pynwb.core.DynamicTable.from_dataframe) for details.

ajtritt (changes requested):
See comments.

tests/integration/ui_write/test_scratch.py
rly: Instead of a camel case method, how about prependign the name with an underscore?

ajtritt: That works just fine

ajtritt (approved):

PR #1049, opened 2019-08-21
Minor refactoring and cleaning of flake8/linter suggestions

rly:
Just a little code and documentation cleanup

### DIFF
--- a/src/pynwb/__init__.py
+++ b/src/pynwb/__init__.py
@@ -223,7 +223,7 @@ class NWBHDF5IO(_HDF5IO):
 from . import io as __io  # noqa: F401,E402
 from .core import NWBContainer, NWBData  # noqa: F401,E402
 from .base import TimeSeries, ProcessingModule  # noqa: F401,E402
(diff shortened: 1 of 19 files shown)


src/pynwb/__init__.py
```diff
@@ -4,28 +4,24 @@
 import os.path
 from copy import deepcopy
 from warnings import warn
-
 import h5py
+from pkg_resources import resource_filename
```
oruebel: This was probably intentionally a function-level import because you don't want to pull it into the pynwb module

rly: Ohhh I see.

rly: This code is always run though (see line 102), so wouldn't it be pulled into the pynwb module anyway?

rly: Never mind. I tested it and you are right. It is not pulled in if it is in the function. Will change.

oruebel (approved):

PR #1050, opened 2019-08-21
[WIP] Pickle TypeMap

ajtritt:
## Motivation

Currently, it takes about 2.5 seconds to PyNWB. Pickling the TypeMap on build/install will help bring this down. 

This is WIP to make sure it works with testing/deployment

@dorukozturk @jcfr Your thoughts/input on this would be greatly appreciated.

## How to test the behavior?
Run test suite

## Checklist

### DIFF
--- a/src/pynwb/__init__.py
+++ b/src/pynwb/__init__.py
@@ -14,11 +14,62 @@ from hdmf.utils import docval, getargs, popargs, call_docval_func  # noqa: E402
 from hdmf.backends.io import HDMFIO  # noqa: E402
 from hdmf.backends.hdf5 import HDF5IO as _HDF5IO  # noqa: E402
 from hdmf.validate import ValidatorMap  # noqa: E402
-from hdmf.build import BuildManager  # noqa: E402
+from hdmf.build import BuildManager, TypeMap  # noqa: E402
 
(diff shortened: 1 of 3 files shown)


src/pynwb/__init__.py
```diff
@@ -14,14 +13,65 @@
...
+def register_class(**kwargs):
+    """Register an NWBContainer class to use for reading and writing a neurodata_type from a specification
+    If container_cls is not specified, returns a decorator for registering an NWBContainer subclass
+    as the class for neurodata_type in namespace.
+    """
+    neurodata_type, namespace, container_cls = getargs('neurodata_type', 'namespace', 'container_cls', kwargs)
+
+    def _dec(cls):
+        __rct_kwargs.append({'data_type': neurodata_type, 'namespace': namespace, 'container_cls': cls})
+        return cls
```
rly, line 34: The above code will not work with extensions that call `register_class`. This will add kwargs to `__rct_kwargs` but they are not registered into the `__TYPE_MAP`. Same for the `register_map` method.

rly (changes requested):

src/pynwb/__init__.py
ajtritt: Good catch... back to the drawing board.

PR #1057, opened 2019-09-02
Better testing

t-b:
Test the code in `docs/code` as well.

### DIFF
--- a/test.py
+++ b/test.py
@@ -51,7 +51,7 @@ def run_example_tests():
     global TOTAL, FAILURES, ERRORS
     logging.info('running example tests')
     examples_scripts = list()
-    for sub in ("gallery"):
+    for sub in ("gallery", "code"):
         for root, dirs, files in os.walk(os.path.join(os.path.dirname(__file__), "docs", sub)):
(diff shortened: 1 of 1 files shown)


rly (approved):

rly:
@t-b where is `docs/code/creating-and-writing-nwbfile.py` referenced? I did a search for the filename and could not find anything. Is it legacy code that can be removed?

t-b:
I thought it was used in sphinx, but turns out that was old knowledge. It is not use anymore since 8a32eb77 (mv examples to extensions, 2018-03-29).

rly:
Thanks for digging that up. I vote to remove that script, since it is redundant with the examples in `docs/gallery/`. It would be one less thing to maintain. And then we would not need this PR.

rly:
I removed the script and directory in https://github.com/NeurodataWithoutBorders/pynwb/pull/1055/commits/69679264abfbd6396cae0b9b95d344b808a40c6d in PR #1055, so this PR is no longer necessary. Thanks for pointing out the dead code, @t-b !

PR #1061, opened 2019-09-10
Enh/rm hdmf common

ajtritt:
## Motivation

Remove HDMF common classes and associated tests from PyNWB


oruebel (dismissed):

src/pynwb/core.py
```diff
@@ -325,82 +114,6 @@ def __init__(self, **kwargs):
...
-class ElementIdentifiers(NWBData):
-
-    @docval({'name': 'name', 'type': str, 'doc': 'the name of this ElementIdentifiers'},
-            {'name': 'data', 'type': ('array_data', 'data'), 'doc': 'a 1D dataset containing identifiers',
-             'default': list()})
-    def __init__(self, **kwargs):
-        call_docval_func(super(ElementIdentifiers, self).__init__, kwargs)
-
-
 class NWBTable(NWBData):
```
rly, line 117: What do NWBTable and NWBTableRegion do? They are not in the core schema. They are not used by DynamicTable or DynamicTableRegion. From the comments, it looks like the ElectrodeTable used to be an NWBTable. Are these classes effectively dead code? If so, can we remove or deprecate their usage in favor of DynamicTable and DynamicTableRegion?

src/pynwb/io/__init__.py
```diff
@@ -1,3 +1,5 @@
+# flake8: noqa: F401
```
rly: This flake8 exception should be handled already by https://github.com/NeurodataWithoutBorders/pynwb/blob/ab493464ddc65df4724b5d82354e31fadc91880d/setup.cfg#L26

rly: I removed it.

src/pynwb/spec.py
```diff
@@ -94,6 +94,26 @@ def neurodata_type_inc(self):
     def neurodata_type_def(self):
         return self.data_type_def
 
+    @classmethod
+    def build_const_args(cls, spec_dict):
```
rly, line 98: Where is this function used?

tests/integration/test_io.py
```diff
@@ -17,6 +17,9 @@
 
 import numpy as np
 
+global CLEAN_NWB
```
rly: Is it necessary to make this a global variable? I see advantages to that. Note that this is effectively redefined in:
pynwb\tests\integration\ui_write\base.py:161 and
pynwb\tests\integration\ui_write\test_modular_storage.py:21

If we keep the global var here, can we use the same var in those places?

src/pynwb/core.py
ajtritt: They are not they’re own type. NWBTable is a class for handling 1D datasets with a compound type, which has table like behavior

rly: Since it seems pretty generic (and not in the core NWB schema), would it make sense to move this into HDMF common? That could be a separate PR.

tests/integration/test_io.py
ajtritt: I moved this to a reusable function in the subpackage pynwb.testing.

rly (approved):

rly:
Looks all good to me!

oruebel (approved):

PR #1072, opened 2019-09-24
add channel_conversion and tests

### LINKED ISSUE #1064: conversion_factor per channel
## Feature Request
We are running into a suboptimal situation storing neuropixel data from a SpikeGLX acquisition system. The system stores a conversion factor *per channel*. We can get this data into NWB by just multiplying the conversion factors across the channels and storing floats in Volts, but then we are dramatically increasing the file size. Neuropixel data is very big, so this is really not an ideal solution. Would there be any convenient way to represent conversion_factor in a per-channel way?

cc @luiztauffer

bendichter:
fix #1064 
depends on https://github.com/NeurodataWithoutBorders/nwb-schema/pull/315


bendichter:
(not passing because it relies on NeurodataWithoutBorders/nwb-schema#315)

rly (approved):

PR #1092, opened 2019-10-09
Expose export_spec from hdmf.spec.write for import from pynwb.spec

rly:
Paired with https://github.com/hdmf-dev/hdmf/pull/170

HDMF now includes a utility function `export_spec` for use by extension developers. To shield the user from needing to know about HDMF, this PR allows a user to import `export_spec` from pynwb.spec instead of from hdmf.spec.write.


oruebel (approved):

rly:
This PR requires an HDMF release.

Issue #1097, opened 2019-10-24
issue writing units table with electrodes defined


bendichter:
Steps to Reproduce

```python
from datetime import datetime
from dateutil.tz import tzlocal
from pynwb import NWBFile, NWBHDF5IO

nwbfile = NWBFile(session_description='session', identifier='1', session_start_time=datetime.now(tzlocal()))
nwbfile.add_unit(electrodes=(0,))

with NWBHDF5IO('test_file.nwb', 'w') as io:
    io.write(nwbfile)
```

```python
/anaconda3/envs/pynwb/bin/python /Users/bendichter/dev/pynwb/scratch_units.py
Traceback (most recent call last):
  File "/Users/bendichter/dev/pynwb/scratch_units.py", line 9, in <module>
    io.write(nwbfile)
  File "/Users/bendichter/dev/hdmf/src/hdmf/utils.py", line 438, in func_call
    return func(self, **parsed['args'])
  File "/Users/bendichter/dev/hdmf/src/hdmf/backends/hdf5/h5tools.py", line 269, in write
    call_docval_func(super(HDF5IO, self).write, kwargs)
  File "/Users/bendichter/dev/hdmf/src/hdmf/utils.py", line 327, in call_docval_func
    return func(*fargs, **fkwargs)
  File "/Users/bendichter/dev/hdmf/src/hdmf/utils.py", line 438, in func_call
    return func(self, **parsed['args'])
  File "/Users/bendichter/dev/hdmf/src/hdmf/backends/io.py", line 43, in write
    f_builder = self.__manager.build(container, source=self.__source)
  File "/Users/bendichter/dev/hdmf/src/hdmf/utils.py", line 438, in func_call
    return func(self, **parsed['args'])
  File "/Users/bendichter/dev/hdmf/src/hdmf/build/map.py", line 166, in build
    result = self.__type_map.build(container, self, source=source, spec_ext=spec_ext)
  File "/Users/bendichter/dev/hdmf/src/hdmf/utils.py", line 438, in func_call
    return func(self, **parsed['args'])
  File "/Users/bendichter/dev/hdmf/src/hdmf/build/map.py", line 1709, in build
    builder = attr_map.build(container, manager, builder=builder, source=source, spec_ext=spec_ext)
  File "/Users/bendichter/dev/hdmf/src/hdmf/utils.py", line 438, in func_call
    return func(self, **parsed['args'])
[...]

PR #1098, opened 2019-10-24
add warning

ajtritt:
## Motivation

add warning when referencing electrode table before it exists

relates to #1097

### DIFF
--- a/src/pynwb/misc.py
+++ b/src/pynwb/misc.py
@@ -186,6 +186,8 @@ class Units(DynamicTable):
                 if self.__electrode_table is None:
                     nwbfile = self.get_ancestor(data_type='NWBFile')
                     elec_col.table = nwbfile.electrodes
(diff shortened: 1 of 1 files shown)


oruebel (dismissed):

src/pynwb/misc.py
```diff
@@ -186,6 +186,8 @@ def add_unit(self, **kwargs):
                 if self.__electrode_table is None:
                     nwbfile = self.get_ancestor(data_type='NWBFile')
                     elec_col.table = nwbfile.electrodes
+                    if elec_col.table is None:
+                        warnings.warn('Reference to electrode table that does not yet exit')
```
bendichter:
```suggestion
                        warnings.warn('Reference to electrode table that does not yet exist')
```

bendichter:
other than that looks good

oruebel (approved):

PR #1104, opened 2019-11-08
Refactor docval calls and improve unit test with implicit args

rly:
Minor improvement. Paired with, but should not require https://github.com/hdmf-dev/hdmf/pull/194

This PR:
1) changes how fixed values were injected into the keyword arguments of a superclass in some cases
2) replaces `fmt_docval_args` and the subsequent superclass constructor call with `call_docval_func`
3) fixes a unit test where the index to a dynamictable was implicitly provided

### DIFF
--- a/src/pynwb/base.py
+++ b/src/pynwb/base.py
@@ -5,7 +5,7 @@ try:
 except ImportError:
     from collections import Iterable  # Python 2.7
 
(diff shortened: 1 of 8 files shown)


oruebel (approved):
Interesting that calling the super constructor inside ``call_docval_func`` rather than directly in ``__init__`` works but I guess since ``super`` takes ``self`` as an argument scoping is not an issue. Looks good to me.

PR #1109, opened 2019-11-21
Fix reference images not being mapped in PlaneSegmentation

### LINKED ISSUE #1108: PlaneSegmentation.reference_images is not mapped
When `DynamicTable` was extracted from PyNWB to HDMF, the `__nwbfields__` field in `DynamicTable` became `__fields__` . `NWBContainer` defines `_fieldsname = '__nwbfields__'`, but `PlaneSegmentation` inherits from `DynamicTable` rather than `NWBContainer` so it does not define `_fieldsname = '__nwbfields__'`; it still has it set to `__fields__`. As a result, `PlaneSegmentation.__nwbfields__` is never read.

The other three classes that inherit from `DynamicTable` do not add fields, just columns. Only `PlaneSegmentation` and extensions have this problem.

```python
class PlaneSegmentation(DynamicTable):
    __nwbfields__ = ('description',
                     'imaging_plane',
                     {'name': 'reference_images', 'child': True})
```

This results in broken links and orphan containers and `reference_images` objects are not being written to disk.

Possible solutions:
1. Stopgap: in `PlaneSegmentation`, change `__nwbfields__` to `__fields__`.
2. Create an `NWBDynamicTable` class that includes `_fieldsname = '__nwbfields__'`. This can be done with the `NWBMixin` class. `DynamicTable` subtypes would inherit from `NWBDynamicTable` instead.

I'll implement 1, but we can discuss 2.

rly:
- Fix #1108 : Change `__nwbfields__` to `__fields__`.
- `DynamicTable` already has `description`, so remove that.
- Add note about double mapping for `NWBFile`. (somewhat unrelated to the above)

### DIFF
--- a/src/pynwb/io/file.py
+++ b/src/pynwb/io/file.py
@@ -12,16 +12,17 @@ class NWBFileMap(ObjectMapper):
         super(NWBFileMap, self).__init__(spec)
 
         acq_spec = self.spec.get_group('acquisition')
-        self.unmap(acq_spec)
         self.map_spec('acquisition', acq_spec.get_neurodata_type('NWBDataInterface'))
         self.map_spec('acquisition', acq_spec.get_neurodata_type('DynamicTable'))
+        # TODO: note that double mapping "acquisition" means __carg2spec and __attr2spec (both unused)
+        # map "acquisition" to the last spec, in this case, DynamicTable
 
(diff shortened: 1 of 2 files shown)


rly:
The round-trip tests should have caught this but didn't. Warnings were raised for orphan containers and broken links though.

TODO: amend the tests

oruebel (approved):

PR #1127, opened 2019-12-29
add json schema and code to run validation over spec files

bendichter:


bendichter:
see hdmf analog here: https://github.com/hdmf-dev/hdmf/pull/223

bendichter:
Now that I have implemented this and used it to clean up our core, I want to take a step back and think about how this tool will be used and how we can best position it to be used and maintained.

I think the .json file itself should be in nwb-schema, versioned alongside the document describing the specification. Since we want nwb-schema to be language-agnostic, I think the code to run the validator should stay in pynwb (I have them both in pynwb for now because it would be a pain to develop otherwise). 

Here are some use-cases I'd like to see for this:
1) CI for nwb-schema changes, protecting our dev branch against changes that do not pass validation
2) CI for extensions, likewise
3) Importable by developers of extensions to use during development
4) Part of the validation pipeline for submitted extensions

Ideally we would have an abstract version of this that works for hdmf and extend from it for nwb. I think that's probably possible but I haven't attempted it.

Note: I have not done a validator for the namespace yaml and the validator currently does not support compound types (though this is not a problem for the current core schema since there aren't any)

PR #1142, opened 2020-01-16
Remove unittest2 dependency

rly:
Python 2 will no longer be supported, so we can remove the dependency and usage of unittest2 and  traceback2. This helps break up the mega PR https://github.com/NeurodataWithoutBorders/pynwb/pull/1117

### DIFF
--- a/tests/unit/test_MultiContainerInterface.py
+++ b/tests/unit/test_MultiContainerInterface.py
@@ -1,4 +1,4 @@
-import unittest as unittest
+import unittest
 
 from pynwb.file import MultiContainerInterface, NWBDataInterface
 from hdmf.utils import docval, get_docval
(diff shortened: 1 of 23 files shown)


tests/unit/test_MultiContainerInterface.py
```diff
@@ -1,4 +1,4 @@
-import unittest2 as unittest
+import unittest as unittest
```
bendichter:
```suggestion
import unittest
```

bendichter: looks good though! Thanks for cleaning up that python 2 clutter!

rly: Thanks! Somehow missed that.

bendichter (approved):

PR #1163, opened 2020-01-29
Add file create data appending

### LINKED ISSUE #990: Update file_create_date on write
## Feature Request

Opening an NWB file in any mode except 'r' should add an entry with the current date and time to `NWBFile.file_create_date`. It would be better if this could be changed only upon modification to the file, but with how h5py works, this does not appear to be trackable. (One could just do `dataset[()] = some new data` without calling `io.write`).

If this is not doable, I suggest 1) changing the documentation and 2) changing the variable type from array of datetimes to a single datetime. 

### Problem/Use Case
`NWBFile.file_create_date` is intended to store a record of the date and time an NWB file was created and all subsequent modifications made. But this is currently set to a list containing one element - the creation date. The list is never subsequently changed.

## Checklist

t-b:
Close #990.

Requires https://github.com/hdmf-dev/hdmf/pull/280.

### DIFF
--- a/tests/unit/test_file.py
+++ b/tests/unit/test_file.py
@@ -1,4 +1,5 @@
 import numpy as np
+import os
 import pandas as pd
 
 from datetime import datetime
@@ -516,3 +517,46 @@ class TestTimestampsRefAware(TestCase):
                     'TEST124',
                     self.start_time,
                     timestamps_reference_time=self.ref_time_notz)
+
+
+class TestFileCreateDateArray(TestCase):
+
+    def setUp(self):
+        self.path = 'unittest_file_create_date.nwb'
+
+    def tearDown(self):
+        if os.path.exists(self.path):
+            os.remove(self.path)
+
+    def test_simple(self):
+        file_create_date = datetime.now(tzlocal())
+        nwbfile_init = NWBFile(' ', ' ',
+                               datetime.now(tzlocal()),
+                               file_create_date=file_create_date,
+                               institution='Rixdorf University, Berlin')
+
+        self.assertEqual(nwbfile_init.file_create_date, [file_create_date])
+        self.assertEqual(len(nwbfile_init.file_create_date), 1)
+
+        with NWBHDF5IO(self.path, 'w') as io:
+            io.write(nwbfile_init)
+
+        with NWBHDF5IO(self.path, 'r') as reader:
+            nwbfile = reader.read()
+
+            # no change as it was opened read-only
+            self.assertEqual(len(nwbfile.file_create_date), 1)
+
+        with NWBHDF5IO(self.path, 'r+') as writer:
+            nwbfile = writer.read()
+
+            # added one more entry as opened read/write
+            self.assertEqual(len(nwbfile.file_create_date), 2)
+
+            writer.write(nwbfile)
+
+        with NWBHDF5IO(self.path, 'r') as reader:
+            nwbfile = reader.read()
+
+            # reopen again to check that it has still two entries
+            self.assertEqual(len(nwbfile.file_create_date), 2)
--- a/src/pynwb/__init__.py
+++ b/src/pynwb/__init__.py
@@ -258,6 +258,15 @@ class NWBHDF5IO(_HDF5IO):
         kwargs['container'] = nwbfile
         call_docval_func(super().export, kwargs)
 
+    def read(self, **kwargs):
+
+        nwbfile = super().read(**kwargs)
+
+        if self.mode != 'r':
+            nwbfile._appendModificationEntry()
+
+        return nwbfile
+
 
 from . import io as __io  # noqa: F401,E402
 from .core import NWBContainer, NWBData  # noqa: F401,E402
--- a/src/pynwb/file.py
+++ b/src/pynwb/file.py
@@ -309,10 +309,12 @@ class NWBFile(MultiContainerInterface):
             raise ValueError("'timestamps_reference_time' must be a timezone-aware datetime object.")
 
         self.fields['file_create_date'] = getargs('file_create_date', kwargs)
+
         if self.fields['file_create_date'] is None:
-            self.fields['file_create_date'] = datetime.now(tzlocal())
-        if isinstance(self.fields['file_create_date'], datetime):
+            self.fields['file_create_date'] = [datetime.now(tzlocal())]
+        elif isinstance(self.fields['file_create_date'], datetime):
             self.fields['file_create_date'] = [self.fields['file_create_date']]
+
         self.fields['file_create_date'] = list(map(_add_missing_timezone, self.fields['file_create_date']))
 
         fieldnames = [
--- a/src/pynwb/file.py
+++ b/src/pynwb/file.py
@@ -751,6 +751,13 @@ class NWBFile(MultiContainerInterface):
 
         return NWBFile(**kwargs)
 
+    def _appendModificationEntry(self):
+        """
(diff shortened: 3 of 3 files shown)


tests/unit/test_file.py
```diff
@@ -468,3 +468,38 @@ def test_reftime_tzaware(self):
...
+            nwbfile = reader.read()
+
+            # no change as it was opened read-only
+            self.assertEqual(len(nwbfile.file_create_date), 1)
+
+        with NWBHDF5IO(self.path, 'r+') as writer:
+            nwbfile = writer.read()
+
+            # added one more entry as opened read/write
+            self.assertEqual(len(nwbfile.file_create_date), 2)
```
rly, line 554: Please also test the second round-trip, i.e., close the file and re-open it in read-mode and confirm that the change to `file_create_date` is still present. I am concerned that the `file_create_date` dataset is not chunked and therefore cannot grow, or the change is not saved for some reason.

rly (dismissed):

tests/unit/test_file.py
t-b: @rly I've pushed something but I need to review that again tomorrow.

t-b: @rly You were right. The additional entry does not reach the file.

```
h5dump -A unittest_file_create_date.nwb | grep -A 10 file_create_date

HDF5 "unittest_file_create_date.nwb" {
GROUP "/" {
   ATTRIBUTE ".specloc" {
      DATATYPE  H5T_REFERENCE { H5T_STD_REF_OBJECT }
      DATASPACE  SCALAR
      DATA {
      (0): GROUP 6512 /specifications 
      }
   }
   ATTRIBUTE "namespace" {
      DATATYPE  H5T_STRING {
--
   DATASET "file_create_date" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_ASCII;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
   }
   GROUP "general" {
      DATASET "institution" {
```

Questions:
- How can I fix that?
- How can I require a newer hdmf version to that the tests pass?

rly: To fix that, the dataset has to be chunked. @ajtritt -- is there a way to chunk only the `NWBFile.file_create_date` dataset? I am also in favor of blanket chunking all datasets in NWB...

To use changes in a newer hdmf version, the changes must have been released on PyPI. The recent "mode" function addition isn't released yet, but we could do that this week if these issues are pressing.

t-b: A new hdmf would be nice!

t-b: How do I force the stored dataset to be chunked?

I tried

```diff
diff --git a/src/pynwb/io/file.py b/src/pynwb/io/file.py
index 1ddeb310..2ec342d9 100644
--- a/src/pynwb/io/file.py
+++ b/src/pynwb/io/file.py
@@ -3,6 +3,7 @@ from hdmf.build import ObjectMapper
 from .. import register_map
 from ..file import NWBFile, Subject
 from ..core import ScratchData
+from hdmf.backends.hdf5.h5_utils import H5DataIO


 @register_map(NWBFile)
@@ -156,6 +157,10 @@ class NWBFileMap(ObjectMapper):
         dates = list(map(dateutil_parse, datestr))
         return dates

+    @ObjectMapper.object_attr('file_create_date')
+    def file_create_date_obj_attr(self, container, manager):
+        return H5DataIO(container.file_create_date, chunks=True)
+
     @ObjectMapper.constructor_arg('file_name')
     def name(self, builder, manager):
         return builder.name

```

but that does not work.

oruebel: Not sure if I have the right solution for you, but a couple of thoughts:

1. I think it is important to expose this behavior explicitly to user. While doing this implicitly behind the scenes is convenient, it make the process intransparent.
1. We should try not to mix front-end and backend functionality, i.e, using the HDF5-specific H5DataIO in the ObjectMapper (or Container) is problematic as this will not translate to other backends.
1. This issue also has come up with DynamicTable at some point, because we wanted all columns of the table to be chunked so they can be extended. @rly @ajtritt was that issue solved and would that same strategy apply here?

Ultimately, I think the core issue is that we want specific datasets to be written in a resizable fashion (so they can grow). In the case of HDF5 that requires chunking but for other backends that may or may not be the case. In that vain, I think what we may need is a generic (backend-agnostic) way to provide write-hints, which in this case would say "make this dataset resizeable".  I'm wondering whether we could add I/O hints on the builder for this and in the object-mapper a way to ask for I/O hints for fields. It would then be up to the backend to decide what to do with those I/O hints.

t-b: @oruebel It totally agree that a HDF5 specific solution is the wrong thing to do here. But up to now I don't have any solution at all.

bendichter:
@rly @t-b what's the status of this?

t-b:
We need to find a way to tell pynwb that certain datasets in HDF5 need to be written as chunked by default. Only then they are appendable. I don't know how to do that.

bendichter:
@t-b ah, ok. Sounds like a job for `H5DataIO`

tests/unit/test_file.py
t-b: I'm starting to work on this again.

@oruebel 

> I think it is important to expose this behavior explicitly to user. While doing this implicitly behind the scenes is convenient, it make the process intransparent.

What implicit part are you concerned about? The "making the dataset chunked" or "adding new entries in the file_create_dataset"? The latter is what nwb-schema says how file_create_dataset should be handled.

> Ultimately, I think the core issue is that we want specific datasets to be written in a resizable fashion (so they can grow). In the case of HDF5 that requires chunking but for other backends that may or may not be the case. In that vain, I think what we may need is a generic (backend-agnostic) way to provide write-hints, which in this case would say "make this dataset resizeable". I'm wondering whether we could add I/O hints on the builder for this and in the object-mapper a way to ask for I/O hints for fields. It would then be up to the backend to decide what to do with those I/O hints.

Yes that would be required. Of course my above hack is a hack and can not be merged as is, but I first wanted to get something working and then make the solution generalizable. I just saw that hdmf.builders.DatasetBuilder has a `chunks` argument as well.

I seem to not understand how the object mappers work. According to https://pynwb.readthedocs.io/en/stable/overview_software_architecture.html?highlight=architecture#objectmapper I would think that 

```diff
$ git diff .
diff --git a/src/pynwb/io/file.py b/src/pynwb/io/file.py
index 2c629ab7..a7057941 100644
--- a/src/pynwb/io/file.py
+++ b/src/pynwb/io/file.py
@@ -3,7 +3,7 @@ from hdmf.build import ObjectMapper
 from .. import register_map
 from ..file import NWBFile, Subject
 from ..core import ScratchData
-
+from hdmf.build import DatasetBuilder

 @register_map(NWBFile)
 class NWBFileMap(ObjectMapper):
@@ -152,6 +152,10 @@ class NWBFileMap(ObjectMapper):
         date = dateutil_parse(datestr)
         return date

+    @ObjectMapper.object_attr('file_create_date')
+    def file_create_date_obj_attr(self, container, manager):
+        return DatasetBuilder('file_create_date', data=container.file_create_date, chunks=True)
+
     @ObjectMapper.constructor_arg('file_create_date')
     def dateconversion_list(self, builder, manager):
         datestr = builder.get('file_create_date').data
```

should work, but it doesn't. Any hints?

PR #1165, opened 2020-01-30
Use proper file removal in tests

t-b:
See https://github.com/NeurodataWithoutBorders/pynwb/commit/cc3aef39b7b905ad4349b771431935d688717f68

### DIFF
--- a/tests/unit/test_file.py
+++ b/tests/unit/test_file.py
@@ -1,5 +1,4 @@
 import numpy as np
-import os
 import pandas as pd
 
 from datetime import datetime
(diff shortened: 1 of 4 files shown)


oruebel (dismissed):

tests/unit/test_file.py
```diff
@@ -406,7 +406,7 @@ def setUp(self):
 
     def tearDown(self):
         if os.path.exists(self.path):
```
rly: Minor: this if statement is already checked within `remove_test_file(self.path)` so it can be removed.

tests/unit/test_file.py
```diff
@@ -428,7 +428,7 @@ def setUp(self):
...
```
rly: Minor: this if statement is already checked within remove_test_file(self.path) so it can be removed.

rly (changes requested):

rly:
Under `tests/integration/hdf5/test_base.py`, in `TestTimeSeriesLinking.test_timestamps_linking`, the created file is not removed. Can you remove that one as well? I've been meaning to do it, but it fits nicely in this PR.

t-b:
@rly Thanks for the review. Done.

rly (dismissed):

rly (approved):

PR #1177, opened 2020-02-06
NF: initial setup for ASV benchmarks

yarikoptic:
## Motivation

To establish benchmarking for pynwb (and possibly later for hdmf).

For more info about ASV, visit https://asv.readthedocs.io/en/stable/.  I took the base for this PR from numpy's benchmarks setup. 

See benchmarks/README.rst within this PR for sample invocation.  But most likely

If it would be decided to proceed with this
- get a dedicated box where to run those benchmarks from time to time, and publish to github gh-pages branch the results so github.io could be "ab"used.  E.g. that is what numpy https://pv.github.io/numpy-bench/ and datalad  http://datalad.github.io/datalad/ do
- populate with (more) meaningful benchmarks
- adopt a helper from datalad to provide `asv compare` run for each PR. See e.g. https://github.com/datalad/datalad/pull/4080/checks?check_run_id=427777839#step:7:223 for a sample report

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,9 @@ tests/coverage/htmlcov
 
 # macos
 .DS_Store
+
+# ASV benchmarks artifacts
+benchmarks/results
+benchmarks/html
+benchmarks/env
+benchmarks/pynwb
--- a/benchmarks/README.rst
+++ b/benchmarks/README.rst
@@ -1,0 +1,62 @@
+..  -*- rst -*-
+
+================
+PyNWB benchmarks
+================
+
+Benchmarking PyNWB with Airspeed Velocity.
+
+
+Usage
+-----
+
+Run ASV commands (record results and generate HTML)::
+
+    cd benchmarks
+    asv run --skip-existing-commits --steps 10 ALL
+    asv publish
+    asv preview
+
+More on how to use ``asv`` can be found in `ASV documentation`_
+Command-line help is available as usual via ``asv --help`` and
+``asv run --help``.
+
+.. _ASV documentation: https://asv.readthedocs.io/
+
+
+Writing benchmarks
+------------------
+
+See `ASV documentation`_ for basics on how to write benchmarks.
+
+Some things to consider:
+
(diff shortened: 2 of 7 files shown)


oruebel:
@yarikoptic this looks very interesting. I like the idea of having regular benchmarks. I think we should be able to run these as cron-job on some linux box regularly. I'm wondering whether it would make sense to have the benchmarks as a separate repo. @ajtritt @rly  thoughts?

yarikoptic:
> I'm wondering whether it would make sense to have the benchmarks as a separate repo.

Not sure, have been thinking myself about it.  Having in the same repo could have benefits:
- you could validate (easier) that changes to API do not break benchmarks
- you could run (easier) comparison of effect of a PR on existing benchmarks (like we do in datalad)
- adding new benchmarks alongside within PRs for added/changed functionality in pynwb (we haven't reached that level of foo yet in datalad, but I would not be surprised if e.g. numpy requires that)

I guess that is why so far other projects (e.g. numpy) seems to carry benchmarks within their code repo.  They do keep results of benchmarks in a separate repo.  In datalad we added it as a submodule (`.asv/`) but I think that was a mistake and useless

oruebel:
> carry benchmarks within their code repo. They do keep results of benchmarks in a separate repo

That makes sense to me. Having the benchmark code in the same repo and the results hosted separately I think would be my preferred way to go as well.

yarikoptic:
wow it is an oldie, rebased now.  FWIW, disclaimer -- this PR was primarily to give you a skeleton to work from, not really for me to lead "benchmark pynwb" effort ;)

yarikoptic:
converted to draft. An interested party is welcome to take over

oruebel:
@yarikoptic thanks for getting this started. I'll take a look.

rly:
Note: We are benchmarking pynwb over in this repo: https://github.com/NeurodataWithoutBorders/nwb_benchmarks

@oruebel please close if the nwb_benchmarks repo can take over

oruebel:
Yes, I think nwb_benchmarks will ultimately take this over

PR #1146, opened 2020-01-17
Support NWB schema 2.2.2

rly:
This is a work in progress to support the changes in the NWB schema between versions 2.1.0 and 2.2.2.

- [x] Moved common data structures such as Container and DynamicTable to hdmf-common-schema. The hdmf-common-schema repo is now included as a submodule. See https://github.com/NeurodataWithoutBorders/nwb-schema/pull/307 for details
  - Make sure imports are still available from original location
- [x] Added "channel_conversion" dataset to ElectricalSeries to represent per-channel conversion factors.
- [x] Added "sampling_rate" and "unit" attributes to "waveform_mean" and "waveform_sd" datasets/columns in Units table.
- [x] Added "description" and "manufacturer" attributes to Device.
- [x] Deprecated ImagingPlane "manifold" in favor of "origin_coords" and "grid_spacing"
- [x] Use "text" data type for all DynamicTable "colnames". Previously, only ASCII was allowed.
- [x] Use "text" data type for electrode table columns "location" and "group_name". Previously, only ASCII was allowed.
- [x] Added to description to make electrode x,y,z consistent with CCF reference. http://help.brain-map.org/display/mousebrain/API#API-DownloadAtlas3-DReferenceModels
- [x] Added "position" dataset with compound data type x,y,z in ElectrodeGroup.
- [x] Avoid enforcing "uint64" for sweep numbers for better compatibility. Use uint instead which is 32bit.
- [x] Set dtype for Image and its subtypes to numeric. (note: technically this breaks backwards compatibility, in the schema, but the pynwb API has always enforced that Images have a numeric type, and realistically we do not think users are storing strings in an Image dataset.)
- [x] Added "resolution" attribute to "spike_times" column of Units.
- [x] Removed "required" key from dataset ImageSeries.field_of_view for schema language compliance.
- [x] Replaced "required" keys with "quantity" keys for ImagingPlane.origin_coords and ImagingPlane.grid_spacing for schema language compliance.
- [x] Refactored ImagingRetinotopy type to reduce redundancy.
- [x] Added "doc" key to ImagingRetinotopy.axis_2_power_map for schema language compliance.
- [x] Fixed makefiles for generating documentation on Windows.
- [x] Added optional "reference" column in "electrodes" table.
- [x] Changed dims of ImageSeries from (frame, y, x) to (frame, x, y) and (frame, z, y, x) to (frame, x, y, z) to be consistent with the dimension ordering in PlaneSegmentation.
[...]

PR #1182, opened 2020-02-08
OpticalSeries API fix

bendichter:
linked to https://github.com/NeurodataWithoutBorders/nwb-schema/pull/396

### DIFF
--- a/src/pynwb/image.py
+++ b/src/pynwb/image.py
@@ -130,7 +130,7 @@ class OpticalSeries(ImageSeries):
                      'field_of_view',
                      'orientation')
 
-    @docval(*get_docval(ImageSeries.__init__, 'name',),
+    @docval(*get_docval(ImageSeries.__init__, 'name'),
             {'name': 'data', 'type': ('array_data', 'data'), 'shape': ([None] * 3, [None, None, None, 3]),
              'doc': 'Images presented to subject, either grayscale or RGB'},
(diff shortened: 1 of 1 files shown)


src/pynwb/image.py
```diff
@@ -130,7 +130,9 @@ class OpticalSeries(ImageSeries):
                      'field_of_view',
                      'orientation')
 
-    @docval(*get_docval(ImageSeries.__init__, 'name', 'data'),  # required
+    @docval(*get_docval(ImageSeries.__init__, 'name'),
+            {'name': 'data', 'type': ('array_data', 'data'), 'shape': ([None] * 3, [None, None, None, 3]),
+             'doc': 'Images presented to subject, either grayscale or RGB'},
```
rly, line 135: The default should be None as it is in `ImageSeries` (in the event that an external file is provided).

rly (changes requested):

rly:
Note that the PyNWB Windows-py3.5-min-req test fails sporadically for this error:
```
ValueError: cannot expose native-only dtype 'g' in non-native byte order '<' via buffer interface
```
For now, if this is the only error, the test should be re-run until it succeeds. I suspect one of our dependencies (probably h5py) fixes the failure in a later version.

rly:
This change has been made in #1146. See the current dev code: https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/image.py#L134-L136

PR #1185, opened 2020-02-09
add efficient methods for getting first and last spike time

bendichter:


bendichter:
@ajtritt judging by the timeseries conversation #1175 , I think this would probably fall into the same category of utilities that should be more general-purpose. It would be great to have this functionality in pynwb eventually. Let's think about how we can do it properly and in the meantime I'll maintain this utility on my own for now.

PR #1188, opened 2020-02-10
Add 'unit' attribute back as an optional attribute in icephys classes

### LINKED ISSUE #1186: backwards compatibility issues in ipfx
When trying to update ipfx from pynwb==1.0.1 to pynwb==1.2.1, I run into this error:

```python
TypeError: VoltageClampStimulusSeries.__init__: unrecognized argument: 'unit'
```

The `unit` arg was removed because it is fixed in the schema. Could we put it back and include a warning that it will not be used?

rly:
Fix #1186. Also simplify calls to super since we no longer support Python 2.

### DIFF
--- a/src/pynwb/icephys.py
+++ b/src/pynwb/icephys.py
@@ -1,11 +1,12 @@
+import warnings
+
+from hdmf.common import DynamicTable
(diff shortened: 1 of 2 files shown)


oruebel (approved):

t-b:
Nice!

PR #1196, opened 2020-02-20
Load namespaces by default except in write mode and during validation

### LINKED ISSUE #1169: Set load_namespaces = True by default?
Should we set load_namespaces = True by default for `NWBHDF5IO`? I don't see why you would not want to load the cached specs that are usually necessary to read the file, but maybe I am not thinking of a use case.

rly:
Fix #1169.

### DIFF
--- a/docs/gallery/general/extensions.py
+++ b/docs/gallery/general/extensions.py
@@ -228,11 +228,9 @@ io.close()
 #     For more information on writing NWB files, see :ref:`basic_writing`.
 
 ####################
-# By default, PyNWB does not use the namespaces cached in a file--you must
-# explicitly specify this. This behavior is enabled by the *load_namespaces*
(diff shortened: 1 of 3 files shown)


t-b (approved):

t-b:
@rly I think the validation fails as we need to use `load_namespaces=True` in the validator when we validate against a spec from the NWB file.

t-b (changes requested):
See https://github.com/NeurodataWithoutBorders/pynwb/pull/1196#issuecomment-626798896

bendichter:
I think at this point this PR has so many merge conflicts we should just start over. Thoughts, @rly?

bendichter:
@rly can this be closed?

rly:
Addressed by #1748

PR #1200, opened 2020-02-29
Replace ic_electrode with icephys_electrode on NWBFile

### LINKED ISSUE #1181: Avoid using `ic` as abbreviation for icephys
See https://github.com/oruebel/ndx-icephys-meta/issues/36.

We probably should also have `add_icephys_electrode` and show a warning for `add_ic_electrode` in pynwb. In addition the tutorials should use the the function call.

oruebel:
## Motivation

Fix #1181 . replace the use of ic prefix as it has an ambigious meaning in the icephys domain. Use the prefix icephys instead

Specifically, this PR:

* Deprecate NWBFile.__init__(ic_electrodes ...) in favor of icephys_electrodes
* Deprecate NWBFile.ic_electrodes property in favor of NWBFile.icephys_electrodes
* Deprecate NWBFile.get/add/create_ic_electrode functions in favor NWBFile.get/add/create_icephys_electrode
* Update tests to use the new functions
* Add tests to check that the deprecation warnings are raised
* Update icephys tutorial to use icephys_electrode instead of ic_electrode

## Checklist

### DIFF
--- a/src/pynwb/file.py
+++ b/src/pynwb/file.py
@@ -140,11 +140,11 @@ class NWBFile(MultiContainerInterface):
             'get': 'get_imaging_plane'
         },
         {
-            'attr': 'ic_electrodes',
-            'add': 'add_ic_electrode',
+            'attr': 'icephys_electrodes',
+            'add': 'add_icephys_electrode',
             'type': IntracellularElectrode,
-            'create': 'create_ic_electrode',
-            'get': 'get_ic_electrode'
+            'create': 'create_icephys_electrode',
+            'get': 'get_icephys_electrode'
         },
         {
             'attr': 'ogen_sites',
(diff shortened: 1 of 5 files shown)


t-b (dismissed):
Fantastic!

src/pynwb/file.py
```diff
@@ -275,7 +275,10 @@ class NWBFile(MultiContainerInterface):
             {'name': 'electrode_groups', 'type': Iterable,
              'doc': 'the ElectrodeGroups that belong to this NWBFile', 'default': None},
             {'name': 'ic_electrodes', 'type': (list, tuple),
-             'doc': 'IntracellularElectrodes that belong to this NWBFile', 'default': None},
+             'doc': 'DEPRECATED use icephys_electrodes parameter instead. '
+                    'IntracellularElectrodes that belong to this NWBFile', 'default': None},
+            {'name': 'icephys_electrodes', 'type': (list, tuple),
```
rly: Until we enforce the use of keyword-only arguments in neurodata_type constructors, the new docval argument should be added to the end of the list to maintain backward compatibility. That said, I cannot imagine anyone specifying positional arguments through this point in the `NWBFile` argument list, so adding the argument here is almost certainly fine.

tests/unit/test_icephys.py
```diff
@@ -4,22 +4,111 @@
...
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
+            nwbfile = NWBFile(
+                session_description='NWBFile icephys test',
+                identifier='NWB123',  # required
+                session_start_time=datetime(2017, 4, 3, 11, tzinfo=tzlocal()))
+            nwbfile.add_ic_electrode(self.icephys_electrode)
+            assert len(w) == 1
+            assert issubclass(w[-1].category, DeprecationWarning)
+            assert "deprecated" in str(w[-1].message)
```
rly: I believe this could be simplified to the following:
```suggestion
        nwbfile = NWBFile(
                session_description='NWBFile icephys test',
                identifier='NWB123',  # required
                session_start_time=datetime(2017, 4, 3, 11, tzinfo=tzlocal()))

        msg = "deprecated, use NWBFile.add_icephys_electrode instead"
        with self.assertWarnsWith(DeprecationWarning, msg):
            nwbfile.add_ic_electrode(self.icephys_electrode)
```

src/pynwb/file.py
oruebel: Thanks for the suggestion @rly! I updated the code accordingly.

oruebel:
@rly this should be good to go now

rly (approved):

Issue #1204, opened 2020-03-06
Adding electrode column rel_x causes unintended behavior


rly:
In PyNWB 1.3.0, `rel_x`, `rel_y`, `rel_z`, and `reference` were added as optional columns. However, existing code in SpikeInterface adds these columns and no longer works. See https://github.com/SpikeInterface/spikeextractors/issues/323

Adding a column that already exists in the electrodes table should raise an error. It does not do this because of a bug in how I initialized the electrodes table. I will fix that.

But more generally, when we add an optional column to a table specification, if a user adds a column with the same name, should an error or a warning be raised? If an error is raised, that kind of breaks backwards compatibility. If a warning is raised, but the behavior of the user's code is changed (e.g., the user-defined column is not added), that also breaks backwards compatibility but quietly (which seems worse!). If a warning is raised but the user's code runs as intended, then the resulting data does not conform to the new spec (which I think is even worse!)

So I think an error should be raised, and this breaking is an unfortunate consequence of updating to the new schema and PyNWB version. @oruebel @ajtritt @bendichter thoughts?

<details><summary>Offending code: (click to expand)</summary>

```python
            if nwbfile.electrodes is None or 'rel_x' not in nwbfile.electrodes.colnames:
                nwbfile.add_electrode_column('rel_x', 'x position of electrode in electrode group')
            if nwbfile.electrodes is None or 'rel_y' not in nwbfile.electrodes.colnames:
                nwbfile.add_electrode_column('rel_y', 'y position of electrode in electrode group')
            for m in channel_ids:
                if m not in nwb_elec_ids:
                    location = recording.get_channel_property(m, 'location')
                    while len(location) < 2:
                        location = np.append(location, [0])
                    impedence = -1.0
                    grp_name = recording.get_channel_groups(channel_ids=[m])
                    grp = nwbfile.electrode_groups[str(grp_name[0])]
                    nwbfile.add_electrode(
                        id=m,
                        x=np.nan, y=np.nan, z=np.nan,
                        rel_x=float(location[0]), rel_y=float(location[1]),
                        imp=impedence,
                        location='unknown',
                        filtering='none',
                        group=grp,
                    )
```
</details>

oruebel:
> But more generally, when we add an optional column to a table specification, if a user adds a column with the same name, should an error or a warning be raised?

[...]

PR #1205, opened 2020-03-10
create Electrodes class to replace electrodes table

rly:
@ajtritt I am trying to create an `Electrodes` class to use for the electrodes table instead of a `DynamicTable` to solve #1204. Without being its own class, we cannot specify in PyNWB that certain columns of the electrodes table are optional.

This involves changing the `ObjectMapper` for `NWBFile` such that it constructs an `Electrodes` object instead of a `DynamicTable` object for the `NWBFile.electrodes` constructor argument. 

I thought of a few ways to do it:
1. Construct an `Electrodes` object from the `/general/extracellular_ephys/electrodes/` group builder by setting each value and creating each `VectorData` individually, but this would be very tedious and involve a lot of hard coding. 
2. Construct an `Electrodes` object from the `/general/extracellular_ephys/electrodes/` group builder by setting the values of the `Electrodes` object to the values of the `DynamicTable` that is already generated. This doesn't work because I cannot change the parent of the `VectorData` classes to the new object. That is what I have coded up in the PR here.
3. Override the container class associated with the `/general/extracellular_ephys/electrodes/` group builder. This doesn't seem possible.

Can you think of a better way to do this? Would this even be possible? Would it require substantial changes to HDMF to work?

Stepping back at the source problem, it might be easier to change how adding columns works in `DynamicTable` to handle optional columns. What do you think?

### DIFF
--- a/src/pynwb/file.py
+++ b/src/pynwb/file.py
@@ -18,8 +18,7 @@ from .icephys import IntracellularElectrode, SweepTable, PatchClampSeries
 from .ophys import ImagingPlane
 from .ogen import OptogeneticStimulusSite
 from .misc import Units
-from .core import NWBContainer, NWBDataInterface, MultiContainerInterface, \
-                  ScratchData, LabelledDict
+from .core import NWBContainer, NWBDataInterface, MultiContainerInterface, ScratchData, LabelledDict
 from hdmf.common import DynamicTableRegion, DynamicTable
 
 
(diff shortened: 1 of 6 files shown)


rly:
I thought of another (slightly hacky) way to get around this issue. We can just promote the `DynamicTable` object constructed by the `ObjectMapper` to be an `Electrodes` object and add the new columns to the class. I updated the PR to reflect this method. This requires https://github.com/hdmf-dev/hdmf/pull/323

rly:
@mavaylon1 is restarting this effort with a new `ElectrodesTable` neurodata type in https://github.com/NeurodataWithoutBorders/pynwb/pull/1890. Will close this PR.

PR #1219, opened 2020-04-09
ENH+RF: inform which unit value is actually overwritten (ignored and set)

yarikoptic:
While at it I RF code to avoid duplication (evil, tiresome to change) in the touched code

## Motivation

Otherwise user is left wondering "what was ignored? it might have been millivolts and now it became volts?" and alike.

### DIFF
--- a/src/pynwb/icephys.py
+++ b/src/pynwb/icephys.py
@@ -116,10 +116,7 @@ class CurrentClampSeries(PatchClampSeries):
              'default': 'volts'})
     def __init__(self, **kwargs):
         name, data, unit, electrode, gain = popargs('name', 'data', 'unit', 'electrode', 'gain', kwargs)
-        if unit != 'volts':
-            warnings.warn("Unit for %s '%s' is ignored and will be set to 'volts' as per NWB 2.1.0."
-                          % (self.__class__.__name__, name))
-        unit = 'volts'
+        unit = ensure_unit(self, name, unit, 'volts', '2.1.0')
         bias_current, bridge_balance, capacitance_compensation = popargs(
             'bias_current', 'bridge_balance', 'capacitance_compensation', kwargs)
         super().__init__(name, data, unit, electrode, gain, **kwargs)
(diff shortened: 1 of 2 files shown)


rly:
Thanks @yarikoptic . Looks good. Could you please fix the flake8 errors:
```
tests/unit/test_icephys.py:252:121: E501 line too long (123 > 120 characters)
tests/unit/test_icephys.py:302:121: E501 line too long (126 > 120 characters)
```

yarikoptic:
doh -- thanks for the buzz. force pushed the fixed version.
Someone could look into configuring pre-commit for this to ensure that all commits are flake8 ok.

FWIW in dandi-cli @satra inflicted on us pre-commit with black magic -- so it reformats code upon commit to be consistent etc: https://github.com/dandi/dandi-cli/blob/master/.pre-commit-config.yaml . I kinda like it now ;)

satra:
an aside:

@rly any reason unit names are plural? btw, we are trying to adopt this: https://people.csail.mit.edu/jaffer/MIXF/MIXF-10 across several projects so units are parseable. i don't know if nwb has any standard for unit specification.

src/pynwb/icephys.py
```diff
@@ -116,10 +116,7 @@ class CurrentClampSeries(PatchClampSeries):
              'default': 'volts'})
     def __init__(self, **kwargs):
         name, data, unit, electrode, gain = popargs('name', 'data', 'unit', 'electrode', 'gain', kwargs)
-        if unit != 'volts':
-            warnings.warn("Unit for %s '%s' is ignored and will be set to 'volts' as per NWB 2.1.0."
-                          % (self.__class__.__name__, name))
-        unit = 'volts'
+        unit = ensure_unit(self, name, unit, 'volts', '2.1.0')
```
yarikoptic, line 119: BTW ignorant question:  It is `CurrentClampSeries`.  Current is measured in `amperes` -- why is it `volts` here?  There is similarly opposite below for `VoltageClampSeries`. 
@rly?

satra: @yarikoptic - a current clamp measures voltage and voltage clamp measures current. https://en.wikipedia.org/wiki/Electrophysiology#Intracellular_recording

yarikoptic: gotcha -- thank you @satra!

rly (approved):

rly:
@yarikoptic the pre-commit hooks look useful! I will try to add them. As far as I can tell, adding the file `.pre-commit-config.yaml` doesn't change anyone's git setup. The user has to install pre-commit and set up the git hook scripts for the hooks to run. So, it's safe to add the YAML file to the root. The flake8 hook would be great too.

@satra Yarik and I have been discussing the units issue here: https://github.com/NeurodataWithoutBorders/nwb-schema/pull/324  I will raise the issue with the others in the core dev team later this week.

PR #1221, opened 2020-04-10
change imaging_rate to be optional

bendichter:
## Motivation

https://github.com/NeurodataWithoutBorders/nwb-schema/pull/429

### DIFF
--- a/docs/gallery/domain/ophys.py
+++ b/docs/gallery/domain/ophys.py
@@ -20,8 +20,6 @@ clarity, we define them here:
 from datetime import datetime
 from dateutil.tz import tzlocal
 
-import numpy as np
-
 from pynwb import NWBFile
 from pynwb.ophys import TwoPhotonSeries, OpticalChannel, ImageSegmentation, Fluorescence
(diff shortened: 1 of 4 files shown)


bendichter:
I had to switch to kwargs on a number of tests. Technically this breaks backwards compatibility for `ImagingPlane` if the user was using ordered args.

rly (approved):

bendichter:
@rly do we also need to update the commit of nwb-schema?

rly:
Yes, good point. I was going to update it in the schema_2.2.3 branch (where this would merge into) but it makes more sense to update the schema commit here. Could you do that? I'll approve after, and then you can merge it.

bendichter:
I am not sure I know how..

bendichter:
Also, I didn't realize this was going into another sub-branch. I thought it was going into dev. I'm OK with changing the nwb-schema commit there.

rly:
From the pynwb project root, `cd src/pynwb/nwb-schema`. Then run `git pull origin dev`

PR #1238, opened 2020-05-18
automatically add core schema URL

### LINKED ISSUE #1236: Store the git hash of the nwb-schema submodule in the NWBFile
## Description

In the event that the nwb-schema submodule of PyNWB is accidentally updated to a non-release version (the default git branches of nwb-schema and hdmf-common-schema contain the bleeding edge and not necessarily the latest release), then storing the git hash of the loaded submodule in the file would help decipher which unreleased version of the schema was used.

Note: the current workflow requires that all unreleased versions of the schema (the state of the default branches of those repos) have the suffix "-alpha" in the version string. This helps determine that an unreleased version of the schema was used, but does not differentiate between multiple commits that occur between schema releases. The suffix "-alpha" might also be accidentally forgotten. Storing the git hash will solve both issues.

It would also be nice to store the version of PyNWB (or git hash) in the NWBFile.

## Proposed solution

Modify 'package_data' in `setup.py` to store the git hash of the nwb-schema submodule as a resource:
https://github.com/NeurodataWithoutBorders/pynwb/blob/9d734631c680d6e5bb430d806cd21d2651f093d8/setup.py#L36

Load the git hash and store it in the file. See as example:
https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/__init__.py#L24-L28

ajtritt:
## Motivation

fix #1236

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,26 @@ from setuptools import setup, find_packages
 
 import versioneer
 
+import configparser
+from os.path import join as pjoin
+
+
+def get_schema_sha(url_path):
+    parser = configparser.RawConfigParser()
+    config_path = pjoin('.git', 'modules', 'src', 'pynwb', 'nwb-schema', 'config')
+    parser.read(config_path)
+    url = parser.get('remote "origin"', 'url')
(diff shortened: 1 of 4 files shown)


oruebel:
@In-vidrio

PR #1242, opened 2020-05-26
pass file object to parent when loading namespaces

NileGraddis:
This is a very minor change, companion to https://github.com/hdmf-dev/hdmf/pull/348. It updates the NWBHDF5IO constructor to pass an argued file (or None) to its parent when loading namespaces. 

This should not be merged until a version of hdmf with pr 438 incorporated is released.


rly (approved):

rly:
Thanks, @NileGraddis !

PR #1245, opened 2020-06-01
Add support for NWB schema 2.3.0

rly:
The current release of PyNWB, version 1.3.2, supports nwb-schema version 2.2.5. This PR adds support for nwb-schema minor version 2.3.0.

This PR tracks the dev branch of nwb-schema to ensure that changes in nwb-schema are compatible with PyNWB or are accompanied with corresponding changes in PyNWB. Those changes in PyNWB should be merged into this branch.

- [x] make waveforms doubly indexed
- [x] Update PyNWB to work with changes in nwb-schema:
  - [x] Add optional ``waveforms`` column to the ``Units`` table.
  - [x] Add optional ``strain`` field to ``Subject``.
  - [x] Add to ``DecompositionSeries`` an optional ``DynamicTableRegion`` called ``source_channels``.
  - [x] Add to ``ImageSeries`` an optional link to ``Device``.
  - [x] Add optional ``continuity`` field to ``TimeSeries``.
  - [x] Add optional ``filtering`` attribute to ``ElectricalSeries``.
  - [x] Clarify documentation for electrode impedance and filtering.
  - [x] Add description of extra fields.
  - [x] Set the ``stimulus_description`` for ``IZeroCurrentClamp`` to have the fixed value ``N/A``.
  - [x] Update hdmf-common-schema from 1.1.3 to version 1.5.0.
    - [x] The HDMF-experimental namespace was added, which includes the ``ExternalResources`` and ``EnumData``
	      data types. Schema in the HDMF-experimental namespace are experimental and subject to breaking changes at any time.
    - [x] Added experimental data type ``ExternalResources`` for storing ontology information / external resource references.
    - [x] Added experimental data type ``EnumData`` to store data from a set of fixed values.
    - [x] Changed dtype for datasets within ``CSRMatrix`` from 'int' to 'uint' and added missing `data_type_inc: Container`
	      to the ``CSRMatrix`` type.
    - [x] Added data type ``SimpleMultiContainer``, a Container for storing other Container and Data objects together.
    - [x] Added data type ``AlignedDynamicTable``, a DynamicTable type with support for categories (or sub-headings) each described by a separate DynamicTable.
    - [x] Fixed missing dtype for ``VectorIndex``.
    - [x] ``VectorIndex`` now extends ``VectorData`` instead of ``Index``.
    - [x] Removed unused and non-functional ``Index`` data type.
- [x] Run validation tests
- [x] Update changelog
- [x] After new version of nwb-schema is released, update the nwb-schema submodule to point to latest release: `cd src/pynwb/nwb-schema/`, `git checkout dev`, `git pull origin dev`, `git fetch --tags`, `git checkout dev`, `cd ../../..`, `git add .`, `git commit -m "Update nwb-schema submodule to [new_release]`, and `git push`

After merging:
- This branch should be merged into dev without rebased commits, NOT squashed commits.


rly:
Only validation errors remain before this is ready to merge. (This is not tested on per-PR CI, but nightly CI.)

The main error is:
```
ValueError: data type 'NWBFile' not found in namespace hdmf-experimental
```

hdmf-experimental is cached within the file but NWB objects should not be validated against it.

requirements-doc.txt
```diff
@@ -2,4 +2,4 @@ sphinx
 matplotlib
 sphinx_rtd_theme
 sphinx-gallery
-allensdk
\ No newline at end of file
+allensdk  # ntoe that as of allensdk 2.10.0, python 3.8 is not supported
```
oruebel:
```suggestion
allensdk  # note that as of allensdk 2.10.0, python 3.8 is not supported
```

src/pynwb/file.py
```diff
@@ -55,7 +56,8 @@ class Subject(NWBContainer):
             {'name': 'subject_id', 'type': str, 'doc': 'a unique identifier for the subject', 'default': None},
             {'name': 'weight', 'type': str, 'doc': 'the weight of the subject', 'default': None},
             {'name': 'date_of_birth', 'type': datetime, 'default': None,
-             'doc': 'datetime of date of birth. May be supplied instead of age.'})
+             'doc': 'datetime of date of birth. May be supplied instead of age.'},
+            {'name': 'strain', 'type': str, 'doc': 'the strain of the subject', 'default': None})
```
oruebel: Do we have a recommendation for how the strain should be specified. If so, it would be nice to add an example in the 'doc' .

src/pynwb/file.py
```diff
@@ -55,7 +56,8 @@ class Subject(NWBContainer):
             {'name': 'subject_id', 'type': str, 'doc': 'a unique identifier for the subject', 'default': None},
             {'name': 'weight', 'type': str, 'doc': 'the weight of the subject', 'default': None},
```
oruebel: Can we add the unit for 'weight' or describe how the string should be formatd. .

Seems like a candidate for changing from str to float and then formatting the string in the API.

src/pynwb/icephys.py
```diff
@@ -71,7 +71,7 @@ class PatchClampSeries(TimeSeries):
              'doc': 'IntracellularElectrode group that describes the electrode that was used to apply '
                      'or record this data.'},
             {'name': 'gain', 'type': 'float', 'doc': 'Units: Volt/Amp (v-clamp) or Volt/Volt (c-clamp)'},  # required
-            {'name': 'stimulus_description', 'type': str, 'doc': 'the stimulus name/protocol', 'default': "NA"},
+            {'name': 'stimulus_description', 'type': str, 'doc': 'the stimulus name/protocol', 'default': "N/A"},
```
oruebel, line 74:
```suggestion
            {'name': 'stimulus_description', 'type': str, 'doc': 'the stimulus name/protocol. Deprecated as of NWB 2.3.0. Set to N/A by default.', 'default': "N/A"},
```

src/pynwb/io/base.py
```diff
@@ -29,6 +29,7 @@ def __init__(self, spec):
         self.map_spec('unit', data_spec.get_attribute('unit'))
         self.map_spec('resolution', data_spec.get_attribute('resolution'))
         self.map_spec('conversion', data_spec.get_attribute('conversion'))
+        self.map_spec('continuity', data_spec.get_attribute('continuity'))
```
oruebel, line 32: It seems strange that this line is not being hit by tests since it is being set in "tests/integration/hdf5/test_base.py" . May be worth checking why this line is reported as not covered.

src/pynwb/misc.py
```diff
@@ -267,15 +268,20 @@ class DecompositionSeries(TimeSeries):
              'doc': 'a table for describing the frequency bands that the signal was decomposed into', 'default': None},
             {'name': 'source_timeseries', 'type': TimeSeries,
              'doc': 'the input TimeSeries from this analysis', 'default': None},
+            {'name': 'source_channels', 'type': DynamicTableRegion, 'doc': 'the channels that provided the source data',
```
oruebel: It would be useful to indicate in the doc string for ``source_channels`` where this DynamicTableRegion points to. E.g., " In the case of electrical recordings this is typically a DynamicTableRegion pointing to the ElectrodesTable as part of NWBFile.electrodes, similar to ElectricalSeries.electrodes." If this is different for an ImageSeries than it would be nice to mention this as well.

oruebel (commented):
I added a few minor comments, but otherwise this looks good.

src/pynwb/icephys.py
rly: It is deprecated only on `IZeroClampSeries`. I'll update the docstring for that.

src/pynwb/io/base.py
rly: Code coverage is assessed using unit tests only. Integration tests will hit the IO layer.

src/pynwb/file.py
rly: I updated the docstring and now allow float in the constructor with a recommendation to use that.

I am wary of allowing only float. That would be a good change for a major release.

It would be nice to change this in the schema to be float, with a 'unit' attribute saying that it is in kg.

src/pynwb/file.py
rly: I updated this. Please review.

src/pynwb/file.py
oruebel: I agree. Allowing both float and str makes sense. Updating it in the schema makes sense as well. I'll create an issue for this. There are a few other things that are problematic with this field.

oruebel (dismissed):

oruebel (dismissed):

oruebel (approved):

PR #1248, opened 2020-06-04
add Units.waveforms, unit test, and integration test

bendichter:
## Motivation
pynwb API changes for adding waveforms to the Units table

depends on https://github.com/NeurodataWithoutBorders/nwb-schema/pull/448

## How to test the behavior?
```
nwbfile.add_unit(waveforms=[[1, 2, 3], [1, 2, 3], [1, 2, 3]])
```

## Checklist

### DIFF
--- a/src/pynwb/misc.py
+++ b/src/pynwb/misc.py
@@ -137,11 +137,11 @@ class Units(DynamicTable):
         'resolution'
     )
 
-    waveforms_desc = 'Individual waveforms for each spike. If the dataset is three-dimensional, the third dimension ' \
-                     'shows the response from different electrodes that all observe this unit simultaneously. In this' \
-                     ' case, the `electrodes` column of this Units table should be used to indicate which electrodes ' \
-                     'are associated with this unit, and the electrodes dimension here should be in the same order as' \
-                     ' the electrodes referenced in the `electrodes` column of this table.'
+    waveforms_desc = ('Individual waveforms for each spike. If the dataset is three-dimensional, the third dimension ' 
+                      'shows the response from different electrodes that all observe this unit simultaneously. In this' 
+                      ' case, the `electrodes` column of this Units table should be used to indicate which electrodes ' 
+                      'are associated with this unit, and the electrodes dimension here should be in the same order as' 
+                      ' the electrodes referenced in the `electrodes` column of this table.')
     __columns__ = (
         {'name': 'spike_times', 'description': 'the spike times for each unit', 'index': True},
         {'name': 'obs_intervals', 'description': 'the observation intervals for each unit',
--- a/src/pynwb/misc.py
+++ b/src/pynwb/misc.py
@@ -137,10 +137,10 @@ class Units(DynamicTable):
         'resolution'
     )
 
-    waveforms_desc = ('Individual waveforms for each spike. If the dataset is three-dimensional, the third dimension ' 
-                      'shows the response from different electrodes that all observe this unit simultaneously. In this' 
-                      ' case, the `electrodes` column of this Units table should be used to indicate which electrodes ' 
-                      'are associated with this unit, and the electrodes dimension here should be in the same order as' 
+    waveforms_desc = ('Individual waveforms for each spike. If the dataset is three-dimensional, the third dimension '
+                      'shows the response from different electrodes that all observe this unit simultaneously. In this'
+                      ' case, the `electrodes` column of this Units table should be used to indicate which electrodes '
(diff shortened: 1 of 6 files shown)


bendichter:
fyi @taravanviegen

src/pynwb/misc.py
```diff
@@ -158,7 +164,7 @@ class Units(DynamicTable):
             {'name': 'waveform_unit', 'type': str,
              'doc': 'Unit of measurement of the waveform means', 'default': 'volts'},
             {'name': 'resolution', 'type': 'float',
-             'doc': 'The smallest possible difference between two spike times', 'default': None},
```
rly, line 161: Note: with trailing commas, when you add an element to this list and view the diff in a version control system like Git, then only the new element will show up as changed, rather than both the new element and the previous line which just has a new comma (see example with your addition of the 'waveforms' column. PEP8 allows this. I recently discovered this and prefer it. You can change it if you want. Just wanted to mention why it might have been added.

tests/unit/test_misc.py
```diff
@@ -113,6 +113,16 @@ def test_add_spike_times(self):
         self.assertEqual(ut['spike_times'][0], [0, 1, 2])
         self.assertEqual(ut['spike_times'][1], [3, 4, 5])
 
+    def test_add_waveforms(self):
+        ut = Units()
+        ut.add_unit(waveforms=[[1, 2, 3], [1, 2, 3]])
```
rly, line 118: Please use floats for the waveform values in both tests. Users are most likely to enter floats. Using floats should also prevent the dtype conversion warning of int -> float on write.

rly (changes requested):

tests/unit/test_misc.py
bendichter: @rly Actually, come to think of it, I think users are also likely to enter `int16`, as this is a very common way for acquisition systems to record data. I know that Neuroscope for instance stores spike waveforms in `int16`. So now I think we should allow any numeric type. I've changed the schema to reflect this. What do you think?

ajtritt: @bendichter Isn't unit waveform an average of many spike waveforms?

If you want to change the data type of a column, please do so in another PR.

ajtritt (commented):
Please address the comment from @rly

tests/unit/test_misc.py
bendichter: @ajtritt the corresponding schema change is [here](https://github.com/NeurodataWithoutBorders/nwb-schema/pull/448), responding to [this issue](https://github.com/NeurodataWithoutBorders/nwb-schema/issues/248). We would like to add a "waveforms" column to the Units table that holds *all* of the sorted waveforms, not the average.

src/pynwb/misc.py
bendichter: hmm I see. It looks like a mistake but I can see why you like it. I guess I'm split on it.

src/pynwb/misc.py
```diff
@@ -137,6 +137,11 @@ class Units(DynamicTable):
         'resolution'
     )
 
+    waveforms_desc = 'Individual waveforms for each spike. If the dataset is three-dimensional, the third dimension ' \
+                     'shows the response from different electrodes that all observe this unit simultaneously. In this' \
+                     ' case, the `electrodes` column of this Units table should be used to indicate which electrodes ' \
+                     'are associated with this unit, and the electrodes dimension here should be in the same order as' \
+                     ' the electrodes referenced in the `electrodes` column of this table.'
```
rly:
```suggestion
    waveforms_desc = ('Individual waveforms for each spike. If the dataset is three-dimensional, the third dimension ' 
                      'shows the response from different electrodes that all observe this unit simultaneously. In this' 
                      ' case, the `electrodes` column of this Units table should be used to indicate which electrodes ' 
                      'are associated with this unit, and the electrodes dimension here should be in the same order as' 
                      ' the electrodes referenced in the `electrodes` column of this table.')
```

rly: Minor: style-wise, PEP8 says we should break long lines by wrapping them in parentheses instead of using backslashes to escape the new line character: https://www.python.org/dev/peps/pep-0008/#:~:text=Long%20lines%20can%20be%20broken,a%20backslash%20for%20line%20continuation.&text=(See%20the%20previous%20discussion%20on,such%20multiline%20with%2Dstatements.)

See also discussion on: https://stackoverflow.com/questions/1874592/how-to-write-very-long-string-that-conforms-with-pep8-and-prevent-e501

rly (approved):

PR #1260, opened 2020-07-11
Rmv MultiContainerInterface

bendichter:
## Motivation

Depends on https://github.com/hdmf-dev/hdmf/pull/399

Move `MultiContainerInterface`  from `pynwb.core` to `hdmf.container`. 

The motivations for this are:
1. allow other hdmf derivatives to use these structures
2. allow for automatically generating `MultiContainerInterface` objects from pynwb extensions

The transition is mostly straightforward. Functions were copy/pasted functions and tests from one repo to another, with straightforward changes like replacing `NWBContainer` to `Container`, etc.

In order to keep the same interface for pynwb users, I am planning to do:

```python
from hdmf.container import MultiContainerInterface as hdmf_MultiContainerInterface

class MultiContainerInterface(NWBMixin, hdmf_MultiContainerInterface):
    pass
```

The tests are failing because this repo still points to an outdated version of hdmf, but if you fix that there is still one issue that comes up across many tests:
```
AttributeError: 'NWBFile' object has no attribute 'file_create_date'
```

I don't understand why this change would have an impact on storing 'file_create_date'. Any idea why this might happen, @rly, @ajtritt, or @oruebel ?


bendichter:
@rly do you know why I might be getting this error with `file_created_date`?

ajtritt:
> I don't understand why this change would have an impact on storing 'file_create_date'. 

@bendichter can you provide the full error statement and point to one of the tests that exposes it?

rly:
Tests now pass with the corresponding changes in HDMF. 

@bendichter the main issue with your earlier code was that the PyNWB version of MultiContainerInterface needs to inherit from `NWBDataInterface`.

This also starts to resolve #1266. The next steps for that is to 1) change all `__nwbfields__` in PyNWB classes to `__fields__` (`__nwbfields__` will still work as evidenced by tests passing) and 2) add a test to ensure that `__nwbfields__` still works. I will wait for feedback on #1266 before moving forward.

bendichter:
@rly woohoo! Thanks!

rly:
I reverted the change to `__nwbfields__`. That will be saved for a subsequent PR.

src/pynwb/core.py
```diff
@@ -332,309 +316,6 @@ def __getitem__(self, idx):
...
-
-    * 'get' to name the method for getting Container instances
-
-    * 'attr' to name the attribute that stores the Container instances
-
-    * 'type' to provide the Container object type
-
-    See LFP or Position for an example of how to use this.
-    '''
+class MultiContainerInterface(NWBDataInterface, hdmf_MultiContainerInterface):
```
oruebel, line 319: Add docstring to the at least give a hint why this class is here e.g,: "Defined in PyNWB for API backward compatibility. See HDMF MultiContainterInterface for details".

oruebel (dismissed):

src/pynwb/core.py
bendichter:
```suggestion
class MultiContainerInterface(NWBDataInterface, hdmf_MultiContainerInterface):
    """Defined in PyNWB for API backward compatibility. See HDMF MultiContainterInterface for details"""
```

oruebel (approved):

PR #1265, opened 2020-07-22
Add optional link to Device in ImageSeries

rly:
## Motivation

Support https://github.com/NeurodataWithoutBorders/nwb-schema/pull/454

Requires https://github.com/NeurodataWithoutBorders/nwb-schema/pull/454 to be merged

## Checklist

### DIFF
--- a/src/pynwb/image.py
+++ b/src/pynwb/image.py
@@ -18,7 +18,8 @@ class ImageSeries(TimeSeries):
     __nwbfields__ = ('dimension',
                      'external_file',
                      'starting_frame',
-                     'format')
+                     'format',
+                     'device')
 
     @docval(*get_docval(TimeSeries.__init__, 'name'),  # required
             {'name': 'data', 'type': ('array_data', 'data', TimeSeries), 'shape': ([None] * 3, [None] * 4),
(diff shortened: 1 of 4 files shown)


bendichter:
what about the descendants, `TwoPhotonSeries` and `OpticalSeries`? Will they support devices?

rly:
Thanks. I just added that. I also added it in `ImageMaskSeries` which is the other subclass of `ImageSeries` because based on the schema you can have an optional link to a Device in a ImageMaskSeries. It doesn't really make sense to, but currently we have no way in the schema to say that an optional link in a parent type is not allowed in the child type.

rly:
I guess since it doesn't make sense to have a separate link to a `Device` in `ImageMaskSeries`, we can restrict that in the API. It depends on whether we want the API to support the full flexibility of the schema or not.

bendichter:
If we leave it out, then it won't be supported on read, will it? I think if someone creates a file that NWB supports, we should be able to read it. We can use the doc of that arg to say that this field would likely not be needed.

rly:
Right, if we leave it out, it would not be supported on read. Good point

oruebel:
Removing/hiding attributes inherited from a parent type should not be allowed. If an attribute on a parent class does not make sense in a child class then that typically points to problem in the model.

bendichter (approved):

PR #1271, opened 2020-07-30
Improve docstrings for data field of TimeSeries types

### LINKED ISSUE #1268: docval doc for the data field of all TimeSeries types should explicitly mention dimensions
It is a common mistake for new users of PyNWB to write the data field of a TimeSeries object in the wrong orientation. While it is possible for them to see the correct orientation in the nwb-schema docs, or in the best practices page, they rarely check these, instead relying on the PyNWB API documentation. It would make the API more user-friendly if they docs for all the TimeSeries were explicit about the expected orientation of the data array.

rly:
## Motivation

Fix #1268.

## Checklist

### DIFF
--- a/src/pynwb/ogen.py
+++ b/src/pynwb/ogen.py
@@ -41,7 +41,7 @@ class OptogeneticSeries(TimeSeries):
 
     @docval(*get_docval(TimeSeries.__init__, 'name'),  # required
             {'name': 'data', 'type': ('array_data', 'data', TimeSeries), 'shape': (None, ),  # required
-             'doc': 'The data values. Must be 1D, where the first dimension must be time.',
+             'doc': 'The data values over time. Must be 1D.',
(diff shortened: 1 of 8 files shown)


src/pynwb/ogen.py
```diff
@@ -41,7 +41,7 @@ class OptogeneticSeries(TimeSeries):
 
     @docval(*get_docval(TimeSeries.__init__, 'name'),  # required
             {'name': 'data', 'type': ('array_data', 'data', TimeSeries), 'shape': (None, ),  # required
-             'doc': 'The data this TimeSeries dataset stores. Can also store binary data e.g. image frames'},
+             'doc': 'The data values. Must be 1D, where the first dimension must be time.',
```
bendichter:
```suggestion
             'doc': 'The data values over time. Must be 1D.',
```

bendichter (commented):
looks great! Thanks for doing this!

bendichter (approved):

PR #1274, opened 2020-08-06
Require TimeSeries.data and .unit, update ImageSeries, OpticalSeries

### LINKED ISSUE #1220: TimeSeries can be written without 'data' dataset
## Description

The `TimeSeries/data` dataset is required in the NWB 2.0+ schema. However, in PyNWB, you can create a `TimeSeries` without data. This results in a validation error. This has come up before but I forget where. Credit to @yarikoptic for identifying the issue again.

## Steps to Reproduce

```python
from pynwb import NWBFile, TimeSeries, NWBHDF5IO, validate
from datetime import datetime


ts = TimeSeries(name='ts_name',
                unit='ts_unit',
                timestamps=[0., 1., 2.])

nwbfile = NWBFile(session_description='ts',
                  identifier='i',
                  session_start_time=datetime.now().astimezone())

nwbfile.add_acquisition(ts)

filename = 'nwbfile.nwb'

with NWBHDF5IO(filename, 'w') as io:
    io.write(nwbfile)

with NWBHDF5IO(filename, 'r') as io:
    errors = validate(io)
    print(errors)
    nwbfile = io.read()
```
## Environment

    Python Executable: Python
    Python Version: Python 3.7
    Operating System: Windows
    PyNWB Version: latest 1.3.0
    HDMF Version: latest dev

## Checklist

rly:
## Motivation

Fix #1220. 

This PR includes breaking changes to the API and will thus be pushed out in PyNWB 2.0.

Changes:
- `data` is now required in `__init__` of `TimeSeries` and all subclasses of `TimeSeries`, except for `ImageSeries` and its subclasses, where `external_file` can be provided instead of `data`. 
  - This creates a weird situation in the inheritance hierarchy where a subclass does not contain a required field of the superclass. As a result, if `data` is not passed into the constructor of `ImageSeries`, then `data` will be set to the default value so that `super().__init__` may be called. This results in the default value being written to disk, which is not ideal. An alternative solution is to split `ImageSeries` into two separate classes (and neurodata_types), one where `data` is required and one where `external_file` is required.
  - If a `TimeSeries` is read without `data`, then `data` will be set to a default value: an empty 1D numpy array. This is also true for all subclasses of `TimeSeries` except for `ImageSeries` and its subclasses and `DecompositionSeries` where the shape of data cannot be 1-D, so the default value is an empty 3D numpy array.
  
- `unit` is now required in `__init__` of `TimeSeries` and all subclasses of `TimeSeries`, except for `ImageSeries` and its subclasses.
  - Like for `data` above, if a `TimeSeries` is read without `unit`, then `unit` will be set to a default value `'unknown'`. 
  - Like for `data` above, if an `ImageSeries` is constructed without `unit`, then `unit` will be set to a default value so that `super().__init__` may be called. This results in the default value being written to disk, which is not ideal.
  - In `ImageSeries`, `unit` is required only if `data` is provided.

- Previously, `data` was required in `OpticalSeries.__init__` even though an `external_file` could be provided. This is a bug, so `data` is now made optional in `OpticalSeries.__init__`. This has the side effect of moving the position of `data` to later in the argument list, which may break code that relies on positional arguments for `OpticalSeries.__init__`.

- This is a separate change, but included in this PR. NWB 2.4 introduces `TimeSeriesReference`. The `TimeIntervals` table has been updated to use `TimeSeriesReference` instead of a bare 3-element tuple.

- Added test files with `TimeSeries` and `ImageSeries` types that are missing `data` and `unit`. 
- Fixed many tests that used data-less or unit-less `TimeSeries` objects.

TODO:
- [x] add more testing
- [x] update changelog

## Checklist


rly:
@oruebel @ajtritt @bendichter Could you please take a look at this PR? It is not ready to merge, but I would like feedback on it before I finalize it. Thanks.

oruebel (approved):

PR #1280, opened 2020-08-12
Add support and documentation for export from HDMF version 2

rly:
## Motivation

See https://github.com/hdmf-dev/hdmf/pull/388

I also changed the docval for `NWBHDF5IO.export` from what it was for `HDF5IO.export`:
```python
HDF5IO.export(path=..., container=...)
NWBHDF5IO.export(path=..., nwbfile=...)
```

## Checklist


docs/gallery/general/modifying_data.py
```diff
@@ -0,0 +1,188 @@
+"""
+.. _modifying_data:
+
+Modifying data in an NWB file
```
oruebel: I think the term "modifying" is confusing here. The reason being that folks will typically think about modifying individual fields, i.e., datasets and attributes. I think we should either add a section on how to update datasets and attributes or change the title to something like "Adding/Removing containers from an NWB file".

docs/gallery/general/modifying_data.py
```diff
@@ -0,0 +1,188 @@
...
+=========================================
+
+This tutorial explains how to modify data from an existing NWB file and either write the data back to the
```
oruebel: Same here, we should explicitly state what we mean by "modify"

oruebel (dismissed):
Aside from the minor comment the tutorial description this looks good.

rly:
@oruebel I made the suggested changes and added another section on Object IDs

oruebel (approved):

Issue #1294, opened 2020-09-02
cannot add hdmf's MultiContainerInterface to NWB file


bendichter:
## Description

We can now automatically generate `MultiContainerInterface` API classes in HDMF, but these new objects cannot be added to the processing module of an NWB file, because it is not one of the accepted types. The `pynwb` brand of `MultiContainerInterface` can be added to processing modules because it inherits from `NWBDataInterface`, but this new HDMF type does not.

Two possible ways to address this:
1) Make this new type one of the accepted types by adding `hdmf.container.MultiContainerInterface` as an allowed type in the docval of `ProcessingModule`
2) Make it so that `MultiContainerInterface` objects that were created using `pynwb.get_class` inherit from one of the accepted types, most likely `NWBDataInterface`

## Steps to Reproduce

```python
from hdmf.spec import GroupSpec, SpecCatalog, SpecNamespace, NamespaceCatalog
from hdmf.build import TypeMap

# create auto-generated MultiContainerInterface
CORE_NAMESPACE = 'test_core'

spec_catalog = SpecCatalog()
bar_spec = GroupSpec('A test group specification with a data type', data_type_def='Bar')
spec_catalog.register_spec(bar_spec, 'test.yaml')
namespace = SpecNamespace('a test namespace', CORE_NAMESPACE, [{'source': 'test.yaml'}],
                          version='0.1.0',
                          catalog=spec_catalog)
namespace_catalog = NamespaceCatalog()
namespace_catalog.add_namespace(CORE_NAMESPACE, namespace)
type_map = TypeMap(namespace_catalog)

multi_spec = GroupSpec('A test extension that contains a multi',
                       data_type_def='Multi',
                       groups=[GroupSpec(
                           data_type_inc=bar_spec,
                           doc='test multi',
                           quantity='*')]
                       )
spec_catalog.register_spec(multi_spec, 'extension.yaml')
Bar = type_map.get_container_cls(CORE_NAMESPACE, 'Bar')
Multi = type_map.get_container_cls(CORE_NAMESPACE, 'Multi')
multi = Multi(bars=[Bar('my_bar')])


# add MCI to NWB file
from pynwb import NWBFile
import datetime


nwbfile = NWBFile(session_description='test foraging2',
                  identifier='behavior_session_uuid',
                  session_start_time=datetime.datetime.now())

mod = nwbfile.create_processing_module('test', 'test')
mod.add(multi)
```

```python
---------------------------------------------------------------------------
[...]

PR #1295, opened 2020-09-03
Set default neurodata_type_inc for NWBGroupSpec, NWBDatasetSpec

rly:
## Motivation

When creating a `NWBGroupSpec` or `NWBDatasetSpec`, the default values for `neurodata_type_inc` should be `NWBContainer` and `NWBData` respectively so that the new neurodata_types work nicely with the rest of the PyNWB API which often assumes that neurodata_types are of one of those classes (or are DynamicTables). 

See comments in #1294 

This PR also adds a call to `deepcopy` on some docval arguments to make sure that one call to e.g., `NWBAttributeSpec.__init__()` does not change the default values for a later call `NWBAttributeSpec.__init__()` (e.g., if `default=list()`).

## Checklist

### DIFF
--- a/src/pynwb/spec.py
+++ b/src/pynwb/spec.py
@@ -8,21 +8,18 @@ from hdmf.utils import docval, get_docval, call_docval_func
 from . import CORE_NAMESPACE
 
 
-def __swap_inc_def(cls, default_nd_type_inc=None):
+def __swap_inc_def(cls):
     args = get_docval(cls.__init__)
     clsname = 'NWB%s' % cls.__name__
     ret = list()
     # do not set default neurodata_type_inc for base hdmf-common types that should not have data_type_inc
-    for arg in args:
-        if arg['name'] == 'Container' or arg['name'] == 'Data':
-            default_nd_type_inc = None
     for arg in args:
         if arg['name'] == 'data_type_def':
             ret.append({'name': 'neurodata_type_def', 'type': str,
                         'doc': 'the NWB data type this spec defines', 'default': None})
         elif arg['name'] == 'data_type_inc':
             ret.append({'name': 'neurodata_type_inc', 'type': (clsname, str),
-                        'doc': 'the NWB data type this spec includes', 'default': default_nd_type_inc})
+                        'doc': 'the NWB data type this spec includes', 'default': None})
         else:
             ret.append(copy(arg))
     return ret
@@ -114,8 +111,7 @@ class BaseStorageOverride(object):
         proxy = super(BaseStorageOverride, cls)
         kwargs[proxy.def_key()] = kwargs.pop(cls.def_key())
         kwargs[proxy.inc_key()] = kwargs.pop(cls.inc_key())
-        args = [kwargs.pop(x['name']) for x in get_docval(proxy.__init__) if 'default' not in x]
-        return args, kwargs
+        return kwargs
 
 
 _dtype_docval = __swap_inc_def(DtypeSpec)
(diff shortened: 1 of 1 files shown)


src/pynwb/spec.py
```diff
@@ -128,34 +128,46 @@ def __init__(self, **kwargs):
...
+    '''
+
+    @docval(*deepcopy(_dataset_docval))
     def __init__(self, **kwargs):
-        args, kwargs = self._translate_kwargs(kwargs)
-        super(NWBDatasetSpec, self).__init__(*args, **kwargs)
+        kwargs = self._translate_kwargs(kwargs)
+        # set data_type_inc to NWBData only if it is not specified and the type is not an HDMF base type
+        if (kwargs['data_type_inc'] is None and kwargs['data_type_def'] is not None and
+                kwargs['data_type_def'] != 'Data'):
```
bendichter:
```suggestion
        if kwargs['data_type_inc'] is None and kwargs['data_type_def'] not in (None, 'Data'):
```

src/pynwb/spec.py
```diff
@@ -128,34 +128,46 @@ def __init__(self, **kwargs):
...
+    '''
+
+    @docval(*deepcopy(_group_docval))
     def __init__(self, **kwargs):
-        args, kwargs = self._translate_kwargs(kwargs)
-        super(NWBGroupSpec, self).__init__(*args, **kwargs)
+        kwargs = self._translate_kwargs(kwargs)
+        # set data_type_inc to NWBData only if it is not specified and the type is not an HDMF base type
+        if (kwargs['data_type_inc'] is None and kwargs['data_type_def'] is not None
+                and kwargs['data_type_def'] != 'Container' and kwargs['data_type_def'] != 'CSRMatrix'):
```
bendichter:
```suggestion
        if kwargs['data_type_inc'] is None and kwargs['data_type_def'] not in (None, 'Container', 'CSRMatrix'):
```

src/pynwb/spec.py
```diff
@@ -128,34 +128,44 @@ def __init__(self, **kwargs):
...
+    Classes will automatically include NWBContainer if None is specified.
+    '''
+
+    @docval(*deepcopy(_group_docval))
     def __init__(self, **kwargs):
-        args, kwargs = self._translate_kwargs(kwargs)
-        super(NWBGroupSpec, self).__init__(*args, **kwargs)
+        kwargs = self._translate_kwargs(kwargs)
+        # set data_type_inc to NWBData only if it is not specified and the type is not an HDMF base type
+        if kwargs['data_type_inc'] is None and kwargs['data_type_def'] not in (None, 'Container', 'CSRMatrix'):
```
ajtritt, line 160: why is there a special case for `CSRMatrix` and not other types from hdmf-common?

rly: `Container`, `Data`, and `CSRMatrix` are the only types in hdmf-common without a `data_type_inc`. This will be fixed in hdmf-common-schema 1.2.1 from the just merged PR https://github.com/hdmf-dev/hdmf-common-schema/pull/40 , so it might be possible to remove `CSRMatrix` from the list here.

The if statement is necessary to prevent an infinite loop that occurs when loading the hdmf-common schema: Without this, loading `data_type_def: Container` with `data_type_inc: None` will result in `data_type_inc: NWBContainer`.

ajtritt: Gotcha. Thanks for the clarifications

ajtritt (approved):

PR #1308, opened 2020-10-17
ENH: Improve PlaneSegmentation.add_roi doc

rly:
## Motivation

The docstring for `PlaneSegmentation.add_roi(...)` should describe why/when to use image_mask vs pixel_mask or voxel_mask.

Based on feedback from the Allen Institute.

## Checklist


oruebel (dismissed):

bendichter:
IMO I'd much prefer image_mask over pixel_mask in all cases. If the image_mask is numerically sparse, gzip should be able to take care of that. This will probably take up even less space than pixel_mask, especially if image_mask in binary

bendichter (changes requested):
IMO I'd prefer `image_mask` over `pixel_mask` in all cases. If the `image_mask` is numerically sparse, gzip should be able to take care of that. `image_mask` + gzip will probably take up even less space than unzipped `pixel_mask`, especially if the `image_mask` data in binary. If you want to go next level, you could also set the `fill_value` to 0 and only fill in the chunks that contain some of the ROI. I can't think of a good compelling case for using `pixel_mask` over `image_mask`

oruebel:
@bendichter with regard to storage in the file that is true, but keep in mind that in general, the same optimizations (chunking, compression) are not available in memory. I.e., once a user has to load the data pixel_mask will in general be more efficient for sparse ROI's in memory than image_mask. For small images the difference is probably not important but for large-images full image masks per ROI can become very expensive.

bendichter:
@oruebel ok that's a good point

rly:
Also, applying compression to an NWB file is an advanced step that I think many users are not yet doing. If we are going to encourage usage that is optimized for the compressed case, then we should make sure to encourage compression in the tutorials and documentation.

I updated the docstring again to be more clear and address the points you raised. Feel free to edit or make further suggestions.

oruebel (approved):

src/pynwb/ophys.py
```diff
@@ -238,21 +238,40 @@ def __init__(self, **kwargs):
...
+        """Add a Region Of Interest (ROI) data to this PlaneSegmentation table
+
+        Add an ROI using an image mask, a pixel mask, or a voxel mask. An image mask is an image with
+        the same size of the data image where positive values mark the ROI weights for each pixel of
+        the image. A pixel mask is a list of triplets [(x1, y1, weight1), (x2, y2, weight2), ...] or
+        Nx3 array that indicates the ROI weights for each pixel. Pixels without a weight have a
+        weight of 0. A voxel mask is a list of quadruplets or Nx4 array, that is similar to the pixel
+        mask, except the format is (x1, y1, z1, weight1).
+
+        If the NWB file will be compressed, then using an image mask will usually result in a smaller
```
bendichter, line 343: Are we suggesting that users compress the entire NWB file? Is there a way of just compressing the image masks?

rly: Good point. I will add 'data' to the allowed types for the masks and test this out.

PR #1309, opened 2020-10-29
Fix up scratch API

### LINKED ISSUE #1307: ScratchData notes passed as None instead of empty string in add_scratch
## Description

If an `np.ndarray` is added in scratch with `nwbfile.add_scratch(ndarray,name='some_name'`) without a `notes` argument, `ScratchData` is created with `notes = None` and fails with
`TypeError: ScratchData.__init__: incorrect type for 'notes' (got 'NoneType', expected 'str')`

Adding something like this before https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/file.py#L695 should fix it.
```
if notes is None:
    notes = ''
```

## Steps to Reproduce

```
data = np.random.randint(0,30,(5,9))
nwbfile.add_scratch(data, name = 'some_name')
```

## Environment

    Python Executable: Conda
    Python Version: 3.7
    Operating System: Linux
    HDMF Version: pynwb 1.3.3, hdmf 1.6.4

## Checklist

- [ ] Have you included a clear description of the problem you are trying to solve?

rly:
## Motivation

Fix #1307, add missing unit tests, fix incorrect docval, clean up warnings and errors. 

`ScratchData` requires notes, so set the default to empty string. A warning is raised in `add_scratch` if the notes argument is None or empty string. A warning is also raised in `add_scratch` if a pandas.DataFrame is provided and the table_description argument (required for DynamicTable) is None or empty string.

@bendichter I remember you saying something about empty strings being problematic. Could you elaborate?

## Checklist


oruebel:
A key problem with empty string is that it is basically just side-stepping a required field by leaving it blank. I think requiring that users provide a description for the data they put into scratch is not much of a burden and at least provides some minimal documentation what those fields are about.

rly:
This PR is ready to go. Previously the default values for 'notes' and 'table_description' were empty string (both are required in the schema). I changed these arguments to be required, which is a breaking change. This will be pushed in the next version, PyNWB 2.0.0, which will include other breaking changes (e.g. TimeSeries.data will be required).

src/pynwb/file.py
```diff
@@ -669,36 +668,54 @@ def add_stimulus_template(self, timeseries):
...
+             'type': ('scalar_data', np.ndarray, list, tuple, pd.DataFrame, DynamicTable, NWBContainer, ScratchData),
+             'doc': 'The data to add to the scratch space'},
             {'name': 'name', 'type': str,
-             'help': 'the name of the data. Only used when passing in numpy.ndarray, list, or tuple',
+             'doc': ('The name of the data. Required and used only when passing in numpy.ndarray, list, tuple, '
+                     'or pandas.DataFrame'),
              'default': None},
             {'name': 'notes', 'type': str,
-             'help': 'notes to add to the data. Only used when passing in numpy.ndarray, list, or tuple',
+             'doc': ('Notes to add to the data. Highly recommended and used only when passing in numpy.ndarray, '
```
oruebel: Why does scalar data not have notes?

src/pynwb/file.py
```diff
@@ -669,36 +668,54 @@ def add_stimulus_template(self, timeseries):
...
+                     'list, or tuple.'),
              'default': None},
             {'name': 'table_description', 'type': str,
-             'help': 'description for the internal DynamicTable used to store a pandas.DataFrame',
-             'default': ''})
+             'doc': ('Description for the internal DynamicTable used to store a pandas.DataFrame. Highly recommended '
```
oruebel: Why is table_description from notes? table_description is supplied when you have a table and notes are supplied otherwise, and it seems they both serve the same purpose. Would it not be simpler to just have a single argument ``description`` instead?

src/pynwb/file.py
rly: Good catch. It does. I'll update the docstring.

src/pynwb/file.py
rly: Yes, that would be simpler. @ajtritt - since you had initially designed this, what do you think?

ajtritt: Yeah, that would be simpler.

ajtritt (changes requested):
I agree with Oliver.

ajtritt (dismissed):

src/pynwb/core.py
```diff
@@ -93,8 +93,8 @@ class ScratchData(NWBData):
     __nwbfields__ = ('notes',)
 
     @docval({'name': 'name', 'type': str, 'doc': 'the name of this container'},
-            {'name': 'data', 'type': ('array_data', 'data', Data), 'doc': 'the source of the data'},
-            {'name': 'notes', 'type': str, 'doc': 'notes about the data', 'default': ''})
+            {'name': 'data', 'type': ('scalar_data', 'array_data', 'data', Data), 'doc': 'the source of the data'},
+            {'name': 'notes', 'type': str, 'doc': 'notes about the data'})
```
oruebel: While we are changing things for Scratch, I think it would make sense to change ``notes`` for ``ScratchData`` to ``description`` for consistency with ``add_scratch``

oruebel (dismissed):

oruebel (approved):

src/pynwb/core.py
rly: Done

Issue #1189, opened 2020-02-10
Use keyword-only arguments in neurodata_type constructors


rly:
**Problem**:
For the spec of a given neurodata_type, if we make a required group/dataset/attribute/link optional, then in the PyNWB API, the only way to make the field optional in the constructor is to add `'default': None` to the docval of the field. If the field has a required field following it in the docval list of fields, and if the user provides positional arguments, then the parsing of arguments by `docval` will be incorrect. This has prevented us from making such changes as making `NWBFile.identifier` optional because it would break all existing code that uses positional arguments.

**Proposed solution**:
I propose we do not allow positional arguments for `__init__` of all neurodata_types. This has the added benefit of making user code more readable, albeit a little bulkier. Users would have to consult documentation to get the argument names, but in my experience, most users do this already because no one has memorized the many possible arguments and their order for each class. The order is also inconsistent across classes.

This would break any existing code that uses positional arguments and so should probably wait for the next major release.

In regard to the autodoc, we can use the syntax of keyword-only arguments that was introduced in Python 3: https://www.python.org/dev/peps/pep-3102/ . Basically, place a * in the arguments list before the keyword-only arguments, e.g., the function signature in the docs would appear as:
```python
class A:
    def __init__(self, *, arg1=None, arg2=None):
        print(kwargs)
```
And we should add an explicit note saying positional arguments are not allowed in this method.

In order to transition to this change in the next major version, I propose we show warnings when positional arguments are passed to neurodata_type constructors, and update our docs and tests to use only keyword arguments.

Now:
- [ ] warn when positional arguments are passed to neurodata_type constructors
- [ ] update tests
- [ ] update docs

Later:
- [ ] change warning to error

rly:
We decided today to move forward with raising a warning if a user uses positional arguments for constructors, and also `add_electrode` and other methods that use arguments based on the schema.

PR #1316, opened 2020-11-11
Add unit to SpatialSeries arguments

### LINKED ISSUE #1315: CompassDirection.spatialSeries.data.unit problem
[`CompassDirection`](https://pynwb.readthedocs.io/en/stable/pynwb.behavior.html#pynwb.behavior.CompassDirection) is a `MultiContainerInterface` that takes `SpatialSeries` objects, and specifies that these objects should have units "degrees" or "radians". `SpatialSeries` in the nwb schema provides "meters" as a default value for data.units, but pynwb's implementation hard-codes that the units must be meters. To accommodate `CompassDirection` and to be consistent with the schema, the units arg of `SpatialSeries` should be a default value of "meters" but should be changeable.

cc @weiglszonja

weiglszonja:
## Motivation

Fixes #1315 

## How to test the behavior?
There is a new unittest added to `src/pynwb/tests/unit/test_behavior.py` to test SpatialSeries.unit default value of "meters" can be changed to another SI unit. 

## Checklist


rly (approved):

rly:
Thanks for this PR, and sorry for the delay. One thing to note is that the 'unit' keyword argument will be inserted in the middle of the `SpatialSeries.__init__()` argument list, so any code that enters `conversion` as a positional argument will have changed behavior. From my experience, no one is calling `SpatialSeries.__init__()` in this way, and we are releasing a new major version, so this breaking change to fix a bug is OK.

rly:
This issue would be more easily addressed after this issue is addressed: https://github.com/NeurodataWithoutBorders/pynwb/issues/1189 but that will not be for a future major release.

oruebel (approved):

PR #1317, opened 2020-11-12
remove NWBTable

ajtritt:
## Motivation

NWBTable is moving to HDMF. See https://github.com/hdmf-dev/hdmf/pull/442


rly:
Closed in favor of #1325 which includes the changes here.

PR #1349, opened 2021-04-05
Add/icephys meta

oruebel:
## Motivation

This PR is for integration of the ndx-icephys-meta structures with NWB core. Related PRs on the other repos are:

https://github.com/NeurodataWithoutBorders/nwb-schema/pull/470
https://github.com/hdmf-dev/hdmf-common-schema/pull/57
https://github.com/hdmf-dev/hdmf/pull/551
https://github.com/oruebel/ndx-icephys-meta/pull/82

Changes on this PR include

* [X] Declare SweepTable as deprecated on init
* [X] Integrated the new types from the ndx-icephys-meta extensions schema
     *  [X] TimeSeriesReferenceVectorData
     *  [X] IntracellularRecordingsTable
     *  [X] SimultaneousRecordingsTable
     *  [X] SequentialRecordingsTable
     *  [X] RepetitionsTable
     *  [X] ExperimentalConditionsTable
     * [X] Port changes from ICEphysFile to NWBFile itself
* [X] Add new tutorial for the ICEPhys data structures and declare previous tutorial using SweepTable as deprecated
* [x] https://github.com/NeurodataWithoutBorders/pynwb/pull/1383

Current status update on the integration of the ndx-icephys-meta with the NWB core:

- [X] **NWB Schema:** PR https://github.com/NeurodataWithoutBorders/nwb-schema/pull/470 has been merged 
- [X] **HDMF 3.0** has been released which included the following relevant changes: 
    - [X] Add ``AlignedDynamicTable``  https://github.com/hdmf-dev/hdmf/pull/551 
    - [X] Add ``AlignedDynamciTable``  https://github.com/hdmf-dev/hdmf/pull/571
    - [X] Fix slicing for ``DynamicTableRegion`` https://github.com/hdmf-dev/hdmf/pull/579
- [ ] **HDMF 3.1** release pending. This included the following relevant changes:
     - [x] Improve support for linked ``DyanmicTable`` (e.g, add ``to_hierarchical_dataframe`` function) https://github.com/hdmf-dev/hdmf/pull/645. This PR has already undergone review and is ready to be merged once approved.
 - [ ] **PyNWB**
     - [x]  This PR implements the integration of the icephys table structure with PyNWB https://github.com/NeurodataWithoutBorders/pynwb/pull/1349 . The implementation is complete and the PR's are ready for review. 
     - [ ] Merge this PR

## Checklist


oruebel:
@rly @ajtritt this PR is also ready for review. We should probably wait with the merge until https://github.com/NeurodataWithoutBorders/nwb-schema/pull/470 and https://github.com/hdmf-dev/hdmf/pull/551 have been merged so all the CI can be run. But other than that this PR should be complete.

lvsltz:
https://github.com/hdmf-dev/hdmf/commit/b8d9987e900eae9f0dcafdaf8521d049ceb64c31 and 2.5.1 trigger this error on NWBFile write
This error did not occur with https://github.com/hdmf-dev/hdmf/commit/67cffecad50fbc31f8406367f9442b6e3bd87a4a

```python
ContainerConfigurationError: CurrentClampSeries 'ccs__ResetITC__001' does not have attribute 'data__continuity' for mapping to spec: {'doc': 'Optionally describe the continuity of the data. Can be "continuous", "instantaneous", or "step". For example, a voltage trace would be "continuous", because samples are recorded from a continuous process. An array of lick times would be "instantaneous", because the data represents distinct moments in time. Times of image presentations would be "step" because the picture remains the same until the next timepoint. This field is optional, but is useful in providing information about the underlying data. It may inform the way this data is interpreted, the way it is visualized, and what analysis methods are applicable.', 'name': 'continuity', 'required': False, 'dtype': 'text'}
```

oruebel:
> [hdmf-dev/hdmf@b8d9987](https://github.com/hdmf-dev/hdmf/commit/b8d9987e900eae9f0dcafdaf8521d049ceb64c31) and 2.5.1 trigger this error on NWBFile write

@lvsltz  thanks for the bug report. We are currently working on a minor release for PyNWB which will fix this issue. I expect this release to be done early next week, so it is simplest to just wait for the release. I.e., rather than syncing this branch now with the upcoming fixes it will be simpler to wait for those fixes to be merged and then catch this branch up with dev.

Just for additional details as to the source of this issue. The problems is due to a mismatch between the schema I am using on this branch and the version of container classes in PyNWB. What happened is the following, in nwb-schema I branched of off the main dev branch to add the schema changes for the icephys classes. The dev branch of nwb-schema, however, contained changes at that time that were not yet supported by the dev branch of PyNWB . And so, I ended up pulling a version of the the schema into this branch that I should not have (i.e., I accidentally used incompatible versions of PyNWB and the schema on this branch). With the new version of HDMF, however, we are now getting errors/warnings when there are fields in the schema that are not being mapped, where previously those errors were silently ignored. I.e., this is an issue that already existed on this branch but that simply went unnoticed, and with the stricter error checking has now been revealed. Once I catch up this branch with the changes from the upcoming release, this issue should be resolved.

src/pynwb/base.py
```diff
@@ -282,3 +282,16 @@ def __init__(self, **kwargs):
...
+
+@register_class('TimeSeriesReferenceVectorData', CORE_NAMESPACE)
+class TimeSeriesReferenceVectorData(VectorData):
+    """
+    Column storing references to a TimeSeries (rows). For each TimeSeries this VectorData
+    column stores the start_index and count to indicate the range in time to be selected
+    as well as an object reference to the TimeSeries.
+    """
+
+    @docval(*get_docval(VectorData.__init__))
```
oruebel: Update to set default_name to 'timeseries'

src/pynwb/icephys.py
```diff
@@ -319,15 +335,492 @@ def __get_row_ids(self, sweep_number):
...
+            allow_extra=True)
+    def add_recording(self, **kwargs):
+        """
+        Add a single recording to the IntracellularRecordingsTable table.
+
+        Typically, both stimulus and response are expected. However, in some cases only a stimulus
+        or a resposne may be recodred as part of a recording. In this case, None, may be given
+        for either stimulus or response, but not both. Internally, this results in both stimulus
+        and response pointing to the same timeseries, while the start_index and index_count for
+        the invalid series will both be set to -1.
```
lvsltz, line 494: 1. This behavior (in which a PatchClampSeries appears to exist but does not, in fact) was quite confusing and at times facilitated bugs in my code. If possible, not having this artificial reference would be easier for a user, I believe. Or `IntracellularRecordingsTable` should be generated with `None`s when those indices are -1.

2. Another confusion comes from the fact that, if either the `start_index` or the `index_count` is -1, the other one must also be -1 (I don't see how one can be positive while the other one -1). But this relation is never (?) enforced or clearly specified, so a user may wonder if they have to check for both values or one would be enough.

src/pynwb/icephys.py
```diff
@@ -319,15 +335,492 @@ def __get_row_ids(self, sweep_number):
...
+                            electrodes=electrodes,
+                            responses=responses,
+                            stimuli=stimuli,
+                            **kwargs)
+        return len(self) - 1
+
+    @staticmethod
+    def __compute_index(start_index, index_count, time_series, name):
+        start_index = start_index if start_index >= 0 else 0
+        num_samples = time_series.num_samples
```
oruebel: Update the function to handle the case if time_series is None and add docs.

```suggestion
    def __compute_index(start_index, index_count, time_series, name):
        """
        Internal helper function to compute the start_index and index_count
        to use for the stimulus and response column
        
        :param start_index: The start_index provided by the user
        :param index_count: The index count provided by the user
        :param time_series: The timeseries object to reference. May be None.
        :param name: Name of the table. Used only to enhance error reporting
        
        :raises IndexError: If index_count cannot be determined or start_index+index_count 
            are outside of the range of the timeseries.
        
        :returns: A tuple of integers with the start_index and index_count to use
        """
        if time_series is None:
            return -1, -1
        start_index = start_index if start_index >= 0 else 0
        num_samples = time_series.num_samples
```

src/pynwb/icephys.py
```diff
@@ -319,15 +335,492 @@ def __get_row_ids(self, sweep_number):
...
+        # Compute the start and stop index if necessary
+        if stimulus is not None:
+            stimulus_start_index, stimulus_index_count = self.__compute_index(stimulus_start_index,
+                                                                              stimulus_index_count,
+                                                                              stimulus, 'stimulus')
+        if response is not None:
+            response_start_index, response_index_count = self.__compute_index(response_start_index,
+                                                                              response_index_count,
+                                                                              response, 'response')
+
```
oruebel: With the updates to ```__compute_index``` we can always compute the start_index and index_count and ensure they are set correctly.

```suggestion
        # Compute the start and stop index if necessary
        stimulus_start_index, stimulus_index_count = self.__compute_index(stimulus_start_index,
                                                                          stimulus_index_count,
                                                                          stimulus, 'stimulus')
        response_start_index, response_index_count = self.__compute_index(response_start_index,
                                                                          response_index_count,
                                                                          response, 'response')
```

src/pynwb/icephys.py
oruebel: Thanks for the helpful feedback. 

>     1. If possible, not having this artificial reference would be easier for a user,

I agree. Unfortunately HDF5 does not a Null value and NAN is not defined for integer types. In the file itself, we therefore unfortunately can't get around storing the data with -1 values. That being said, replacing the -1 values on read with None (i.e, when slicing into the table or when converting it Pandas) should be doable. I'll take a look at this next week. 

>     2. Another confusion comes from the fact that, if either the `start_index` or the `index_count` is -1

I think this is similar to the issue you raised in 1. I think what we need to do is: 1) update the slicing and conversion to pandas for the ``TimeSeriesReferenceVectorData`` to convert entries that have -1 as start_index to None values, 2) update the docs of ``TimeSeriesReferenceVectorData`` to describe this behavior. The -1 value is here again used as a way to indicate None. The reason we need this here, is because our stimulus and response columns must align.

For the  ``add_recording`` function, I think we may also want to change the default value of start_index and index_count parameters to None.  The ``add_recording`` function calls ``__compute_index`` to determine the correct start_index and index_count for the given TimeSeries to use. I.e., if, e.g., not stimulus is given then both ``start_index`` and ``index_count`` will be automatically set to -1. Similarly, if a TimeSeries is provided for a stimulus and ``start_index`` and ``index_count`` were set to -1 then they will be converted by ``__compute_index`` to the appropriate values for the given TimeSeries, i.e, start_index will be set to 0 and the index_count to the length of the TimeSeries. 

>     But this relation is never (?) enforced

It is enforced when calling ``add_recording`` but it should also be enforced in ``TimeSeriesReferenceVectorData`` and the docs should be updated to mention this as well.

oruebel: @lvsltz I have addressed this issue in https://github.com/NeurodataWithoutBorders/pynwb/pull/1383. When creating a file you can now use "None" for missing stimulus/response and when reading the data missing values are masked using numpy Masked arrays. As such, a user should no longer see the -1 values when accessing the data via PyNWB. 

BTW, I am mostly finished with https://github.com/NeurodataWithoutBorders/pynwb/pull/1383 and https://github.com/hdmf-dev/hdmf/pull/645 (mostly adding tests and documentation). Once those are complete things should be ready for release. I expect that we should have all of this released in HDMF and PyNWB by the end of this month.

src/pynwb/base.py
oruebel: This has been addressed #1383

oruebel:
Just a quick status update on the integration of the ndx-icephys-meta with the NWB core:

- [X] **NWB Schema:** PR https://github.com/NeurodataWithoutBorders/nwb-schema/pull/470 has been merged 
- [X] **HDMF 3.0** has been released which included the following relevant changes: 
    - [X] Add ``AlignedDynamicTable``  https://github.com/hdmf-dev/hdmf/pull/551 
    - [X] Add ``AlignedDynamciTable``  https://github.com/hdmf-dev/hdmf/pull/571
    - [X] Fix slicing for ``DynamicTableRegion`` https://github.com/hdmf-dev/hdmf/pull/579
    - [X]
- [ ] **HDMF 3.1** release pending. This included the following relevant changes:
     - [ ] Improve support for linked ``DyanmicTable`` (e.g, add ``to_hierarchical_dataframe`` function) https://github.com/hdmf-dev/hdmf/pull/645. This PR has already undergone review and is ready to be merged once approved.
 - [ ] **PyNWB**
     - [ ]  The following to PRs implement the integration of the icephys table structure with PyNWB https://github.com/NeurodataWithoutBorders/pynwb/pull/1383 and https://github.com/NeurodataWithoutBorders/pynwb/pull/1349 . The implementation is complete and the PR's are ready for review.

docs/gallery/domain/plot_icephys.py
```diff
@@ -0,0 +1,668 @@
...
+                                 starting_time=123.6,
+                                 rate=20e3,
+                                 electrode=ex_electrode,
+                                 gain=0.02,
+                                 capacitance_slow=100e-12,
+                                 resistance_comp_correction=70.0)
+
+# (A) Add an intracellular recording to the file
+#     NOTE: We can optionally define time-ranges for the stimulus/response via
+#     the corresponding option _start_index and _index_count parameters.
```
rly:
```suggestion
#     the corresponding optional _start_index and _index_count parameters.
```

docs/gallery/domain/plot_icephys.py
```diff
@@ -0,0 +1,703 @@
...
+# Add the table as a new category to our intracellular_recordings
+nwbfile.intracellular_recordings.add_category(category=lab_category)
+# Note, the name of the category is name of the table, i.e., 'recording_lab_data'
+
+#####################################################################
+# .. note:: In an ``AlignedDynamicTable`` all category tables MUST align with the master table,
+#           i.e., all tables must have the same number of rows and rows are expected to
+#           correspond to each other by index
+
+#####################################################################
```
rly, line 420: This section was previously before the above section, but I think it makes more sense after the above section because it references subcategory tables, which are introduced only in the above section.

docs/gallery/domain/plot_icephys_pandas.py
```diff
@@ -0,0 +1,406 @@
...
+# Set pandas rendering option to avoid very wide tables in the html docs
+pandas.set_option("display.max_colwidth", 30)
+pandas.set_option("display.max_rows", 10)
+pandas.set_option("display.max_columns", 6)
+
+#######################
+# Example setup
+# ---------------
+#
+# Generate a simple example NWBFile with dummy intracellular electrophysiology data.
```
rly, line 32:
```suggestion
# Generate a simple example NWBFile with dummy intracellular electrophysiology data.
# This example uses a utility function ``create_icephys_testfile`` to create a
# dummy NWB file with random icephys data.
```

src/pynwb/base.py
```diff
@@ -282,3 +284,85 @@ def __init__(self, **kwargs):
...
+            if vals[0] < 0 or vals[1] < 0:
+                return np.ma.core.MaskedArray(data=vals, mask=[True, True, True])
+            else:
+                return vals
+        else:  # key selected multiple rows
+            # When loading from HDF5 we get an np.ndarray otherwise we get list-of-list. This
+            # makes things the values consistent
+            if isinstance(vals, np.ndarray):
+                vals = [[v[0], v[1], v[2]] for v in vals]
+            vals = np.asarray(vals)
```
rly: If vals is a numpy array, what does this sequence do?

src/pynwb/base.py
```diff
@@ -282,3 +284,85 @@ def __init__(self, **kwargs):
...
+            mask = np.asarray([(v[0] < 0 or v[1] < 0) for v in vals])
+            # For some reason, when we only select 0 or 1 element it is sufficient to mask that element
+            # but when selecting more than 1 element we need to mask all 3 components of each element
+            if len(vals) > 1:
+                mask = np.asarray([v for v in mask for i in range(3)])
+            # If we are masking any elements than create a masked array
+            if np.any(mask):
+                # Depending on whether we select 1 or multiple values we need to for some reason
+                # either define mask values for ea
```
rly:
```suggestion
```

docs/gallery/domain/plot_icephys.py
oruebel: Sounds good

docs/gallery/domain/plot_icephys.py
```diff
@@ -0,0 +1,703 @@
...
+    description='category table for lab-specific recording metadata',
+    colnames=['location', ],
+    columns=[location_column, ]
+)
+# Add the table as a new category to our intracellular_recordings
+nwbfile.intracellular_recordings.add_category(category=lab_category)
+# Note, the name of the category is name of the table, i.e., 'recording_lab_data'
+
+#####################################################################
+# .. note:: In an ``AlignedDynamicTable`` all category tables MUST align with the master table,
```
oruebel:
```suggestion
# .. note:: In an ``AlignedDynamicTable`` all category tables MUST align with the main table,
```

docs/gallery/domain/plot_icephys_pandas.py
```diff
@@ -121,53 +118,54 @@
...
 # -------------------------------------------------------
 #
 
-#######################
+#####################################################################
 # Using nested DataFrames
 # ^^^^^^^^^^^^^^^^^^^^^^^
 # Using the :py:meth:`~hdmf.common.table.DynamicTable.to_dataframe` method we can easily convert tables
-# to pandas  `DataFrame <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.html>`_.
+# to pandas `DataFrames <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.html>`_.
```
oruebel, line 134: @rly note sphinx reuses names of links, i.e., when using `` `DataFrame`_  `` later on, sphinx will create the same link. Changing the link name here to ``DataFrames`` means the link will need to be redefined again later in order to avoid broken links. No big deal, I just wanted to let you know, since this is a bit obscure behavior of Sphinx.

rly: Ah, right. Good catch.

src/pynwb/icephys.py
```diff
@@ -319,15 +338,513 @@ def __get_row_ids(self, sweep_number):
...
+        :param start_index: The start_index provided by the user
+        :param index_count: The index count provided by the user
+        :param time_series: The timeseries object to reference. May be None.
+        :param name: Name of the table. Used only to enhance error reporting
+
+        :raises IndexError: If index_count cannot be determined or start_index+index_count
+            are outside of the range of the timeseries.
+
+        :returns: A tuple of integers with the start_index and index_count to use
```
rly:
```suggestion
        :returns: A tuple of integers with the start_index and index_count to use.
            If time_series is None, then (-1, -1) is returned.
```

src/pynwb/icephys.py
```diff
@@ -319,15 +338,513 @@ def __get_row_ids(self, sweep_number):
...
+        """
+        if time_series is None:
+            return -1, -1
+        start_index = -1 if start_index is None else start_index
+        start_index = start_index if start_index >= 0 else 0
```
rly:
```suggestion
        if start_index is None or start_index < 0:
            start_index = 0
```

src/pynwb/icephys.py
```diff
@@ -319,15 +338,513 @@ def __get_row_ids(self, sweep_number):
...
+        num_samples = time_series.num_samples
+        index_count = -1 if index_count is None else index_count
+        index_count = (index_count
+                       if index_count >= 0
+                       else ((num_samples - start_index)
+                             if num_samples is not None
+                             else None))
```
rly: This structure is difficult for me to read.
```suggestion
        if index_count is None or index_count < 0:
            if num_samples is not None:
                index_count = num_samples - start_index
            else:
                index_count = None
```

src/pynwb/icephys.py
```diff
@@ -319,15 +338,513 @@ def __get_row_ids(self, sweep_number):
...
+        if index_count is None:
+            raise IndexError("Invalid %s_index_count cannot be determined from %s data." % (name, name))
```
rly: This structure is difficult for me to read.
```suggestion
        if index_count is None or index_count < 0:
            if num_samples is not None:
                index_count = num_samples - start_index
            else:
                raise IndexError("Invalid %s_index_count cannot be determined from %s data." % (name, name))
```

src/pynwb/icephys.py
```diff
@@ -319,15 +338,513 @@ def __get_row_ids(self, sweep_number):
...
+        if num_samples is not None:
+            if start_index >= num_samples:
+                raise IndexError("%s_start_index out of range" % name)
+            if (start_index + index_count) > num_samples:
```
rly, line 603: Should this be `>=` ?

tests/unit/test_icephys_metadata_tables.py
```diff
@@ -0,0 +1,1377 @@
...
+class IntracellularElectrodesTableTests(TestCase):
+    """
+    The IntracellularElectrodesTable is covered by the
+    IntracellularRecordingsTableTests as this table is part of that table.
+    """
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
```
rly:
```suggestion
    pass
```

tests/unit/test_icephys_metadata_tables.py
```diff
@@ -0,0 +1,1377 @@
...
+class IntracellularStimuliTableTests(TestCase):
+    """
+    The IntracellularStimuliTable is covered by the
+    IntracellularRecordingsTableTests as this table is part of that table.
+    """
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
```
rly:
```suggestion
	pass
```

tests/unit/test_icephys_metadata_tables.py
```diff
@@ -0,0 +1,1377 @@
...
+class IntracellularResponsesTableTests(TestCase):
+    """
+    The IntracellularResponsesTable is covered by the
+    IntracellularRecordingsTableTests as this table is part of that table.
+    """
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
```
rly:
```suggestion
    pass
```

tests/unit/test_icephys_metadata_tables.py
```diff
@@ -0,0 +1,1377 @@
...
+
+
+class IntracellularRecordingsTableTests(ICEphysMetaTestBase):
+    """
+    Class for testing the IntracellularRecordingsTable Container
+    """
+
+    def test_init(self):
+        _ = IntracellularRecordingsTable()
+        self.assertTrue(True)
```
rly:
```suggestion
        ret = IntracellularRecordingsTable()
        self.assertEqual(ret.name, 'intracellular_recordings')
```

tests/unit/test_icephys_metadata_tables.py
```diff
@@ -0,0 +1,1377 @@
...
+    Test class for testing the SimultaneousRecordingsTable Container class
+    """
+
+    def test_init(self):
+        """
+        Test  __init__ to make sure we can instantiate the SimultaneousRecordingsTable container
+        """
+        ir = IntracellularRecordingsTable()
+        _ = SimultaneousRecordingsTable(intracellular_recordings_table=ir)
+        self.assertTrue(True)
```
rly:
```suggestion
        ir = IntracellularRecordingsTable()
        ret = SimultaneousRecordingsTable(intracellular_recordings_table=ir)
        self.assertIs(ret.recordings.table, ir)
        self.assertEqual(ret.name, 'simultaneous_recordings')
```

tests/unit/test_icephys_metadata_tables.py
```diff
@@ -0,0 +1,1377 @@
...
+    """
+
+    def test_init(self):
+        """
+        Test  __init__ to make sure we can instantiate the SequentialRecordingsTable container
+        """
+        ir = IntracellularRecordingsTable()
+        sw = SimultaneousRecordingsTable(intracellular_recordings_table=ir)
+        _ = SequentialRecordingsTable(simultaneous_recordings_table=sw)
+        self.assertTrue(True)
```
rly:
```suggestion
        ir = IntracellularRecordingsTable()
        sw = SimultaneousRecordingsTable(intracellular_recordings_table=ir)
        ret = SequentialRecordingsTable(simultaneous_recordings_table=sw)
        self.assertIs(ret.simultaneous_recordings.table, sw)
        self.assertEqual(ret.name, 'sequential_recordings')
```

tests/unit/test_icephys_metadata_tables.py
```diff
@@ -0,0 +1,1377 @@
...
+
+    def test_init(self):
+        """
+        Test  __init__ to make sure we can instantiate the RepetitionsTable container
+        """
+        ir = IntracellularRecordingsTable()
+        sw = SimultaneousRecordingsTable(intracellular_recordings_table=ir)
+        sws = SequentialRecordingsTable(simultaneous_recordings_table=sw)
+        _ = RepetitionsTable(sequential_recordings_table=sws)
+        self.assertTrue(True)
```
rly:
```suggestion
        ir = IntracellularRecordingsTable()
        sw = SimultaneousRecordingsTable(intracellular_recordings_table=ir)
        sws = SequentialRecordingsTable(simultaneous_recordings_table=sw)
        ret = RepetitionsTable(sequential_recordings_table=sws)
        self.assertIs(ret.sequential_recordings.table, sws)
        self.assertEqual(ret.name, 'repetitions')
```

tests/unit/test_icephys_metadata_tables.py
```diff
@@ -0,0 +1,1377 @@
...
+    def test_init(self):
+        """
+        Test  __init__ to make sure we can instantiate the ExperimentalConditionsTable container
+        """
+        ir = IntracellularRecordingsTable()
+        sw = SimultaneousRecordingsTable(intracellular_recordings_table=ir)
+        sws = SequentialRecordingsTable(simultaneous_recordings_table=sw)
+        repetitions = RepetitionsTable(sequential_recordings_table=sws)
+        _ = ExperimentalConditionsTable(repetitions_table=repetitions)
+        self.assertTrue(True)
```
rly:
```suggestion
        ir = IntracellularRecordingsTable()
        sw = SimultaneousRecordingsTable(intracellular_recordings_table=ir)
        sws = SequentialRecordingsTable(simultaneous_recordings_table=sw)
        repetitions = RepetitionsTable(sequential_recordings_table=sws)
        ret = ExperimentalConditionsTable(repetitions_table=repetitions)
        self.assertIs(ret.repetitions.table, repetitions)
        self.assertEqual(ret.name, 'experimental_conditions')
```

tests/unit/test_icephys_metadata_tables.py
```diff
@@ -0,0 +1,1377 @@
...
+            gain=0.02,
+            capacitance_slow=100e-12,
+            resistance_comp_correction=70.0,
+            sweep_number=np.uint64(15))
+
+    def test_init(self):
+        """
+        Test  __init__ to make sure we can instantiate the NWBFileTests container
+        """
+        _ = self.__get_icephysfile()
```
rly: Not needed anymore
```suggestion
```

src/pynwb/base.py
oruebel: @rly I have to admit, this is a bit of a tricky problem. The case ``if isinstance(vals, np.ndarray)`` occurs when the data is being read from disk. In this case ``self.data`` will be a ``hdmf.backends.hdf5.h5_utils.ContainerH5TableDataset`` and the data is read from file as a numpy structured array. In contrast, when we create the file, the compound data is treated as a list of tuples. This means that the structure of the data is quite different in those  two cases.  So when calling ``np.asarray(vals)`` we get:
* A numpy array with ``dtype=object`` and ``shape=(1,3)`` when ``self.data`` is a list of tuples
* A numpy array with ``dtype=[('idx_start', '<i4'), ('count', '<i4'), ('timeseries', 'O')] `` and ``shape=(1,)`` when ``self.data`` is read from file. 

This leads to two key issues:

* Dealing with the difference in shape (i.e., 1D vs 2D array) downstream is annoying
* Dealing with the difference in dtype downstream is annoying
* Conversion to pandas DataFrame fails with the structured dtype with the error shown below.

In part because of the latter issue I had opted for converting the data on read to a list of tuples to keep things consistent. I agree that this should be documented more clearly in the code to avoid confusion later on.

How does VectorData usually deal with compound types? Are they treated as 2D arrays or as 1D structured arrays? Ideally we should try to keep the behavior here consistent with how other Container classes treat compound datatypes. If we should use a structured array instead (which may be more intuitive) it would require: 1) converting ``vals`` to a structured array if it is a list of tuples (essentially the opposite of the current logic) and 2) require dealing with the conversion to pandas issues downstream. I'm not sure yet what it will take to deal with the conversion issue. Pandas appears to attempt a reduction on the masked array which fails to deal with the complex dtype. 

What are your thoughts? What should the right behavior be?

```
ERROR: test_add_row_no_stimulus (test_icephys_metadata_tables.IntracellularRecordingsTableTests)
  File "[...]/pynwb/tests/unit/test_icephys_metadata_tables.py", line 316, in test_add_row_no_stimulus
    self.write_test_helper(ir)
  File “[…]/hdmf/src/hdmf/utils.py", line 583, in func_call
    return func(args[0], **pargs)
  File "[...]/pynwb/tests/unit/test_icephys_metadata_tables.py", line 125, in write_test_helper
    assert_frame_equal(ir.to_dataframe(**to_dataframe_kwargs), in_ir.to_dataframe(**to_dataframe_kwargs))
  File "[...]/hdmf/src/hdmf/utils.py", line 583, in func_call
    return func(args[0], **pargs)
  File "[...]/pynwb/src/pynwb/icephys.py", line 616, in to_dataframe
    res = super().to_dataframe(ignore_category_ids=getargs('ignore_category_ids', kwargs))
  File "[...]/hdmf/src/hdmf/utils.py", line 583, in func_call
    return func(args[0], **pargs)
  File "[...]/hdmf/src/hdmf/common/alignedtable.py", line 246, in to_dataframe
    dfs += [category.to_dataframe().reset_index() for category in self.category_tables.values()]
  File "[...]/hdmf/src/hdmf/common/alignedtable.py", line 246, in <listcomp>
    dfs += [category.to_dataframe().reset_index() for category in self.category_tables.values()]
  File "[...]/hdmf/src/hdmf/utils.py", line 583, in func_call
    return func(args[0], **pargs)
  File "[...]/hdmf/src/hdmf/common/table.py", line 1055, in to_dataframe
    ret = self.__get_selection_as_df(sel)
  File "[...]/hdmf/src/hdmf/common/table.py", line 966, in __get_selection_as_df
    ret = pd.DataFrame(df_input, index=pd.Index(name=self.id.name, data=id_index))
  File "[...]/anaconda3/envs/icephys-port/lib/python3.8/site-packages/pandas-1.3.0-py3.8-macosx-10.9-x86_64.egg/pandas/core/frame.py", line 614, in __init__
    mgr = dict_to_mgr(data, index, columns, dtype=dtype, copy=copy, typ=manager)
  File "[...]/anaconda3/envs/icephys-port/lib/python3.8/site-packages/pandas-1.3.0-py3.8-macosx-10.9-x86_64.egg/pandas/core/internals/construction.py", line 462, in dict_to_mgr
    return arrays_to_mgr(
  File "[...]/anaconda3/envs/icephys-port/lib/python3.8/site-packages/pandas-1.3.0-py3.8-macosx-10.9-x86_64.egg/pandas/core/internals/construction.py", line 122, in arrays_to_mgr
    arrays = _homogenize(arrays, index, dtype)
  File "[...]/anaconda3/envs/icephys-port/lib/python3.8/site-packages/pandas-1.3.0-py3.8-macosx-10.9-x86_64.egg/pandas/core/internals/construction.py", line 577, in _homogenize
    val = sanitize_array(
  File "[...]/anaconda3/envs/icephys-port/lib/python3.8/site-packages/pandas-1.3.0-py3.8-macosx-10.9-x86_64.egg/pandas/core/construction.py", line 497, in sanitize_array
    data = sanitize_masked_array(data)
  File "[...]/anaconda3/envs/icephys-port/lib/python3.8/site-packages/pandas-1.3.0-py3.8-macosx-10.9-x86_64.egg/pandas/core/construction.py", line 454, in sanitize_masked_array
    if mask.any():
  File "[...]/anaconda3/envs/icephys-port/lib/python3.8/site-packages/numpy/core/_methods.py", line 45, in _any
    return umr_any(a, axis, dtype, out, keepdims)
TypeError: cannot perform reduce with flexible type

```

rly: In my opinion, compound-data-type data should always be stored in memory as a numpy structured array. A tuple (or list of tuples) is hard to interpret without the field names. This would also help move us away from building functionality around lists and tuples. We could define `TimeSeriesReferenceVectorData.add_row` to take in a tuple and convert it to the correct structured array element (or take in a structured array object and check its fields) before appending it to the internal numpy structured array. 

And to answer your question about VectorData and compound types, for compound data not defined in the schema, see https://github.com/hdmf-dev/hdmf/issues/655. If the compound type is defined in the schema, then before write, I believe it is  retrieved as a list of tuples, and after write, it is retrieved as a 1D numpy structured array.

rly:
I made some minor text edits throughout (it seemed tedious to mark them all as issues that require your attention, especially typos), reformatted the code a bit, and made some suggestions to code. Please review my commits. All those aside, this looks good to me!

rly:
There are two errors in `gallery38`. One is due to the allensdk not being compatible with h5py 3 yet. The other is:
```
2021-07-28 00:59:06,514 - INFO - Executing docs/gallery/domain/plot_icephys_pandas.py
ExperimentalConditionsTable
  File "test.py", line 82, in run_example_tests
    _import_from_file(script)
  File "test.py", line 61, in _import_from_file
    return imp.load_source(os.path.basename(script), script)
  File "/usr/local/lib/python3.8/imp.py", line 171, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 702, in _load
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "docs/gallery/domain/plot_icephys_pandas.py", line 104, in <module>
    print("Has Foreign Columns:", root_table.has_foreign_columns())
AttributeError: 'ExperimentalConditionsTable' object has no attribute 'has_foreign_columns'
```

oruebel:
> AttributeError: 'ExperimentalConditionsTable' object has no attribute 'has_foreign_columns'

This is because the changes to support the functions for hierarchical tables in HDMF have not been released yet. If you run the galleries against the HDMF dev branch locally, then things should work.

rly:
> > AttributeError: 'ExperimentalConditionsTable' object has no attribute 'has_foreign_columns'
> 
> This is because the changes to support the functions for hierarchical tables in HDMF have not been released yet. If you run the galleries against the HDMF dev branch locally, then things should work.

Oh, right. I forgot.

src/pynwb/base.py
oruebel: I'm in favor of using structured array here. If we should use the structured array type, do you have any idea about the pandas conversion failure in that case?

rly: From the stack trace, it seems like the issue is with the fact that the structured array is masked. I'm digging into it.

rly: I believe this is a not-yet-resolved issue in numpy: https://github.com/numpy/numpy/issues/7492 
that is not addressed by pandas: https://github.com/pandas-dev/pandas/issues/42739 (note there is a concerning comment there about masked records being deprecated)

We can get around it by calling `.filled()` on the masked array before passing it to `pd.DataFrame(...)` in `DynamicTable.__get_selection_as_df` but then `IntracellularRecordingsTable.to_dataframe` needs to be updated because you will no longer get a masked array in the dataframe, but a normal array with a fillvalue for masked data.

rly: Numpy/pandas might be able to resolve this issue, but it would not help us unless we want to force users to use only the latest numpy/pandas version.

rly: See https://github.com/hdmf-dev/hdmf/pull/658 for a fix in `DynamicTable.__get_selection_as_df`

oruebel: I guess the question is whether we should look for an alternative to using numpy masked array here. I think masked arrays are a convenient way to describe missing data values, but if they cause too much trouble, then we could also opt for an alternative representation. E.g., if we always used list of tuples then we could use ``(None, None, None)`` for masked elements.

oruebel: If we end up filling the masked array anyways, then it seems it would probably make more sense to either do this here directly (to avoid inconsistency between what a user sees here and when converting to pandas) or to change the representation for missing values altogether. I'm honestly not sure what ``filled`` will use as fill value in this case.

oruebel: Based on the fact that pandas doesn't support the masked array behavior we decided to instead return lists of named tuples with masked values being named tuples of ``(None, None, None)``. The named tuple would look something like:

```
from collections import namedtuple
TimeSeriesReference = namedtuple('TimeSeriesReference', ['idx_start', 'count', 'TimeSeries])
```

src/pynwb/icephys.py
oruebel: I updated this function to clarify the logic and add comments.

src/pynwb/icephys.py
oruebel: ``>`` should be correct here. E.g., if we have a TimeSeries of length 10  (i.e., ``num_samples=10``) and we want to select the full series, then we should have ``start_index=0`` and ``index_count=10``, since we want to start at the beginning and select all ``10`` values. In this case ``start_index + index_count == num_sample``. The same happens any time we want to select all values after ``start_index``.

src/pynwb/icephys.py
oruebel: I simplified the logic in the latest commit and added comments.

src/pynwb/base.py
oruebel: The following commit https://github.com/NeurodataWithoutBorders/pynwb/pull/1349/commits/79271f90c8712cc2da4c5ae52bd91d5d0fad041c implements the changes we had discussed to use lists of ``namedtuple`` instead of numpy masked arrays.

src/pynwb/base.py
```diff
@@ -286,25 +287,50 @@ def __init__(self, **kwargs):
...
+class TimeSeriesReference(namedtuple('TimeSeriesReference', ['idx_start', 'count', 'timeseries'])):
+    """
+    Class used to represent data values of a :py:class:`~pynwb.base.TimeSeriesReferenceVectorData`
+    This is a ``namedtuple('TimeSeriesReference', ['idx_start', 'count', 'timeseries'])`` type.
+    """
+    pass
+
+
 @register_class('TimeSeriesReferenceVectorData', CORE_NAMESPACE)
 class TimeSeriesReferenceVectorData(VectorData):
```
rly, line 413: Can we override `add_row` and `append` to restrict the type of the input to either a TimeSeriesReference or a tuple with length 3 and the correct data types?

It is currently possible to do:
```python
v = pynwb.base.TimeSeriesReferenceVectorData(name='a', description='a')
v.add_row(1)
v.append(1)
```

rly: Closing to comment on the PR instead of the commit.

src/pynwb/base.py
```diff
@@ -282,3 +285,107 @@ def __init__(self, **kwargs):
...
+            return re
+
+    def extend(self, ar):
+        """
+        The extend_data method adds all the elements of the iterable arg to the
+        end of the data of this Data container.
+
+        :param arg: The iterable to add to the end of this VectorData
+        """
+        self.data = extend_data(self.data, ar)
```
rly: Is this necessary? Also it doesn't work. 

```python
>>> v = pynwb.base.TimeSeriesReferenceVectorData(name='a', description='a')
>>> v.extend([1])
  File "<stdin>", line 1, in <module>
  File "c:\users\ryan\documents\nwb\pynwb\src\pynwb\base.py", line 391, in extend
    self.data = extend_data(self.data, ar)
AttributeError: can't set attribute
```

rly: Removing the code and using `VectorData.extend` seems to work just fine.

src/pynwb/base.py
oruebel: Done in https://github.com/NeurodataWithoutBorders/pynwb/pull/1349/commits/a58921aca2401c9c809b00a7bec47bf30d977558

src/pynwb/base.py
oruebel: Fixed in https://github.com/NeurodataWithoutBorders/pynwb/pull/1349/commits/a58921aca2401c9c809b00a7bec47bf30d977558

src/pynwb/base.py
```diff
@@ -287,12 +286,124 @@ def __init__(self, **kwargs):
...
+        :returns: True if successful. If unsuccessful `TypeError` is raised.
+
+        :raises TypeError: If one of the fields does not match the expected type
+        """
+        if not isinstance(self.idx_start, (int, np.integer)):
+            raise TypeError("idx_start must be an integer not %s" % str(type(self.idx_start)))
+        if not isinstance(self.count, (int, np.integer)):
+            raise TypeError("count must be an integer %s" % str(type(self.count)))
+        if not isinstance(self.timeseries, TimeSeries):
+            raise TypeError("timeseries must of of type TimeSeries. %s" % str(type(self.timeseries)))
```
rly:
```suggestion
            raise TypeError("timeseries must of type TimeSeries. %s" % str(type(self.timeseries)))
```

src/pynwb/base.py
```diff
@@ -287,12 +286,124 @@ def __init__(self, **kwargs):
...
+        return True
+
+    def isvalid(self):
+        """
+        Check whether the reference is valid. Setting :py:meth:`~pynwb.base.TimeSeriesReference.idx_start` and
+        :py:meth:`~pynwb.base.TimeSeriesReference.count` to -1 is used to indicate invalid references. This is
+        useful to allow for missing data in :py:class:`~pynwb.base.TimeSeriesReferenceVectorData`
+
+        :returns: True if the selection is valid. Returns False if bot
```
rly:
```suggestion
        :returns: True if the selection is valid. Returns False if both
```

src/pynwb/base.py
```diff
@@ -287,12 +286,124 @@ def __init__(self, **kwargs):
...
+                raise IndexError("'count' %i invalid. 'count' must be positive" % self.count)
+            if (self.idx_start + self.count) > num_samples:
+                raise IndexError("'idx_start + count' out of range for timeseries '%s'" % self.timeseries.name)
+        return True
+
+    @property
+    def timestamps(self):
+        """
+        Get the floating point timestamp offsets in seconds from the timeseries that correspond to the array.
+        These are either loaded directlyu from the :py:meth:`~pynwb.base.TimeSeriesReference.timeseries`
```
rly:
```suggestion
        These are either loaded directly from the :py:meth:`~pynwb.base.TimeSeriesReference.timeseries`
```

src/pynwb/base.py
```diff
@@ -287,12 +286,124 @@ def __init__(self, **kwargs):
...
+        or calculated from the
```
rly:
```suggestion
        timestamps or calculated from the starting time and sampling rate.
```

src/pynwb/base.py
```diff
@@ -287,12 +286,124 @@ def __init__(self, **kwargs):
...
+        :raises IndexError: If the combination of :py:meth:`~pynwb.base.TimeSeriesReference.idx_start` and
+            :py:meth:`~pynwb.base.TimeSeriesReference.count` are not valid for the given timeseries.
+
+        :raises TypeError: If one of the fields does not match the expected type
+
+        :returns: Array with the timestamps.
+        """
+        self.isvalid()
+        if self.idx_start < 0 and self.count < 0:
+            return None
```
rly:
```suggestion
        if not self.isvalid():
            return None
```

src/pynwb/base.py
```diff
@@ -287,12 +286,124 @@ def __init__(self, **kwargs):
...
+        :py:meth:`~pynwb.base.TimeSeriesReference.idx_start` and
+        :py:meth:`~pynwb.base.TimeSeriesReference.count`
+
+        :raises IndexError: If the combination of :py:meth:`~pynwb.base.TimeSeriesReference.idx_start` and
+            :py:meth:`~pynwb.base.TimeSeriesReference.count` are not valid for the given timeseries.
+
+        :raises TypeError: If one of the fields does not match the expected type
+
+        :returns: Result of ``self.timeseries.data[self.idx_start: (self.idx_start + self.count)]``. Returns
+            None in case the reference is invalid (i.e,. if both :py:meth:`~pynwb.base.TimeSeriesReference.idx_start`
```
rly:
```suggestion
            None in case the reference is invalid (i.e., if both :py:meth:`~pynwb.base.TimeSeriesReference.idx_start`
```

src/pynwb/base.py
```diff
@@ -287,12 +286,124 @@ def __init__(self, **kwargs):
...
+            and :py:meth:`~pynwb.base.TimeSeriesReference.count` are negative.
+        """
+        self.isvalid()
+        if self.idx_start < 0 and self.count < 0:
+            return None
```
rly:
```suggestion
        if not self.isvalid():
            return None
```

tests/unit/test_base.py
```diff
@@ -328,3 +328,117 @@ def test_get_length5_with_invalid_data(self):
...
+        with self.assertRaisesWith(TypeError, "idx_start must be an integer not <class 'float'>"):
+            tsr.check_types()
+        # invalid types, use float instead of int for count only
+        tsr = TimeSeriesReference(1, 5.0, TimeSeries(name='test'+str(0), description='test',
+                                                     data=np.arange(10), starting_time=5.0, rate=0.1))
+        with self.assertRaisesWith(TypeError, "count must be an integer <class 'float'>"):
+            tsr.check_types()
+        # invalid type for TimeSeries but valid idx_start and count
+        tsr = TimeSeriesReference(1, 5, None)
+        with self.assertRaisesWith(TypeError, "timeseries must of of type TimeSeries. <class 'NoneType'>"):
```
rly:
```suggestion
        with self.assertRaisesWith(TypeError, "timeseries must of type TimeSeries. <class 'NoneType'>"):
```

src/pynwb/base.py
oruebel: Will be resolved in upcoming commit since I also need to update tests to match.

tests/unit/test_base.py
oruebel: Already resolved in commit.

rly (approved):

rly:
This is good to merge into rc/2.0.0. Great work!

PR #1366, opened 2021-06-09
Fix integration tests with invalid test data that will be caught by future hdmf validator version

dsleiter:
Fix integration tests with invalid test data that will begin to fail with an update to hdmf validator that increases validation coverage.

See the discussion here: https://github.com/hdmf-dev/hdmf/pull/609#issuecomment-857418760

* Ref https://github.com/hdmf-dev/hdmf/issues/585
* Ref https://github.com/hdmf-dev/hdmf/pull/609

## Motivation

Integration tests which use `TestElectricalSeriesIO.make_electrode_table` are failing in the pynwb CI tests for https://github.com/hdmf-dev/hdmf/pull/609. That PR is fixing a validator issue which caused the validator to not validate certain builders. With that fix, the validator is now correctly validating many builders which were not previously being validated, including in unit tests.

This PR resolves item 3 in [the discussion for hdmf PR#609](https://github.com/hdmf-dev/hdmf/pull/609#issuecomment-857418760)

The issue is that when adding rows to an electrode table in `make_electrode_table`, [the value for `x` in `table.add_row` is provided as an int](https://github.com/NeurodataWithoutBorders/pynwb/blob/c1f4bd76931e0a9179cc4591807789e49394547a/tests/integration/hdf5/test_ecephys.py#L42-L44) while [the schema expects a float](https://nwb-schema.readthedocs.io/en/latest/format.html?highlight=root%2Fintervals#groups-general-extracellular-ephys-electrodes).

## How to test the behavior?
1. checkout the dev branch for pynwb
2. checkout and pip install the hdmf branch from this PR: https://github.com/hdmf-dev/hdmf/pull/609
3. execute the ecephys integration tests (not using tox, since tox would manage the hdmf dependency):
  ```
  $ python -m unittest tests/integration/hdf5/test_ecephys.py 
  ```
  you should see 13 test failures with the following message:
  ```
  Exception: root/general/extracellular_ephys/electrodes/x (general/extracellular_ephys/electrodes/x): incorrect type - expected 'float32', got 'int64'
  ```
4. checkout the pynwb branch for this PR (or just manually apply the changes from here) and rerun the same tests. 13 tests should fail again, however this time they should fail for a different reason:
  ```
  Exception: root/general/extracellular_ephys/electrodes/filtering (general/extracellular_ephys/electrodes/filtering): incorrect type - expected 'float32', got 'utf'
  ```
  This failure is due to item 4 [the discussion for hdmf PR#609](https://github.com/hdmf-dev/hdmf/pull/609#issuecomment-857418760), and is due to what appears to be a disagreement between nwb-schema and pynwb, and will not be fixed just by unit tests.

Note: I understand we can't merge in failing tests. However, with the current hdmf version these tests will continue to pass. HDMF PR#609 won't be merged in until we resolve the discrepancy in the dtype of the electrode table `filtering` column, so these changes are safe to merge in here now.

## Checklist


tests/integration/test_io.py
```diff
@@ -275,7 +275,7 @@ def test_append(self):
             self.assertIs(nwb.processing['test_proc_mod']['LFP'].electrical_series['test_es'].electrodes,
                           nwb.acquisition['timeseries2'].electrodes)
             errors = validate(io)
-            self.assertTrue(len(errors) == 0)
+            self.assertEqual(len(errors), 0, errors)
```
dsleiter, line 278: Note: this change doesn't change any logic, but just displays what the validation errors are if any, making it easier to debug. (with the updated hdmf validator, this test fails due to the incompatible type for the filtering column of the electrodes table, but this test will currently still pass with the current dev version of hdmf.)

t-b (dismissed):

oruebel (dismissed):

tests/integration/test_io.py
rly: Sounds good. I didn't know `assertEqual` had that functionality

rly (dismissed):

rly (approved):

PR #1374, opened 2021-06-18
make MotionCorrection test

### LINKED ISSUE #1373: MotionCorrection not working
The init requires arg named "`corrected_images_stacks`", but this is not correctly mapped on write:

```python
~/opt/anaconda3/lib/python3.7/site-packages/hdmf/utils.py in func_call(*args, **kwargs)
    578             def func_call(*args, **kwargs):
    579                 pargs = _check_args(args, kwargs)
--> 580                 return func(args[0], **pargs)
    581         else:
    582             def func_call(*args, **kwargs):

~/opt/anaconda3/lib/python3.7/site-packages/hdmf/build/objectmapper.py in get_attr_value(self, **kwargs)
    550                 msg = ("%s '%s' does not have attribute '%s' for mapping to spec: %s"
    551                        % (container.__class__.__name__, container.name, attr_name, spec))
--> 552                 raise ContainerConfigurationError(msg)
    553             if attr_val is not None:
    554                 attr_val = self.__convert_string(attr_val, spec)

ContainerConfigurationError: MotionCorrection 'MotionCorrection' does not have attribute 'corrected_image_stacks' for mapping to spec: {'doc': 'Reuslts from motion correction of an image stack.', 'quantity': '+', 'neurodata_type_inc': 'CorrectedImageStack'}
```

the arg name should be `corrected_image_stacks`

bendichter:
fix #1373 

Attempt to make MotionCorrection test. I don't understand why it's failing

### DIFF
--- a/src/pynwb/ophys.py
+++ b/src/pynwb/ophys.py
@@ -193,7 +193,7 @@ class MotionCorrection(MultiContainerInterface):
         'get': 'get_corrected_image_stack',
         'create': 'create_corrected_image_stack',
         'type': CorrectedImageStack,
-        'attr': 'corrected_images_stacks'
+        'attr': 'corrected_image_stacks'
     }
 
 
--- a/src/pynwb/ophys.py
+++ b/src/pynwb/ophys.py
@@ -161,8 +161,8 @@ class CorrectedImageStack(NWBDataInterface):
     assumed to be 2-D (has only x & y dimensions).
     """
 
-    __nwbfields__ = ({'name': 'corrected', 'child': True},
-                     {'name': 'xy_translation', 'child': True},
+    __nwbfields__ = ({'name': 'corrected', 'child': True, 'required_name': 'corrected'},
+                     {'name': 'xy_translation', 'child': True, 'required_name': 'xy_translation'},
                      'original')
 
     @docval({'name': 'name', 'type': str,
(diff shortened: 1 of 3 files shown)


bendichter:
@rly please help?

rly:
The root problem is that the schema says that the schema for `CorrectedImageStack` says that the name of the child `ImageSeries` has to be "corrected", but the API class for `CorrectedImageStack` allows any name for the child `ImageSeries`. I updated the PR to add a check for this. 

This feels clunky, like we should not have this gotcha when defining schema. I know Akshay ran into this with his extensions, where the names have to match in order for the data to be read and validated correctly. In some cases, it is better not to enforce a name on the container containing a child neurodata type, but names are helpful and some data types like `CorrectedImageStack` or let's say a type that contains two specific `TimeSeries` need to use the name to disambiguate.

oruebel:
> The root problem is that the schema says that the schema for `CorrectedImageStack` says that the name of the child `ImageSeries` has to be "corrected", but the API class for `CorrectedImageStack` allows any name for the child `ImageSeries`. I updated the PR to add a check for this

I think the `required_name` approach is appropriate for this. Similar to how we warn on data-type changes we could here also fix the name on write to be compliant with the schema and issue a "InvaldiContainerName" warning.

bendichter:
Thanks @rly ! This looks good to me!

rly (approved):

PR #1382, opened 2021-07-09
doc: fix putative typo second/minute

jcolomb:
not sure about the link, it may break if included like this, what would you consider better?

## Motivation

error between text and unit given (m - seconds).

### DIFF
--- a/docs/gallery/general/file.py
+++ b/docs/gallery/general/file.py
@@ -42,17 +42,18 @@ from pynwb import TimeSeries
 
 data = list(range(100, 200, 10))
 timestamps = list(range(10))
-test_ts = TimeSeries(name='test_timeseries', data=data, unit='m', timestamps=timestamps)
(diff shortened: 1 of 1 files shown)


bendichter:
@jcolomb thanks for the feedback, but the tutorial is correct. It's a little tricky, let me explain. 

A `TimeSeries` object stores a signal that changes over time, like a voltage that is sampled regularly. The data stored in this object is a temporal sequence of measurements. The times that those samples are collected is stored either as  a starting time and sampling rate or a sequence of timestamps. If timestamps are used, they will be in seconds, as you suggest, and this is not changeable by the user. In theory it would have been possible for us to allow other units of time like hours and days but we chose to just stick with seconds for consistency. The `units` here are the units of measurement of the data, which is "volts" in my example and is "meters" in the tutorial.

I think your confusion here might highlight that we can do a better job of explaining that a `TimeSeries` object does not just store time information, but also data that are associated with those times.

Issue #1367, opened 2021-06-09
Create short tutorial on data streaming with ros3


oruebel:
It would be useful to create a short tutorial to illustrate how one can stream data from S3 using the latest HDMF/PyNWB. 

DANDI example code (from Ben Dichter)

```
## test out streaming
from pynwb import NWBHDF5IO
from nwbwidgets import nwb2widget
import requests

def get_s3_url(url):
    s3_url = requests.request(url=url, method='head').url
    return s3_url[:s3_url.index('?')]

url = "https://api.dandiarchive.org/api/dandisets/000003/versions/draft/assets/4112d7b0-071e-4193-bd30-73d52e7576c6/download/"

s3_path = get_s3_url(url)

io = NWBHDF5IO(s3_path, mode='r', load_namespaces=True, driver='ros3')
nwb = io.read()
nwb2widget(nwb)

# Use https://api.dandiarchive.org/api/dandisets/000003/versions/draft/assets/ too look up 
# assests (substituting 000003 for any dandiset ID) and use the asset_id for any NWB file.
```

**Additional context**
Related also to #1319 

## Checklist

- [X] Have you ensured the feature or change was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues)?
- [X] Have you included a brief and descriptive title?
- [X] Have you included a clear description of the problem you are trying to solve?
- [X] Have you included a minimal code snippet that reproduces the issue you are encountering?
- [X] Have you checked our [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?

oruebel:
Fixed in #1393

PR #1393, opened 2021-08-11
Test ROS3 streaming in CI and add ROS3 tutorial

### LINKED ISSUE #1392: Add tests for ROS3 streaming support
## Description

PyNWB 2.0 supports streaming with the HDF5 ROS3 driver. We should test this in CI.

## Checklist

- [x] Have you ensured the bug was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues)?

rly:
## Motivation

Resolve #1392.

## How to test the behavior?
```
python test.py -r
```

## Checklist


rly:
I added a ROS3 streaming test to the CI. The test pulls a small file from DANDI. Please review the test to make sure it makes sense. I hope this doesn't put too much of a burden on DANDI.

The CI runs on CircleCI (linux) on every push and nightly, and it runs on Azure (Windows, MacOS) nightly.

tests/integration/ros3/test_ros3.py
```diff
@@ -0,0 +1,52 @@
...
+def get_dandi_s3_url(dandiset_id, filepath):
+    """Get the s3 location for any NWB file on DANDI"""
+
+    asset_id = get_dandi_asset_id(dandiset_id, filepath)
+    url = f"https://api.dandiarchive.org/api/dandisets/{dandiset_id}/versions/draft/assets/{asset_id}/download/"
+
+    s3_url = requests.request(url=url, method='head').url
+    if '?' in s3_url:
+        return s3_url[:s3_url.index('?')]
+    return s3_url
```
oruebel: I think those functions may make sense to put into ``pynwb.testing`` instead

tests/integration/ros3/test_ros3.py
```diff
@@ -0,0 +1,52 @@
...
+
+class TestRos3Streaming(TestCase):
+    # requires h5py to be built with the ROS3 driver: conda install -c conda-forge h5py
+    # also requires requests package to be installed
+
+    def test_read(self):
+        # ephys, Svoboda Lab (450 kB)
+        dandiset_id = '000006'
+        filepath = 'sub-anm372795/sub-anm372795_ses-20170718.nwb'
+        s3_path = get_dandi_s3_url(dandiset_id, filepath)
```
oruebel: Rather than using an actual user asset, I think it would make sense for DANDI to have an S3 bucket with small test files. This is both to make this easier to control but also because running tests against actual data will throw off usage stats for DANDI.

oruebel: Having an S3 test bucket on DANDI would also allow testing against NWB files with different versions (as part of a nightly test).

rly:
@oruebel I added a short tutorial on how to do streaming. Feel free to add to it.

Fix #1367.

tests/integration/ros3/test_ros3.py
rly: Yeah, I agree. Might be useful for others when writing tests for extensions and such.

rly: They might also be useful for people who want to stream data from DANDI, in which case it should not be in `pynwb.testing` but maybe a new module `pynwb.utils`.

rly: I have now done that

tests/integration/ros3/test_ros3.py
rly: The test now points to `https://dandiarchive.s3.amazonaws.com/ros3test.nwb` assuming that @yarikoptic or someone else on the DANDI team can host that.

rly: Assuming this is an easy thing for the DANDI team to do, I'd like to hold the release until tomorrow so that we can get this PR in. @oruebel would that be all right?

yarikoptic: which file we should host?  I would prefer if you provide one so it is not encumbered by any possible fiasco with user wishes/license/suboptimal nwb version etc.  Just create one (or some) you would like to see representative of the features you would gain with ROS3 (not fetching "large" data arrays etc) and share and we will gladly host it/them

yarikoptic: BTW, might be worth testing that you/h5py can access it through a redirect. E.g. try on https://api.dandiarchive.org/api/assets/1c095f5f-d1e2-45db-b807-fdcfea08c6de/download/

as long as our API is stable, we (you) can actually even create a dedicated pynwb testing dandiset and  release it on dandiarchive and get a DOI and "guaranteed" availability in the future.  If we do not mess up -- it will be truly "guaranteed"! ;)

rly: @yarikoptic woops, sorry. I sent you the file on Slack and it looks like @satra took care of it!

rly: > as long as our API is stable, we (you) can actually even create a dedicated pynwb testing dandiset and release it on dandiarchive and get a DOI and "guaranteed" availability in the future. If we do not mess up -- it will be truly "guaranteed"! ;)

would you prefer us to do that? it seems like it would clutter dandi, but it would be a nice test

oruebel:
> I added a short tutorial on how to do streaming. Feel free to add to it.

Thanks! Looks good. I created a thumbnail for the tutorial.

docs/gallery/advanced_io/streaming.py
```diff
@@ -0,0 +1,51 @@
...
+# will download metadata about the file from the S3 bucket to memory. Datasets are accessed lazily, just like when
+# reading an NWB file stored locally. So, slicing into a dataset will require additional time to download the sliced
+# data to memory.
+#
+# .. code-block:: python
+#
+#   from pynwb import NWBHDF5IO
+#
+#   with NWBHDF5IO(s3_path, mode='r', load_namespaces=True, driver='ros3') as io:
+#       nwbfile = io.read()
```
oruebel, line 50: Is there a reason for using a context-manager here on read. I'm just wondering since we normally tend to use contexts only on write.

tests/integration/ros3/test_ros3.py
yarikoptic: Dandi is there to be useful, I wouldn't mind at all, and I doubt @satra will. Can even generate/upload as many as you like for didn't versions of schema etc as long as they would be useful, and Publish to make it last ;-)

Edit: I don't expect them be of a file size worth noting ;-)

docs/gallery/advanced_io/streaming.py
rly: That is a good question. I had always followed the rule of thumb to use a context manager when opening a file to be safe. It should not make a difference in virtually all cases - the file will be closed on garbage collection and having the file open doesn't stop other processes from reading or modifying the file. However, if someone is using Windows and going to delete the file before garbage collection (or thinks another process might delete it), then the file must be explicitly closed before it can be deleted. So I think to be safe, it is ever so slightly better to use a context manager.

```python
import h5py
import os

filename = 'nwbfile.nwb'
f = h5py.File(filename, 'r')
os.remove(filename)  # fails on windows
``` 

See also:
https://stackoverflow.com/questions/56149237/do-i-need-to-manually-close-a-hdf5-file
https://stackoverflow.com/questions/7395542/is-explicitly-closing-files-important/7395906#7395906

tests/integration/ros3/test_ros3.py
rly: > BTW, might be worth testing that you/h5py can access it through a redirect.

FYI this does not work. 

```
Python 3.9.6 | packaged by conda-forge | (default, Jul 11 2021, 03:37:25) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import h5py
>>> h5py.File('https://api.dandiarchive.org/api/assets/902af029-f411-40b5-81f1-67930d56ee05/download/', 'r', driver='ros3')
  File "<stdin>", line 1, in <module>
  File "C:\Users\Ryan\miniconda3\envs\dev\lib\site-packages\h5py\_hl\files.py", line 444, in __init__
    fid = make_fid(name, mode, userblock_size,
  File "C:\Users\Ryan\miniconda3\envs\dev\lib\site-packages\h5py\_hl\files.py", line 199, in make_fid
    fid = h5f.open(name, flags, fapl=fapl)
  File "h5py\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py\h5f.pyx", line 100, in h5py.h5f.open
OSError: Unable to open file (could not convert found "Content-Length" response ("0"))
```

rly: In this PR, we add a utility function that uses a `requests.request` call to get around the redirect.

When constructing the asset URL, which of these is more stable / preferred (or does it not matter)?
f'https://api.dandiarchive.org/api/dandisets/{dandiset_id}/versions/draft/assets/{asset_id}/download/'
f'https://api.dandiarchive.org/api/assets/{asset_id}/download/'

src/pynwb/utils/dandi.py
```diff
@@ -0,0 +1,35 @@
...
+    import requests
+
+    asset_id = get_dandi_asset_id(dandiset_id, filepath)
+    url = f'https://api.dandiarchive.org/api/dandisets/{dandiset_id}/versions/draft/assets/{asset_id}/download/'
+    # another form is https://api.dandiarchive.org/api/assets/{asset_id}/download/
+
+    s3_url = requests.request(url=url, method='head').url  # this gets past the redirect
+    if '?' in s3_url:
+        return s3_url[:s3_url.index('?')]
+    return s3_url
```
bendichter: There's a better way to do this now. I'll push an update

tests/integration/ros3/test_ros3.py
yarikoptic: the hope is -- "not matter" ;-)

re redirect -- sorry, didn't know, and allowing for HEAD requests was resolved only recently: https://github.com/dandi/dandi-api/issues/380 ... so we might need to go all the way to HDF5 lib to see what they would say about redirects.  Sorry for leading you the rabbit hole ;)

bendichter:
@rly you used functions that I wrote a while back before these features were exposed in the dandi python package. Now that those tools are available, it is much simpler to use them. I pushed the change. What do you think?

tests/integration/ros3/test_ros3.py
satra: @rly - nothing is stable till you publish. but once published then you should expect some stability. 

regarding redirect, i'm 90% sure the hdf5 library doesn't support a redirect.

bendichter: That's a good point. Now that we have [published datasets](https://gui.dandiarchive.org/#/dandiset/search?search=published), maybe we should use one of those.

yarikoptic: filed https://github.com/HDFGroup/hdf5/issues/905

rly:
@bendichter This looks good, thanks!

tests/integration/ros3/test_ros3.py
rly: After a dataset is published, can I still add files to it, or is it locked forever?

satra: you can think of it as software releases. you can add, but the published version contains only the assets it was published with.

bendichter (dismissed):

oruebel (dismissed):
Looks good. Just for record-keeping, can we add a note in the release notes that there is now a DANDI asset for nightly tests with a link. Not that this is technically a feature of PyNWB but it would be nice to have a record of it.

rly:
> Looks good. Just for record-keeping, can we add a note in the release notes that there is now a DANDI asset for nightly tests with a link. Not that this is technically a feature of PyNWB but it would be nice to have a record of it.

Updated now

oruebel (approved):

PR #1424, opened 2022-01-20
Allow TimeSeries to have offset attribute

CodyCBakerPhD:
## Motivation

Analogous to but ultimately independent of the `MeasurementData` (subclass of `VectorData`) changes upstream (https://github.com/hdmf-dev/hdmf/pull/688), it would be great if NWB supported an `offet` in addition to a `conversion` factor for `TimeSeries` objects.

To convert the `TimeSeries` `data` field to the specified `units`, one would perform the operation

```python
data_in_units = data_in_storage * conversion + offset
```

To-Do:
- [x] Make a sister PR with the `nwb-schema`: https://github.com/NeurodataWithoutBorders/nwb-schema/pull/494

## How to test the behavior?
Added tests that the attribute is propagated.

## Checklist


oruebel:
Not that this is something we should do in this PR, but this is another argument for wrapping datasets to allow us to apply these transformations directly rather than putting the burden on the user to transform the data.

CodyCBakerPhD:
> Not that this is something we should do in this PR, but this is another argument for wrapping datasets to allow us to apply these transformations directly rather than putting the burden on the user to transform the data.

@oruebel That is a good point; our approach so far (for `ecephys` anyway) has been to rely on external APIs such as SpikeInterface (https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/baserecording.py#L110-L114) to handle this transformation but the lack of proper offsets within NWBFiles has required awkward workarounds to get it to work properly.

A core functionality for allowing it to work similarly at the `NWBHDF5IO` level would be useful for people not aware and familiar of those external tools, however; we'll have to explore and discuss some possibilities for that design (I remember hearing something about `h5py` filters that can work in that way?).

src/pynwb/base.py
```diff
@@ -111,7 +112,12 @@ class TimeSeries(NWBDataInterface):
              'doc': 'The smallest meaningful difference (in specified unit) between values in data', 'default': -1.0},
             {'name': 'conversion', 'type': (str, 'float'),
              'doc': 'Scalar to multiply each element in data to convert it to the specified unit', 'default': 1.0},
-
+            {'name': 'offset', 'type': (str, 'float'),
```
rly: I do not think string should be allowed here. Or for resolution and conversion. That is surely a bug.

tests/unit/test_base.py
```diff
@@ -92,9 +92,24 @@ def test_init_data(self):
...
+    def test_init_conversion(self):
+        dat = [0, 1, 2, 3, 4]
+        conversion = 2.1
+        ts = TimeSeries('test_ts', dat, 'volts', timestamps=[0.1, 0.2, 0.3, 0.4], conversion=conversion)
+        self.assertIs(ts.data, dat)
+        self.assertEqual(ts.conversion, conversion)
+
+    def test_init_offset(self):
+        dat = [0, 1, 2, 3, 4]
+        offset = 0.0
```
rly: I would test with a non-default value.

CodyCBakerPhD:
```suggestion
        offset = 1.2
```

CodyCBakerPhD: Fantastic idea

src/pynwb/base.py
CodyCBakerPhD: Fixed! I did think that was a bit strange, lol.

CodyCBakerPhD:
@rly Any idea why the tests are failing? They pass for me locally

rly:
The nwb-schema submodule needed to be updated to point to the latest commit on nwb-schema's dev branch. 

miniconda37 fails due to a known package dependency issue.
gallery tests fail because as of 4 days ago, there is an issue with importing allensdk which uses the jinja2 which uses now deprecated functionality from markupsafe.

CodyCBakerPhD:
@rly Hey, sorry, I know your busy - any ETA on when this will be available? It's already in the schema, and is a highly requested feature for all the formats (EDF/intan) that require it for optimal performance.

rly:
Thanks for the ping on this PR. This looks good to go but ideally we should make a new release of the schema first. See https://github.com/NeurodataWithoutBorders/nwb-schema/pull/511 for progress on that.

rly:
Actually, it's almost good to go. "offset" needs to be added to the constructor docval of every subclass of TimeSeries, like for "conversion", e.g., https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/ecephys.py#L78

CodyCBakerPhD:
> Actually, it's almost good to go. "offset" needs to be added to the constructor docval of every subclass of TimeSeries, like for "conversion", e.g., https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/ecephys.py#L78

Oh man, those docvals don't get inherited?

CodyCBakerPhD:
Ohh, I see, I just have to go in and add it to each

```python
            *get_docval(TimeSeries.__init__, 'conversion', 'resolution', 'timestamps', 'starting_time', 'rate',
                        'comments', 'description', 'control', 'control_description'))
```

type of line, like so

```python
            *get_docval(TimeSeries.__init__, 'conversion', 'resolution', 'timestamps', 'starting_time', 'rate',
                        'comments', 'description', 'control', 'control_description', 'offset'))
```

right?

Should I add a new test for this for each subclass? @rly

CodyCBakerPhD:
@rly OK I think I updated all of the docvals, using all those ones we had to change in the schema https://github.com/NeurodataWithoutBorders/nwb-schema/pull/494/files# as a reference of what needed propagation.

Let me know if you think I missed any.

rly:
> Should I add a new test for this for each subclass? @rly

No need

rly:
I think you missed a few such as `RoiResponseSeries`. If you go to the "TimeSeries" bullet point in [the nwb-schema docs](https://nwb-schema.readthedocs.io/en/latest/format.html#type-hierarchy) you'll see a type hierarchy of all types with `TimeSeries` as an ancestor.

Please also revert the changes to ImagingPlane (see https://github.com/NeurodataWithoutBorders/nwb-schema/pull/513)

CodyCBakerPhD:
> Please also revert the changes to ImagingPlane (see https://github.com/NeurodataWithoutBorders/nwb-schema/pull/513)

Done

> I think you missed a few such as RoiResponseSeries. If you go to the "TimeSeries" bullet point in [the nwb-schema docs](https://nwb-schema.readthedocs.io/en/latest/format.html#type-hierarchy) you'll see a type hierarchy of all types with TimeSeries as an ancestor.

Oops, sorry about that. Guess basing it off the schema changes didn't account for all the ones that don't need it overridden. I've gone through the list on the readthedocs schema view and added it to all except: `IntervalSeries` and `AnnotationSeries` since they also don't take `conversion`, the necessary compliment to this.

rly (approved):

rly:
Looks good to me. Thanks!

PR #1426, opened 2022-02-02
add time series check

bendichter:
Check if the length of data is equal to the length of timestamps


rly:
This approach is good for now. 

@oruebel and I discussed adding a general `validate` or `check` method to `hdmf.container.AbstractContainer` that can be manually called by the user and automatically called by the BuildManager or ObjectMapper on write. By having these checks only in `__init__` in this PR, users can still set or modify certain properties later and bypass those checks. 

We can refactor these checks (as well as the other ones currently in various other `__init__` methods) to the new `validate` method later on.

src/pynwb/base.py
```diff
@@ -146,6 +147,14 @@ def __init__(self, **kwargs):
                 "control",
                 "control_description",
                 "continuity")
+
+        if (
+            isinstance(kwargs["data"], (np.ndarray, list))
+            and isinstance(kwargs["timestamps"], (np.ndarray, list))
+            and not len(kwargs["data"]) == len(kwargs["timestamps"])
+        ):
```
oruebel: For these shape-based checks, I think it may be useful to use [get_data_shape](https://github.com/hdmf-dev/hdmf/blob/87e071efc41fde88ef914e8f2e22ccb1f84ae471/src/hdmf/utils.py#L766) from HDMF. I think one advantage of using ``get_data_shape`` is that we don't need to deal with differences between lists, ndarray, h5py.Dataset, or DataChunkIterator etc. for each of the checks, but those details are handled in ``get_data_shape``. Another benefit is that I think it would allow us to cover more cases than just the nd.array and list case that is covered by the current check, but it should also work for DataChunkIterator in the cases, e.g., when maxshape is set on the DataChunkIterator. Using the ``get_data_shape`` function, the check would I think look something like this:

```
from hdmf.utils import get_data_shape
...

data_shape = get_data_shape(data=kwargs["data"], strict_no_data_load=True)
timestamps_shape = get_data_shape(data=kwargs["data"], strict_no_data_load=True)
if( (data_shape is not None and timestamps_shape is not None)   # check that the shape is known
     and (len(data_shape) > 0 and len(timestamps_shape) > 0)   # check for scalars. Should never happen
     and (data_shape[0] is not None and timestaps_shape[0] is not None)  # check that the length of the first dimension is known
     and (data_shape[0] != timestamps_shape[0])  # check that the data and timestamps match
   )
       warn(...)
```

oruebel:
> We can refactor these checks (as well as the other ones currently in various other `__init__` methods) to the new `validate` method later on.

I agree. This approach is fine for now. I think the overhead to refactor this later should be reasonable.

oruebel (dismissed):
Looks good to me

rly (dismissed):

rly:
Please add an entry to the changelog and then we are good to merge.

rly (approved):

PR #1453, opened 2022-04-15
Add tutorial about reading and exploring an NWB File

weiglszonja:
Add new tutorial about reading, exploring and doing basic visualisations with an existing NWB File.

This tutorial demonstrates how to:
- download an NWB file from DANDI or stream from S3
- read the NWB file
- access stimulus data
- plot images from OpticalSeries data
- plot spiking activity relative to stimulus onset
- access trials
- plot stimuli matched on trials `stim_on_time` column
- tools for exploring an NWB file: `nwbwidgets`, `HDFView`

## Checklist


bendichter:
this looks great, @weiglszonja! One issue with how this is currently set up is that it requires the NWB files to exist locally, which is tricky, because the server that runs this won't have access to that file. Maybe we could also offer an [s3 streaming version](https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/streaming.html#sphx-glr-tutorials-advanced-io-streaming-py) and show the rest through that.

bendichter:
in fact, I think we could combine the s3 streaming tutorial with this one

docs/gallery/general/plot_read_demo.py
```diff
@@ -0,0 +1,261 @@
...
+# metadata about each trial in a tabular form, where each row represents a trial and has a start time, stop time,
+# and additional metadata.
+#
+# .. seealso::
+#     You can learn more about trials in the :ref:`basic_trials` tutorial section.
+#
+# Similarly to :py:class:`~pynwb.misc.Units` we can view trials as a :py:class:`pandas.DataFrame`.
+
+trials_df = nwbfile.trials.to_dataframe()
+print(trials_df)
```
oruebel: To get a prettier display of the table, you  can just place the dataframe itself as the last entry of a cell so that it will be rendered in html directly using a display that is essentially the same as how you would see the table in a Jupyter notebook. I.e., instead pf calling ``print(trials_df)`` you can simply write ``trials_df`` and sphinx gallery will render it

```suggestion
trials_df
```

docs/gallery/general/plot_read_demo.py
```diff
@@ -0,0 +1,261 @@
...
+
+stim_on_times_landscapes = trials_df[trials_df.category_name == 'landscapes'].stim_on_time
+for time in stim_on_times_landscapes[:3]:
+    img = np.squeeze(stimulus_presentation.data[np.where(stimulus_presentation.timestamps[:] == time)])
+    # Reverse the last dimension because the data were stored in BGR instead of RGB
+    img = img[..., ::-1]
+    plt.figure()
+    plt.imshow(img, aspect='auto')
+
+####################
```
oruebel: We should add a link to https://nwb-overview.readthedocs.io/en/latest/tools/tools_home.html for details somewhere in the intro of this section for details.

docs/gallery/general/plot_read_demo.py
```diff
@@ -0,0 +1,261 @@
+"""
+.. _reading_and_exploring_NWB_File:
+
+Reading and Exploring an NWB File
```
oruebel: Thanks for putting together this tutorial. Just a few (random) general thoughts. 

Ultimately, I think it will be useful to have tutorials like this for each main application area (i.e., at least ecephys, ophys, and icephys) where we can then cover details that are specific to the area. E.g., for ecephys that would involve items like using ``ElectricalSeries.electrodes`` to locate and use electrode metadata. Or for icephys it involves dealing with the icephys metadata tables, for which we already have a tutorial here https://pynwb.readthedocs.io/en/stable/tutorials/domain/plot_icephys_pandas.html . As such, I would suggest that we focus here on introducing common key concepts and then link to tutorials for the different areas for further details in those areas. I think overal this tutorial is a great start for this, but we may (eventually) want to move items that are specific for a particular area to (e.g., ``units``) to the corresponding domain-specific read tutorial. 

I don't think we necessarily have to cover all these concepts here, but I just wanted to mention a few general concepts that I think could be useful for read tutorials to:
* ``Container.children`` and ``NWBFile.all_children`` to iterate over containers
* ``NWBFile.acquistion``, ``NWBFile.processing`` and ``NWBFile.all_children`` use  ``LabelledDict`` which has some (basic) features that are useful for query (see https://hdmf.readthedocs.io/en/stable/hdmf.utils.html?highlight=LabelledDict#hdmf.utils.LabelledDict). E.g., if you want to find all ``ElectricalSeries`` in acquisition you could do something like ``NWBFile.acquistion['neurodata_type==ElectricalSeries']``
* A few general concepts from the ``Container`` base class that  could be of interest for this tutorial, are, e.g,: ``type_hierarchy()``, ``.parent``, ``objectid``, ``get_ancestor``. Some of these may be too much detail or not useful here, but I just wanted to mention them in case they are useful (see https://hdmf.readthedocs.io/en/stable/hdmf.container.html?highlight=fields#hdmf.container.AbstractContainer)
* Similarly, ``hdmf.container.Data`` has a few additional features that could be useful for read, e.g., the ``transform`` method. 
* ...

oruebel:
> in fact, I think we could combine the s3 streaming tutorial with this one

If only for find-ability, I think it will be useful to keep a separate tutorial around for S3, even if that tutorial is then very brief and only shows how to open a file and then points to this tutorial for details. Other than that, I think using S3 for this tutorial is fine, but I think it would be useful to at least briefly discuss the main difference between downloading a file and using it via S3.

docs/gallery/general/plot_read_demo.py
weiglszonja: How about something like this:

https://github.com/NeurodataWithoutBorders/pynwb/blob/c8f47016f779bca28ec6a9bdd401996ab93c82d1/docs/gallery/general/plot_read_demo.py#L16-L18

docs/gallery/general/plot_read_demo.py
weiglszonja: Thank you! I removed them

docs/gallery/general/plot_read_demo.py
```diff
@@ -54,6 +66,50 @@
...
+# Then, we will use the ``DandiAPIClient`` to obtain the S3 URL that points to the NWB File
+# stored in S3. We will need the identifier of the dataset (``dandiset_id``) and the path
+# to the NWB File.
+# We can read these from the DANDI archive URL where ``dandiset_id`` is "000004" and
+# file is located in "sub-P11HMH" folder.
+#
+# .. image:: ../../_static/demo_dandi_url.png
+#   :width: 700
+#   :alt: dandi url explained
+#   :align: center
```
weiglszonja: I'm not sure this image is needed, I can remove it. Just wanted to show how to get these parameters from the archive url

bendichter: Do you not think it's clear how to infer them from the dandi web ui?

weiglszonja: Yeah now I think the image is not necessary..

docs/gallery/general/plot_read_demo.py
weiglszonja: copying @rly suggestion from meeting:
instead of `NWBFile.acquistion['neurodata_type==ElectricalSeries'] `
recommended:` [x for x in nwbfile.objects if x isinstance(TimeSeries)]`

weiglszonja:
I noticed there is a section [here](https://nwb-overview.readthedocs.io/en/latest/file_read/file_read.html#reading-with-pynwb) about reading NWB files with PyNWB but is currently empty. Do you think @oruebel it would make sense to place a link there that redirects people to this tutorial?

bendichter:
@weiglszonja yes, once this tutorial is ready we should route people to it. We should also use this section to route users to more advanced and specialized read tutorials as they become available. In general, the nwb-overview documentation is for providing a bit of context and then routing users to more in-depth tutorials on specific repos.

docs/gallery/general/plot_read_demo.py
weiglszonja: thank you @oruebel for the suggestions so far. I added a new issue #1457 for creating an advanced read tutorial later where these base class features can be covered.

oruebel:
> replace all code sections with inline code-blocks

As far as I know, using inline code-blocks has the disadvantage that the code-blocks are not being executed and as such are not being tested.

weiglszonja:
@oruebel 
Yes, the reason why I did this is because `dandi` is a dependency that is not being installed automatically when the CI test is running for this tutorial. The streaming tutorial uses inline code-blocks too, so I thought I'd apply them here to avoid this issue. Do you have a suggestion how to resolve this issue?

oruebel:
> I did this is because `dandi` is a dependency that is not being installed automatically

I see, thanks for the clarification. I think the S3 tutorial we didn't end up making DANDI a dependency because it was such a brief tutorial. Since you are significantly expanding this now here, I think it is worth revisiting this question since, ideally, we want to be able to have all these tutorial be run to ensure they keep working in the future. The options we have then are:

1. Add dandi to the ``requirements-doc.txt`` so that we can use it in the docs
2. Download a file via python directly in the tutorial so that it can then be used locally 
3. Have code that requires web access in code-blocks (so they are not being run) and then work of small local files that are either added to the repo or dummy files that are generated on the fly 

3. has the problem that it makes the tutorials less readable and works around the problem. Because of this, I think 1. or 2. are probably preferred but have the caveat that they require web access (so tests could fail if web access fails)  and they generate traffic on DANDI. @rly @bendichter  thoughts?

weiglszonja:
I'm not sure how frequently this readthedocs build is happening in the background, but it will definitely slow it down if we go for option 1. Also is it feasible given that streaming requires a special setup with the [ROS3 driver](https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/streaming.html#sphx-glr-tutorials-advanced-io-streaming-py)? (let me know @CodyCBakerPhD if I'm missing something here)

If we go for option 3, is it possible to hide code snippets? So the tutorial would look as it is (showing how to download a file or stream it from S3) but in the background it would actually read a small file added to the repo?

CodyCBakerPhD:
> Also is it feasible given that streaming requires a special setup with the [ROS3 driver](https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/streaming.html#sphx-glr-tutorials-advanced-io-streaming-py)? (let me know @CodyCBakerPhD if I'm missing something here)

What I've had to do in the past so setup CI to perform streaming would be difficult to do on the readthedocs environment side, and also pretty costly to do every time a new build is done. If you just want to test the code blocks inside the docs to make sure they always continue running, I'd suggest just adding them as tests to the main testing suite that I assume is already setup to test other streaming behavior.

> If we go for option 3, is it possible to hide code snippets? So the tutorial would look as it is (showing how to download a file or stream it from S3) but in the background it would actually read a small file added to the repo?

We did something like that a while back on the NWBCT - I no longer have the sphinx builds from that, but this snippet is what did it: https://github.com/catalystneuro/nwb-conversion-tools/commit/8e9e6dcc264d6fb738ceadd8ffd372c78038f0d3#diff-5adb8bffbac6c004685f555404d3586533e8c94f26c82fbe3592f81d977ae513R165-R176

Basically it adds an arrow that expands the code block when you click it, but otherwise minimizes it.

weiglszonja:
> Basically it adds an arrow that expands the code block when you click it, but otherwise minimizes it.

Thanks @CodyCBakerPhD, I think I had something else in mind based on [this](https://stackoverflow.com/questions/49813426/hide-single-section-in-sphinx-documentation):

```
Visible section
================
Here I will show how to stream the data using inline code-blocks

.. class:: hidden

   Not visible section
   ===================
   Here is the code snippet that reads the small file from the repo
   but for the current build I want to hide from the final document

Another section
===============
Other visible section in both text and final document
```

But this example is for text, I'm not sure this would work in this case.

rly:
If I understand correctly, the problem is that we want to test this tutorial in our CI, but downloading/streaming a large file is prohibitive. 

All Python files in `docs/gallery` are run during our nightly CI tests. Only Python files that have the prefix `plot_` are executed and their outputs captured and incorporated in the HTML output when you call `make html` locally or on readthedocs. This also means that if you want a pandas DataFrame to be displayed prettily, then the file must have the prefix `plot_`.

I think @oruebel 's **solution 1** (add `dandi` in `requirements-doc.txt`, do not use inline code blocks) is fine if we can use a file that is a lot smaller. Yes, this will slow down documentation builds, but if the file is small, it might not be significant. Can we test how much time it would add?

Note that the allensdk tutorial downloads a 450 MB file in CI tests. However, because the tutorial file does not have the prefix `plot_`, the download does not happen on readthedocs. We can do the same thing with this tutorial (call this **solution 1b**) -- do not use in-line code blocks and do not use the prefix `plot_`. Then the DANDI file will not be downloaded and outputs will not be rendered on readthedocs. Perhaps we can display output as in-line code blocks where appropriate. This might eventually result in a mismatch between the hard-coded output and the actual output of the code, but I expect this to be rare. What do you think?

> I'd suggest just adding them as tests to the main testing suite that I assume is already setup to test other streaming behavior.

This is OK but means that if the API is updated, then both the test in the testing suite and the documentation example must both be updated, and if either the test or example is updated, then the other file must also be updated so they stay in sync. This is tricky and I would prefer not to have this implicit linkage.

> how frequently this readthedocs build is happening in the background

On every push.

CodyCBakerPhD:
> I think @oruebel 's solution 1 (add dandi in requirements-doc.txt, do not use inline code blocks) is fine if we can use a file that is a lot smaller. Yes, this will slow down documentation builds, but if the file is small, it might not be significant. Can we test how much time it would add?

So, my complaint on time expense is actually not related to the actual time to open/read the file in stream mode but actually to perform the `conda-forge` install of `h5py`, which is one of the major slowdowns currently on the NWBInspector CI. Size of file doesn't really relate too much to streaming speed, even for icephys files with lots of items in them.

But I guess if the build is only nightly and doesn't trigger on PRs, it could be OK to do that way, then.

weiglszonja:
> Note that the allensdk tutorial downloads a 450 MB file in CI tests. However, because the tutorial file does not have the prefix plot_, the download does not happen on readthedocs. We can do the same thing with this tutorial (call this solution 1b) -- do not use in-line code blocks and do not use the prefix plot_. Then the DANDI file will not be downloaded and outputs will not be rendered on readthedocs. Perhaps we can display output as in-line code blocks where appropriate. This might eventually result in a mismatch between the hard-coded output and the actual output of the code, but I expect this to be rare. What do you think?

Thanks for the clarification how the CI tests work for the tutorial @rly, this sounds good to me!

weiglszonja:
@rly I added dandi to `requirements-doc.txt` and the tutorial does not have the "plot_" prefix before the file name. But there are still failing [tests](https://app.circleci.com/pipelines/github/NeurodataWithoutBorders/pynwb/2773/workflows/fabc1f42-492e-4a55-b355-41b3fd339d37/jobs/49448): 
`h5py was built without ROS3 support, can't use ros3 driver`

Is there something else that I'm missing?

rly:
Ah, the sphinx-gallery tests need h5py installed from conda-forge instead of from pypi. I'll try to fix it.

weiglszonja:
Thank you @rly, looks good! 👍

docs/gallery/general/read_basics.py
```diff
@@ -0,0 +1,374 @@
...
+# which can be installed from pip:
+#
+# .. code-block:: bash
+#
+#    $ pip install -U dandi
+#
+# .. seealso::
+#     You can learn more about streaming data in the :ref:`streaming` tutorial.
+#
+# Then, we will use the ``DandiAPIClient`` to obtain the S3 URL that points to the NWB File
```
bendichter: can we use intersphinx to link to the [API documentation for DANDIAPIClient](https://dandi.readthedocs.io/en/stable/modref/dandiapi.html#dandi.dandiapi.DandiAPIClient)?

weiglszonja: oh sure, I'll add it!

bendichter:
@weiglszonja it looks like you are becoming a sphinx expert!

bendichter (dismissed):

docs/gallery/general/read_basics.py
```diff
@@ -24,6 +24,7 @@
 The following examples will reference variables that may not be defined within the block they are used in. For
 clarity, we define them here:
 """
+# sphinx_gallery_thumbnail_path = 'figures/gallery_thumbnails_read_basics.png'
```
weiglszonja, line 27: @oruebel, I took the liberty to try creating a thumbnail myself, let me know if you don't like it though!

oruebel: The thumbnail looks great! Just a minor suggestion, can you crop the image to remove the empty space on the top/right to make sure it renders at the same size as the other thumbnails in the gallery.

weiglszonja: Thanks! I cropped and resized it to be the same size as the others

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Create behavioral tutorial @weiglszonja (#1464) 
 - Enhance display of icephys pandas tutorial by using ``dataframe_image`` to render and display large tables as images. @oruebel (#1469)
 
+- Create tutorial about reading and exploring an existing `NWBFile` @weiglszonja (#1453)
```
oruebel:
```suggestion
- Create tutorial about reading and exploring an existing `NWBFile` @weiglszonja (#1453)
```

oruebel (dismissed):
Looks good to me. Thanks for all the hard work in putting this tutorial together.

rly:
In two spots in this tutorial, it is noted "# Reverse the last dimension because the data were stored in BGR instead of RGB"

@bendichter Could we correct this in the dandiset and release a new version where this transformation is not necessary?

bendichter:
I don't have permission to merge this. @weiglszonja, do you?

CodyCBakerPhD (approved):

bendichter:
@rly I generally don't make changes to datasets that I wasn't involved in converting and publishing to begin with

bendichter:
also for this particular dataset I know they have Python and MATLAB code associated with it, and I would not want to make changes that might break their code

weiglszonja:
@bendichter I don't have merge permission

Issue #1318, opened 2020-11-16
Verification: ImageSeries with list of external files & its timestamps


CodyCBakerPhD:
Hey all,

I'm trying to convert a list of `.mkv` files into `ImageSeries` without actually encoding the data, that is, via the `format=external` option. My concern, however, is dealing with the timestamps for each of these `.mkv` separately.

Attempting to do the intuitive thing, and pass timestamps as a list of arrays, each element of the list corresponding to the `.mkv `file, causes an error. Obviously, I can simply concatenate all the timestamps, but I'm just trying to seek verification that that is the correct approach to take, especially in the event that there is perhaps a mismatch between the number of timestamps and the number of frames for a given file (flattening the timestamps would make it much harder to debug such as issue).

Here is some simple code to replicate the issue, granted you'll have to find some example .mkv/.csv files to test it yourself.

```
from pynwb import NWBFile
import uuid
from pathlib import Path
from datetime import datetime
import pandas as pd
import numpy as np
from pynwb.image import ImageSeries

nwbfile = NWBFile(
    session_description="Replicating ImageSeries issue.",
    identifier=str(uuid.uuid4()),
    session_start_time=datetime.now()
)

video_folder = Path("D:/ExampleImageSeriesIssue")
video_file_path_list = [str(x.absolute()) for x in video_folder.iterdir() if x.suffix == ".mkv"]

video_timestamps = list()
for video_file_path in video_file_path_list:
    video_time_file = pd.read_csv(video_file_path.replace(".mkv", "_timestamps.csv"), header=0)
    video_timestamps.append(np.array([int(x.split(";")[1]) for x in video_time_file['frame; timestamp']]))

# Encounters an error
try:
    videos1 = ImageSeries(
        name='Videos',
        description="Videos recorded by TIS camera.",
        format="external",
        external_file=video_file_path_list,
        timestamps=video_timestamps
    )
    nwbfile.add_acquisition(videos1)
except ValueError:
    print("For Testing: attempting to make the ImageSeries returned "
          "ValueError: ImageSeries.__init__: incorrect shape for 'timestamps' (got '(2, 57409)', expected '(None,)')")

# Doesn't fail, but also can't check if the timestamps align with number of frames in external file
videos2 = ImageSeries(
    name='Videos',
    description="Videos recorded by TIS camera.",
    format="external",
    external_file=video_file_path_list,
    timestamps=np.concatenate(video_timestamps)
)
[...]

PR #1470, opened 2022-05-04
Add images tutorial

weiglszonja:
## Motivation

Create a new tutorial that demonstrates how to use the [image](https://pynwb.readthedocs.io/en/stable/pynwb.image.html#module-pynwb.image) module for adding images to an NWB File.

The images tutorial covers the following data types:
- [x] `pynwb.image.GrayscaleImage`, `pynwb.image.RGBImage`, `pynwb.image.RGBAImage`
- [x] `pynwb.image.OpticalSeries `
- [x] `pynwb.image.ImageSeries`,
- [ ] `pynwb.image.ImageMaskSeries `,
- [ ] `pynwb.image.IndexSeries` (which was changed [recently](https://github.com/NeurodataWithoutBorders/nwb-schema/pull/459))

## Checklist


docs/gallery/domain/images.py
```diff
@@ -0,0 +1,214 @@
...
+####################
+#
+# .. seealso::
+#    You can learn more about the :py:class:`~pynwb.file.NWBFile` format in the :ref:`basics` tutorial.
+#
+# OpticalSeries: Storing series of images as stimuli
+# --------------------------------------------------
+#
+# :py:class:`~pynwb.image.OpticalSeries` is for series of images that were presented
+# to the experimental subject as stimuli.
```
bendichter:
```suggestion
# to the subject as stimuli.
```

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,214 @@
...
+# We will create an :py:class:`~pynwb.image.OpticalSeries` object with the name
+# ``"StimulusPresentation"`` representing what images were shown to the subject and at what times.
+#
+# Image data can be stored either in the HDF5 file or as an external image file.
+# For this tutorial, we will use a fake image data with shape of (200, 50, 50, 3).
```
bendichter:
```suggestion
# For this tutorial, we will use fake image data with shape of (200, 50, 50, 3).
```

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,214 @@
...
+# Images may be 3D or 4D (grayscale or RBG), where the first dimension must be time (frame).
```
bendichter, line 69:
```suggestion
# Images may be 3D or 4D (RGB or RBGA), and the first dimension must be time (frame).
```

bendichter:
```suggestion
```

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,214 @@
...
+# The second and third dimensions represent x and y.
+# The fourth dimension represents the RGB value (length of 3) for color images.
+#
+# NWB differentiates between raw, acquired data and data that was presented as stimulus.
```
bendichter:
```suggestion
# NWB differentiates between acquired data and data that was presented as stimulus.
```

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,214 @@
...
+# We can add it to the :py:class:`~pynwb.file.NWBFile` object as stimulus data using
+# the :py:meth:`~pynwb.file.NWBFile.add_stimulus` method.
+#
+
+image_data = np.random.uniform(low=0, high=255, size=(200, 50, 50, 3)).astype(int)
```
bendichter: I think this should be uint8, which is the standard for imaging data

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,214 @@
...
+optical_series = OpticalSeries(
+    name="StimulusPresentation",  # required
+    distance=0.7,  # required
+    field_of_view=[0.2, 0.3, 0.7],  # required
+    orientation="lower left",  # required
+    data=image_data,
+    unit="m",
```
bendichter: What does unit mean here? Maybe we should change this to "n.a."

bendichter:
@weiglszonja let's move forward with this. I think the content we already have here is good. It doesn't cover everything yet but let's get this out before the user days to increase our tutorial coverage

bendichter:
This content would need to be removed from NWBFile basics

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,246 @@
...
+# External files (e.g. video files of the behaving animal) can be added to the :py:class:`~pynwb.file.NWBFile` by creating
+# an :py:class:`~pynwb.image.ImageSeries` object using the ``external_file`` attribute that specifies the
+# path to the external file(s) on disk. The file(s) path must be relative to the path of the NWB file.
+# Either ``external_file`` or ``data`` must be specified, but not both.
+
+behavior_external_file = ImageSeries(
+    name="ExternalFiles",
+    description="Behavior video of animal moving in environment.",
+    unit="Frames",
+    external_file=["/path/to/image/file"],
```
bendichter:
```suggestion
    external_file=["/path/to/image/file.avi"],
```

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,246 @@
...
+####################
+# External files (e.g. video files of the behaving animal) can be added to the :py:class:`~pynwb.file.NWBFile` by creating
+# an :py:class:`~pynwb.image.ImageSeries` object using the ``external_file`` attribute that specifies the
+# path to the external file(s) on disk. The file(s) path must be relative to the path of the NWB file.
+# Either ``external_file`` or ``data`` must be specified, but not both.
+
+behavior_external_file = ImageSeries(
+    name="ExternalFiles",
+    description="Behavior video of animal moving in environment.",
+    unit="Frames",
```
bendichter:
```suggestion
    unit="n.a.",
```

"unit" refers to the unit of the data. For instance, in an ElectricalSeries, the unit is volts. This needs to be consistent. We can't have the unit referring to anything else.

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,246 @@
...
+# For this tutorial, we will use fake image data with shape of (200, 50, 50, 3).
+# The second and third dimensions represent x and y.
+# The fourth dimension represents the RGB value (length of 3) for color images.
+#
+# NWB differentiates between acquired data and data that was presented as stimulus.
+# We can add it to the :py:class:`~pynwb.file.NWBFile` object as stimulus data using
+# the :py:meth:`~pynwb.file.NWBFile.add_stimulus` method.
+#
+
+image_data = np.random.uniform(low=0, high=255, size=(200, 50, 50, 3)).astype(np.uint8)
```
bendichter: should use [randint](https://numpy.org/doc/stable/reference/random/generated/numpy.random.randint.html#numpy.random.randint)

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,246 @@
...
+#
+# :py:class:`~pynwb.image.ImageSeries` is a general container for series of images acquired during
+# the experiment. Image data can be stored either in the HDF5 file or as an external image file.
+# When color images are stored in the HDF5 file the color channel order is expected to be RGB.
+#
+# We can add raw data to the :py:class:`~pynwb.file.NWBFile` object as *acquisition* using
+# the :py:meth:`~pynwb.file.NWBFile.add_acquisition` method.
+#
+
+image_data = np.random.uniform(low=0, high=255, size=(200, 50, 50, 3)).astype(np.uint8)
```
bendichter: should use randint

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,246 @@
...
+
+behavior_images = ImageSeries(
+    name="ImageSeries",
+    data=image_data,
+    description="Image data of an animal moving in environment.",
+    unit="pixels",
+    format="raw",
+    starting_frame=[0.0],
```
bendichter: This argument should only be used if it is an external file.

docs/gallery/domain/images.py
weiglszonja: I see, thank you for clarifying it.

docs/gallery/domain/images.py
weiglszonja: Ah, you're right!

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,246 @@
...
+
+Images
+==================================
+
+This tutorial will demonstrate the usage of the :py:mod:`pynwb.image` module for adding
+images to an :py:class:`~pynwb.file.NWBFile`.
+
+Image data can be of individual images or of movie segments (as a movie is simply a series of images),
+about the subject, the environment, the presented stimuli, or other kind
+that relates to the experiment.
```
oruebel:
```suggestion
Image data can be a collection of individual images or movie segments (as a movie is simply a series of images),
about the subject, the environment, the presented stimuli, or other parts
related to the experiment. This tutorial focuses in particular on the usage of:
```

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,246 @@
...
+#
+# :py:class:`~pynwb.image.OpticalSeries` is for series of images that were presented
+# to the subject as stimuli.
+# We will create an :py:class:`~pynwb.image.OpticalSeries` object with the name
+# ``"StimulusPresentation"`` representing what images were shown to the subject and at what times.
+#
+# Image data can be stored either in the HDF5 file or as an external image file.
+# For this tutorial, we will use fake image data with shape of (200, 50, 50, 3).
+# The second and third dimensions represent x and y.
+# The fourth dimension represents the RGB value (length of 3) for color images.
```
oruebel, line 80:
```suggestion
# As in all :py:class:`~pynwb.base.TimeSeries`, the first dimension is time.
# The second and third dimensions represent x and y.
# The fourth dimension represents the RGB value (length of 3) for color images.
```

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,246 @@
...
+# OpticalSeries: Storing series of images as stimuli
+# --------------------------------------------------
+#
+# :py:class:`~pynwb.image.OpticalSeries` is for series of images that were presented
+# to the subject as stimuli.
+# We will create an :py:class:`~pynwb.image.OpticalSeries` object with the name
+# ``"StimulusPresentation"`` representing what images were shown to the subject and at what times.
+#
+# Image data can be stored either in the HDF5 file or as an external image file.
+# For this tutorial, we will use fake image data with shape of (200, 50, 50, 3).
```
oruebel:
```suggestion
# For this tutorial, we will use fake image data with shape of ``('time', 'x', 'y', 'RGB') = (200, 50, 50, 3)``.
```

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,244 @@
...
+
+####################
+#
+# .. seealso::
+#    You can learn more about the :py:class:`~pynwb.file.NWBFile` format in the :ref:`basics` tutorial.
+#
+# OpticalSeries: Storing series of images as stimuli
+# --------------------------------------------------
+#
+# :py:class:`~pynwb.image.OpticalSeries` is for series of images that were presented
```
oruebel:
```suggestion
# :py:class:`~pynwb.image.OpticalSeries` is for time series of images that were presented
```

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,244 @@
...
+    description="The images presented to the subject as stimuli",
+)
+
+nwbfile.add_stimulus(timeseries=optical_series)
+
+####################
+# ImageSeries: Storing series of images as acquisition
+# ----------------------------------------------------
+#
+# :py:class:`~pynwb.image.ImageSeries` is a general container for series of images acquired during
```
oruebel:
```suggestion
# :py:class:`~pynwb.image.ImageSeries` is a general container for time series of images acquired during
```

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,244 @@
...
+# External files (e.g. video files of the behaving animal) can be added to the :py:class:`~pynwb.file.NWBFile` by creating
+# an :py:class:`~pynwb.image.ImageSeries` object using the ``external_file`` attribute that specifies the
+# path to the external file(s) on disk. The file(s) path must be relative to the path of the NWB file.
+# Either ``external_file`` or ``data`` must be specified, but not both.
+
+behavior_external_file = ImageSeries(
+    name="ExternalFiles",
+    description="Behavior video of animal moving in environment.",
+    unit="n.a.",
+    external_file=["/path/to/image/file.avi"],
```
oruebel, line 127: Having `` external_file=["/path/to/image/file.avi"]`` here may give the false impression of an absolute path. It would be nice to show something that makes it clear that this is a relative path and shows users how to construct relative paths in Python.  I would suggest to: 1) define the paths to the nwb file and movie file as variables beforehand with descriptive variable names, e.g., ``nwbfile_path="/path/nwbfile.nwb"`` and ``movie_path="..."`` , 2) Explicitly construct the relative path, e.g., via ``movie_relative_path = os.path.relpath(nwbfile_path, movie_path)`` as a separate variable or directly in the assignment of ``external_file=[]``.

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,244 @@
...
+    format="external",
+    starting_frame=[0.0],
+    rate=1.0,
+)
+
+nwbfile.add_acquisition(behavior_external_file)
+
+####################
```
oruebel, line 164: As part of the discussion with DANDI we had at some point created a best practice for how to store external files alongside NWB for upload to DANDI. If this best practice has already been published, then it would be good to add a note here to point to this best practice here. @bendichter do you know where this best practice is being described, is this in the docs of DANDI or the nwbinspector?

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,244 @@
...
+# path to the external file(s) on disk. The file(s) path must be relative to the path of the NWB file.
+# Either ``external_file`` or ``data`` must be specified, but not both.
+
+behavior_external_file = ImageSeries(
+    name="ExternalFiles",
+    description="Behavior video of animal moving in environment.",
+    unit="n.a.",
+    external_file=["/path/to/image/file.avi"],
+    format="external",
+    starting_frame=[0.0],
```
oruebel:
```suggestion
    starting_frame=[0],
```

Starting frame should be an int per https://nwb-schema.readthedocs.io/en/latest/format.html#imageseries

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,244 @@
...
+    rate=1.0,
```
oruebel: If rate is specified, I think we should also set the ``starting_time``

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,244 @@
...
+)
+
+nwbfile.add_acquisition(behavior_images)
+
+####################
+# External files (e.g. video files of the behaving animal) can be added to the :py:class:`~pynwb.file.NWBFile` by creating
+# an :py:class:`~pynwb.image.ImageSeries` object using the ``external_file`` attribute that specifies the
+# path to the external file(s) on disk. The file(s) path must be relative to the path of the NWB file.
+# Either ``external_file`` or ``data`` must be specified, but not both.
+
```
oruebel: Here you show the use of a single movie file with fixed starting_time and sampling_rate. In addition,  I think it would be useful to also show how this will look when you have a collection of images (e.g,. a bunch of Tiff files) and possibly using explicit timestamps for those images. I think this will be useful mainly because I believe this is also a fairly common use.

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,244 @@
...
+from dateutil import tz
+
+import numpy as np
+from PIL import Image
+
+from pynwb import NWBFile, NWBHDF5IO
+from pynwb.base import Images
+from pynwb.image import RGBAImage, RGBImage, GrayscaleImage, OpticalSeries, ImageSeries
+
+####################
```
oruebel:
```suggestion
from pynwb.image import RGBAImage, RGBImage, GrayscaleImage, OpticalSeries, ImageSeries

# Define file paths used in the tutorial
import os
nwbfile_path = os.path.abspath("images_tutorial.nwb")
moviefile_path = os.path.abspath("image/file.avi")

####################
```

docs/gallery/domain/images.py
oruebel:
```suggestion
    external_file=[os.path.relpath(nwbfile_path, moviefile_path)],
```

Having external_file=["/path/to/image/file.avi"] here may give the false impression of an absolute path. It would be nice to show something that makes it clear that this is a relative path and shows users how to construct relative paths in Python. I would suggest to: 1) define the paths to the nwb file and movie file as variables beforehand and 2) Explicitly construct the relative path. I added a suggestion above and here to make that change.

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,244 @@
...
+
+nwbfile.add_acquisition(images)
+
+####################
+# Writing the images to an NWB File
+# ---------------------------------------
+# As demonstrated in the :ref:`basic_writing` tutorial, we will use :py:class:`~pynwb.NWBHDF5IO`
+# to write the file.
+
+with NWBHDF5IO("images_tutorial.nwb", "w") as io:
```
oruebel:
```suggestion
with NWBHDF5IO(nwbfile_path, "w") as io:
```

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,244 @@
...
+#
+# We can also access :py:class:`~pynwb.image.OpticalSeries` data that was added to the NWB File
+# as stimuli by indexing ``nwbfile.stimulus`` with the name of the :py:class:`~pynwb.image.OpticalSeries`
+# object "StimulusPresentation".
+# Data arrays are read passively from the file.
+# Accessing the ``data`` attribute of the :py:class:`~pynwb.image.OpticalSeries` object
+# does not read the data values into memory, but returns an HDF5 object that can be indexed to read data.
+# Use the ``[:]`` operator to read the entire data array into memory.
+
+with NWBHDF5IO("images_tutorial.nwb", "r") as io:
```
oruebel:
```suggestion
with NWBHDF5IO(nwbfile_path, "r") as io:
```

oruebel:
I added a few minor suggestions for edits, but overall this looks good. Thanks for putting together this tutorial! This fills a critical gap in our docs. 👍

docs/gallery/domain/images.py
weiglszonja:
```suggestion
    rate=1.0,
    starting_time=0.0,
```

docs/gallery/domain/images.py
weiglszonja: @oruebel maybe you're referring to this? https://www.dandiarchive.org/2022/03/03/external-links-organize.html

docs/gallery/domain/images.py
weiglszonja: Is it correct that `os.path.relpath(nwbfile_path, moviefile_path)` evaluates to ` '../../images_tutorial.nwb'`? I suspect that the order should be reversed based on [this](https://docs.python.org/3/library/os.path.html#os.path.relpath)

docs/gallery/domain/images.py
oruebel: Yes, that link looks good

docs/gallery/domain/images.py
oruebel: Yes, you are correct. It should be ``os.path.relpath(moviefile_path, nwbfile_path)`` instead.

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,251 @@
...
+# External files (e.g. video files of the behaving animal) can be added to the :py:class:`~pynwb.file.NWBFile` by creating
+# an :py:class:`~pynwb.image.ImageSeries` object using the ``external_file`` attribute that specifies the
+# path to the external file(s) on disk. The file(s) path must be relative to the path of the NWB file.
+# Either ``external_file`` or ``data`` must be specified, but not both.
+
+behavior_external_file = ImageSeries(
+    name="ExternalFiles",
+    description="Behavior video of animal moving in environment.",
+    unit="n.a.",
+    external_file=[os.path.relpath(nwbfile_path, moviefile_path)],
```
oruebel:
```suggestion
    external_file=[os.path.relpath(moviefile_path, nwbfile_path)],
```

docs/gallery/domain/images.py
oruebel: I just committed the fix. Thanks for catching this error.

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,251 @@
...
+    starting_frame=[0],
+    rate=1.0,
+    starting_time=0.0,
+)
+
+nwbfile.add_acquisition(behavior_external_file)
+
+####################
+# Static images
+# -------------
```
oruebel, line 171:
```suggestion
# .. note::
#          See the `External Links in NWB and DANDI <https://www.dandiarchive.org/2022/03/03/external-links-organize.html>`_ 
#          guidelines of the DANDI data archive for best practices on how to organize 
#          external files, e.g., movies and images. 
#
# Static images
# -------------
```

docs/gallery/domain/images.py
oruebel: I added a suggestion to address this comment. Resolving this conversation.

docs/gallery/domain/images.py
weiglszonja: Thank you @oruebel! I tried to apply these suggestions.

docs/gallery/domain/images.py
weiglszonja: Thanks for making the changes!

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,270 @@
...
+    os.path.relpath(movie_path, nwbfile_path) for movie_path in moviefiles_path
+]
+timestamps = [0.0, 0.04, 0.07, 0.1, 0.14, 0.16, 0.21]
+behavior_external_file = ImageSeries(
+    name="ExternalFiles",
+    description="Behavior video of animal moving in environment.",
+    unit="n.a.",
+    external_file=external_file,
+    format="external",
+    starting_frame=[0],
```
oruebel: If you use multiple movie files, I believe you would need a starting_frame value for each file.

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,270 @@
...
+# path to the external file(s) on disk. The file(s) path must be relative to the path of the NWB file.
+# Either ``external_file`` or ``data`` must be specified, but not both.
+#
+# If the sampling rate is constant, use ``rate`` and ``strating_time`` to specify time.
+# For irregularly sampled recordings, use ``timestamps`` to specify time for each sample.
+
+external_file = [
+    os.path.relpath(movie_path, nwbfile_path) for movie_path in moviefiles_path
+]
+timestamps = [0.0, 0.04, 0.07, 0.1, 0.14, 0.16, 0.21]
```
oruebel, line 151: I have not used ImageSeries with multiple movie files before, but I think if you use explicit timestamps with movie files you probably need to specify timestamps for all frames instead of just 1 timestamp per movie file. @bendichter thoughts?

weiglszonja: Yes, the `timestamps` should correspond to all frames, this is actually what i intended to do here, is the `timestamps` vector in this example too short to indicate this?

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,270 @@
...
+from PIL import Image
+
+from pynwb import NWBFile, NWBHDF5IO
+from pynwb.base import Images
+from pynwb.image import RGBAImage, RGBImage, GrayscaleImage, OpticalSeries, ImageSeries
+
+# Define file paths used in the tutorial
+import os
+nwbfile_path = os.path.abspath("images_tutorial.nwb")
+moviefiles_path = [
```
oruebel, line 35:
```suggestion
# Collection of example movie files. In this example, each tiff file contains 2 frames 
moviefiles_path = [
```

docs/gallery/domain/images.py
oruebel: What you have here is probably correct and I should have just looked closer at what ``moviefiles_path`` contains. Just to make it a bit clearer I would suggest adding a brief comment here (and/or where ``moviefiles_path`` is defined).

oruebel:
```suggestion
]
# We have 3 movie files each containing multiple frames. We here need to specify the timestamp for each frame. 
timestamps = [0.0, 0.04, 0.07, 0.1, 0.14, 0.16, 0.21]
```

docs/gallery/domain/images.py
```diff
@@ -0,0 +1,270 @@
...
+# External Files
+# ^^^^^^^^^^^^^^
+#
+# External files (e.g. video files of the behaving animal) can be added to the :py:class:`~pynwb.file.NWBFile` by creating
+# an :py:class:`~pynwb.image.ImageSeries` object using the ``external_file`` attribute that specifies the
+# path to the external file(s) on disk. The file(s) path must be relative to the path of the NWB file.
+# Either ``external_file`` or ``data`` must be specified, but not both.
+#
+# If the sampling rate is constant, use ``rate`` and ``strating_time`` to specify time.
+# For irregularly sampled recordings, use ``timestamps`` to specify time for each sample.
```
oruebel:
```suggestion
# For irregularly sampled recordings, use ``timestamps`` to specify time for each sample image.
```

docs/gallery/domain/images.py
oruebel: I'm not sure how many frames are in each movie file, but assuming "file_1.tiff" and "file_2.tiff" contain 2 frames and "file_2.tiff" contains 3 frames, I believe the correct value for starting_frame should be```starting_frame=[0, 2, 6],```

According to the schema of ImageSeries: 

```
        Each external image may contain one or more consecutive frames of the full
        ImageSeries. This attribute serves as an index to indicate which frames each file
        contains, to faciliate random access. The 'starting_frame' attribute, hence,
        contains a list of frame numbers within the full ImageSeries of the first frame
        of each file listed in the parent 'external_file' dataset. Zero-based indexing is
        used (hence, the first element will always be zero). For example, if the
        'external_file' dataset has three paths to files and the first file has 5 frames,
        the second file has 10 frames, and the third file has 20 frames, then this
        attribute will have values [0, 5, 15]. If there is a single external file that
        holds all of the frames of the ImageSeries (and so there is a single element in
        the 'external_file' dataset), then this attribute should have value [0].
```

CodyCBakerPhD: Related to #1318 - if we get an answer on that issue I can close it

oruebel: @CodyCBakerPhD thanks for pointing us to this issue. I just added a response based on my understanding of the schema. I hope this helps clarify things a bit.

weiglszonja: @oruebel thank you for this example! I'll extend the external file section with this description (or a shorter version of it) to make this clear.

docs/gallery/domain/images.py
```diff
@@ -130,12 +130,18 @@
...
-# If the sampling rate is constant, use ``rate`` and ``strating_time`` to specify time.
-# For irregularly sampled recordings, use ``timestamps`` to specify time for each sample image.
+# If the sampling rate is constant, use :py:attr:`~pynwb.base.TimeSeries.rate` and :py:attr:`~pynwb.base.TimeSeries.starting_time` to specify time.
+# For irregularly sampled recordings, use :py:attr:`~pynwb.base.TimeSeries.timestamps` to specify time for each sample image.
+#
+# Each external image may contain one or more consecutive frames of the full :py:class:`~pynwb.image.ImageSeries`.
+# The :py:attr:`~pynwb.image.ImageSeries.starting_frame` attribute serves as an index to indicate which frame
+# each file contains.
+# For example, if the ``external_file`` dataset has three paths to files and the first and the second file have 2 frames,
+# and the third file has 3 frames, then this attribute will have values `[0, 2, 4]`.
```
weiglszonja, line 145: @oruebel, following this example I believe the correct indices are `[0, 2, 4]` and not `[0, 2, 6]`, but let me know if I get it wrong.

oruebel: Yes, based on the example, that looks correct to me.

oruebel (dismissed):
Looks good to me. Thanks for creating this tutorial and for your patients in iterating on edits.

weiglszonja:
@oruebel I learnt a lot from the suggestions, thank you for them! Now only the thumbnail image is missing, do you have an idea how it should look like?

oruebel:
> Now only the thumbnail image is missing, do you have an idea how it should look like?

That's great! The PowerPoint file with the source of all the other thumbnails is here
https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/source/figures/gallery_thumbnails.pptx . Just copy one of the existing thumbnails to a new slide, update the title, and add a small image. For the picture I would maybe either: a) add an actual photo from an experiment or b) just take the mouse from the "Behavior Data" thumbnail and add a clip-art of a camera taking a photo of the mouse. But just have fun and be creative with the image.

weiglszonja:
@oruebel how about this?
gallery_thumbnails_image_data

oruebel:
> how about this?

Looks great!

oruebel (dismissed):

oruebel (approved):
This looks all good to me! Thanks for the hard work!

weiglszonja:
@oruebel @bendichter thank you for the suggestions!

PR #1484, opened 2022-05-25
Fix warnings, use keyword args where possible

### LINKED ISSUE #1189: Use keyword-only arguments in neurodata_type constructors

rly:
Use the just released HDMF 3.3.1. Note that HDMF 3.3 introduces warnings when positional arguments are used in subclasses of many base types. This PR replaces positional args with keyword args in source code and tests throughout. Fix #1189

Also remove old-style `super()` calls and calls to `call_docval_func`

## Checklist


rly:
This looks like a really big PR, but no functionality has changed.

This PR does not introduce warnings for using positional arguments for most NWB neurodata type classes, but sets the foundation for it. That change will be in the next minor release.

A minor bug was fixed in `src/pynwb/io/epoch.py`

There is still a warning about the length of data not matching the length of timestamps when an ImageSeries is created where an external file is used and timestamps are provided. In this case, a placeholder 'data' is set with shape (0,0,0). Again, this is a case where ImageSeries, when not used to store data in the `data` field breaks inheritance rules and causes headache. I think we should revisit creation of a new neurodata type that is just for external file backed `ImageSeries`. See also https://github.com/NeurodataWithoutBorders/nwb-schema/pull/481 and #1220 and #1274

oruebel (approved):

PR #1486, opened 2022-05-30
Avoid raising data timestamps length mismatch warning with external file

weiglszonja:
## Motivation

Minor issue, but storing images (e.g. `ImageSeries`) with external mode raises a `UserWarning` that the length of data does not match the length of timestamps. However when `external_file` is specified (not None), data is not used and therefore this kind of warning should not be raised.

https://github.com/NeurodataWithoutBorders/pynwb/blob/d38cc78bcae086272d3946aaad6165c6bc58b154/src/pynwb/base.py#L184-L185

## How to test the behavior?

New unittest is added at `test_image.ImageSeriesConstructor.test_external_file_no_warning` to test new behavior.

## Checklist


src/pynwb/base.py
```diff
@@ -180,6 +180,9 @@ def __init__(self, **kwargs):
 
             # check that the data and timestamps match
             and (data_shape[0] != timestamps_shape[0])
+
+            # check that data has any values
+            and np.any(args_to_process["data"])
```
bendichter: why not check if `format == "external"`?

bendichter:
from @oruebel: better would be to implement the check as a protected method in `TimeSeries` and overwrite that method in `ImageSeries`

src/pynwb/image.py
```diff
@@ -80,6 +81,10 @@ def __init__(self, **kwargs):
         for key, val in args_to_set.items():
             setattr(self, key, val)
 
+    @staticmethod
+    def _check_data_timestamps_mismatch(data, timestamps):
+        return not np.array_equal(data, ImageSeries.DEFAULT_DATA) and timestamps is not None
+
```
weiglszonja: @rly I have an issue with this, since this check is in the TimeSeries init, I'm not sure how 
to pass the external_file or format as kwargs. I noticed this TODO,
https://github.com/NeurodataWithoutBorders/pynwb/blob/a62c4e951df08e7fff0095789f244513b7d57a7c/src/pynwb/image.py#L76
which is connected to this warning I believe. So instead of overwriting that method with an additional check, I could ignore the warning when external file is provided within the ImageSeries init?

bendichter: does this method need to be static?

weiglszonja: No, but currently that would not help I think, because the check method gets called here
https://github.com/NeurodataWithoutBorders/pynwb/blob/a62c4e951df08e7fff0095789f244513b7d57a7c/src/pynwb/image.py#L77
before setting `self.external_file` with whatever value is in `args_to_set`
https://github.com/NeurodataWithoutBorders/pynwb/blob/a62c4e951df08e7fff0095789f244513b7d57a7c/src/pynwb/image.py#L78-L82

rly (dismissed):

rly (dismissed):

weiglszonja:
Thank you @rly!

bendichter:
@weiglszonja add to changelog please :-)

tests/unit/test_image.py
```diff
@@ -73,6 +75,35 @@ def test_external_file_no_unit(self):
...
+        with warnings.catch_warnings(record=True) as w:
+            ImageSeries(
+                name='test_iS',
+                external_file=['external_file'],
+                format='external',
+                unit='Frames',
+                starting_frame=[0],
+                timestamps=[1, 2, 3, 4]
+            )
+            self.assertEqual(w, [])
```
bendichter, line 105: Can we also add a test to ensure that using an external file and a `rate` instead of `timestamps` does not raise a warning?

weiglszonja: Sure, I added a new unittest.

bendichter (approved):

PR #1493, opened 2022-07-04
Improve package modularity

CodyCBakerPhD:
## Motivation

It can be ambiguous to have callable methods in the `__init__` of a package whenever those methods also share a name with a submodule exposed at the same level. *E.g.*, currently `pynwb.validate` is both a submodule with methods you can import from and a method intended to be called on an open `io` object.

It also makes the package a bit harder to navigate, as it is not intuitive where certain functions might be found, such as namespace-related items (more properly  belonging to the `spec` module?) and IO functions (like `NWBHDF5IO`, should perhaps be in the `io` submodule?). It's also not a very common practice to include more than simple import routing, logging, or necessary package-level initialization procedures within the `__init__.py` files.

This PR moves all the actual code bits from `__init__` to a new module called `tools`, _however_, <ins>the global import structure should remain unchanged</ins>. I understand this is the most likely point to cause concern for this type of PR, which is why I'm explicitly testing this structure: https://github.com/NeurodataWithoutBorders/pynwb/blob/modularize_init_1/tests/back_compat/test_imports.py (though I can certainly add deeper orders of nesting to that).

This is just a first draft suggestion, I'm open to discussion of where each particular item should go and/or what structure would be most convenient for everyone. Also open to not doing this at all, but I strongly recommend it and it will make some subsequent PR's I have in the works for improvements to the `validate` functionality _a lot_ easier to set up.

## How to test the behavior?

A new test/class is added to the `test_backcompat` set of tests that checks the import structure matches, at least up to newly introduced exceptions.

## Checklist


src/pynwb/globals.py
```diff
@@ -0,0 +1,4 @@
+"""Definitions of global values used across the project."""
+
+CORE_NAMESPACE = 'core'
+__core_ns_file_name = 'nwb.namespace.yaml'
```
CodyCBakerPhD, line 4: Alternative: also not unusual to call such a collection `consts` such as in https://github.com/dandi/dandi-cli/blob/master/dandi/consts.py

Basically just a collection of all the hand-typed or hard-coded things local to the package

bendichter:
I agree I'm not a fan of the inits being so big, but I don't know if moving all of these functions to `tools` is the best approach. I think a better approach would be to take it module-by-module and see if we can find the most appropriate place for each function.

CodyCBakerPhD:
> I agree I'm not a fan of the inits being so big, but I don't know if moving all of these functions to tools is the best approach. I think a better approach would be to take it module-by-module and see if we can find the most appropriate place for each function.

Absolutely! The `tools` module is just a temporary home and proof of concept, we can go through each function one by one and decide the best place for them.

CodyCBakerPhD:
List of method/variabless and current proposed destinations. `???` indicate I'm unsure where the best place might be.

``
<ins>Globals</ins>

`__resources` -> spec? (followed by loading of namespaces from the resources)

`__NS_CATALOG ` -> `spec`

`__TYPE_MAP ` -> `spec`

`hdmf_typemap` (not capitalized?) -> `spec`


Last two here are purely hand-coded strings

`CORE_NAMESPACE` -> `globals`

`__core_ns_file_name ` -> `globals`


<ins>Methods</ins>

`__get_resources` -> `spec`?

`_get_resources` -> `legacy`

`get_type_map` -> `spec`?

`get_manager` -> `spec`?

`load_namespaces` -> `spec`

`available_namespaces` -> `spec`

`register_class` -> `registration`?

`register_map` -> `registration`?

`get_class` -> ???

`validate` -> `validation`

`NWBHDF5IO` -> `io`

`export` -> `io` ?

PR #1495, opened 2022-07-07
make NWB File Basics the first tutorial in the list

bendichter:


docs/source/conf.py
```diff
@@ -61,6 +63,16 @@
...
 from sphinx_gallery.sorting import ExampleTitleSortKey
 
+
+class CustomSorter(sphinx_gallery.sorting._SortKey):
+
+    def __call__(self, filename):
+        title = ExampleTitleSortKey(self.src_dir)(filename)
+        if title == "NWB File Basics":
+            return ''  # make this first
+        return title
```
oruebel: It looks like ``CustomSorter`` just modifies the behavior of ``ExampleTitleSortKey``. As such, I think the code would be more explicit and readable if ``CustomSorter`` would inherit from ``ExampleTitleSortKey`` directly. Also, to avoid confusion, I would suggest using a more explicit name for the class, e.g., ``CustomSphinxGallerySectionSortKey``

```suggestion
class CustomSorter(ExampleTitleSortKey):
   """
   Define the key to be used to sort sphinx galleries sections
   """

    def __call__(self, filename):
        """
        Return the key to be used for sorting the given sphinx gallery file.
        
        :param filename: Name of the shpinx gallery file 
        """
        title = super()(filename)
        if title == "NWB File Basics":
            return ''  # make this first
        return title
```

oruebel:
@bendichter what are your thoughts on using ``sphinx_gallery.sorting.ExplicitOrder``? This would provide explicit control of the ordering of all galleries, which seems to be what you are want here. The one main downside I see for using an explicit order is that it will require updating the order in ``conf.py`` every time we add a tutorial or change a title.

bendichter:
@oruebel is that possible? My reading of [the docs](https://sphinx-gallery.github.io/stable/configuration.html#sorting-gallery-examples) was that `ExplicitOrder` worked for organizing subsections but not for organizing the gallery examples themselves. 

Also, this was really the only change I wanted. I think the rest can be ordered alphabetically.

docs/source/conf.py
```diff
@@ -13,6 +13,8 @@
 
 import sys
 import os
+
+import sphinx_gallery.sorting
```
bendichter:
```suggestion
```

oruebel:
> .. is that possible? 

I believe yes. If I understand these classes correctly, then all the classes do is return the key that should be used when calling ``sort``. The key can be (as far as I can tell) any type that the standard sort function can sort (e.g., ``str`` or ``int``). As such, using ``ExplictOrder`` should work also for sorting galleries, not just subsections. The only quirk is that ``ExplictOrder`` only uses a single flat ordered list (i.e., it does not have a ``self.src_dir`` to distinguish between subsections), so when using it as is, would probably mean that one would need to list full paths, rather than being able to just list filenames. However, implementing our own ``ExplicitOrder`` for galleries that uses a nested dictionary for ``[subsection][galleryfile]`` should be fairly simple.

bendichter:
@oruebel I'd rather move forward with this for now. I like that most of them are alphabetical. If we find other ways we want to reorder the gallery then we can look into specifying explicit order.

oruebel:
Sounds good. Just for reference, here how I believe it would look to implement an explicit order for galleries. It's basically just a dict of lists and then looking up the index in the list. 

```
ExplicitGallerySortKey(sphinx_gallery.sorting._SortKey)

    # Define an ordered list of galleries for all subsections
    GALLERY_ORDER = {
             'basics/' : ['nwbfile.py', ....],
             'advanced_io': [...],
    }

    def __init__(self, src_dir):
        self.src_dir = src_dir

    def __call__(self, filename):
        # Get the ordered list of gallery files for the current source dir
        if self.src_dir not in self.GALLERY_ORDER:
             raise ConfigError('...')
        curr_order = GALLERY_ORDER [self.src_dir]
        # Look up the index of filename 
        if filename in curr_order:
            return curr_order.index(item)
        else:
            raise ConfigError('...')
```

oruebel (approved):

rly (approved):

PR #1499, opened 2022-07-13
S3fs

bendichter:
## Motivation

Relax args to enable s3fs

Requires https://github.com/hdmf-dev/hdmf/pull/746

## How to test the behavior?
```python
import s3fs
import pynwb
import h5py

fs = s3fs.S3FileSystem(anon=True)

f = fs.open("s3://dandiarchive/blobs/43b/f3a/43bf3a81-4a0b-433f-b471-1f10303f9d35", 'rb')
file = h5py.File(f)
io = pynwb.NWBHDF5IO(file=file, load_namespaces=True)

io.read()
```

## to do
- [x] change log
- [x] tutorial
- [ ] test

## Checklist


bendichter:
@rly I wanted to check with you with this PR and the hdmf PR before moving forward with the docs, tests, and change log

src/pynwb/__init__.py
```diff
@@ -200,17 +200,18 @@ def validate(**kwargs):
 
 class NWBHDF5IO(_HDF5IO):
 
-    @docval({'name': 'path', 'type': (str, Path), 'doc': 'the path to the HDF5 file'},
+    @docval({'name': 'path', 'type': (str, Path), 'doc': 'the path to the HDF5 file', 'default': None},
             {'name': 'mode', 'type': str,
-             'doc': 'the mode to open the HDF5 file with, one of ("w", "r", "r+", "a", "w-", "x")'},
+             'doc': 'the mode to open the HDF5 file with, one of ("w", "r", "r+", "a", "w-", "x")',
+             'default': 'r'},
```
oruebel, line 206: Does this change the behavior of ``NWBHDF5IO`` when ``path`` and ``mode`` are used as positional arguments?  I think the answer is no, since all other arguments already had defaults and since this adds defaults for both ``path`` and ``mode``.  Overall the changes look fine to me, however, I just wanted to confirm as this question affects what version for release this change should go to.

bendichter: The only thing we need to watch out for here is if we have an arg without a default *after* an arg with a default. In that case the arg without a default will be an earlier positional arg. This actually came up in the partner PR. I had to move manager later so that it maintained the same position. It shouldn't be a problem here though.

oruebel: Sounds good

rly:
> @rly I wanted to check with you with this PR and the hdmf PR before moving forward with the docs, tests, and change log

This approach looks good to me. Can you add a comment about Windows incompatibility?

docs/gallery/advanced_io/streaming.py
```diff
@@ -52,6 +57,42 @@
...
+#   io = pynwb.NWBHDF5IO(file=file, load_namespaces=True)
+#
+#   io.read()
+#
+# The above snippet opens an arbitrary file on DANDI. You can use the ``DandiAPIClient`` to find the s3 path,
+# but you will need to adjust this url to give it a prefix of "s3://dandiarchive/" as shown above.
+#
+# The s3fs approach has the advantage of being more robust that ROS3. Sometimes s3 requests are interrupted,
+# and s3fs has internal mechanisms to retry these requests automatically, whereas ROS3 does not. However, it may not
+# be available on all platforms.
```
bendichter:
```suggestion
# be available on all platforms. s3fs does not currently work for Windows.
```
```suggestion
# be available on all platforms.
```

oruebel (approved):

PR #1503, opened 2022-07-13
add tests for NWBData

bendichter:
add tests for NWBData, including tests for append and extend, which [appear to be untested](https://app.codecov.io/gh/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/core.py)


rly:
This looks good for a base test, but we can increase the test coverage of `NWBData`. I'd be happy to do that in this or a separate PR.

rly (approved):

PR #1505, opened 2022-07-13
Test nwbdata

bendichter:
fix (how I would expect) append and extend are supposed to work for np arrays, and add tests

### DIFF
--- a/src/pynwb/core.py
+++ b/src/pynwb/core.py
@@ -71,7 +71,7 @@ class NWBData(NWBMixin, Data):
         if isinstance(self.data, list):
             self.data.append(arg)
         elif isinstance(self.data, np.ndarray):
-            self.__data = np.vstack((self.__data, [arg]))
+            self.__data = np.concatenate((self.__data, [arg]))
         else:
             msg = "NWBData cannot append to object of type '%s'" % type(self.__data)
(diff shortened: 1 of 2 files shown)


src/pynwb/core.py
```diff
@@ -71,7 +71,7 @@ def append(self, arg):
         if isinstance(self.data, list):
             self.data.append(arg)
         elif isinstance(self.data, np.ndarray):
-            self.__data = np.append(self.__data, [arg])
+            self.__data = np.vstack((self.__data, [arg]))
```
rly: I think this will not work for 1-dimensional data. 

```python
>>> import numpy as np
>>> data = np.array([1, 2, 3])
>>> np.vstack((data, [[4, 5, 6]]))
array([[1, 2, 3],
       [4, 5, 6]])
>>> np.append(data, [[4, 5, 6]])
array([1, 2, 3, 4, 5, 6])
```

rly: The above code is more related to the `extend` case below.

bendichter: oh true ok I'll think of a more general solution

rly (approved):

rly:
LGTM. Thanks!

PR #1507, opened 2022-07-14
SpikeEventSeries: add test for timestamp check

bendichter:


rly (dismissed):

rly (dismissed):

rly (dismissed):

rly:
This was amusing: one test tested that `ElectrodeGroup(...)` would raise a TypeError, but the code in `ElectrodeGroup.__init__` generated a ValueError. That's odd. And the test passed! Why? A TypeError was raised because of an error in the code that created the ValueError. So fixing the test to catch a ValueError made the test fail until I also fixed the code that created the ValueError.

```
if args_to_set['position'] and len(args_to_set['position']) != 3:
>           raise ValueError('ElectrodeGroup position argument must have three elements: x, y, z, but received: %s'
                             % args_to_set['position'])
E           TypeError: not all arguments converted during string formatting
```

rly (approved):

PR #1516, opened 2022-07-20
Enh/ `ImageSeries`: add checks when `external_file` is used

### LINKED ISSUE #1510: [Feature]: `ImageSeries`: check `starting_frame` when `external_file` is used

Related to #1318 and #1470, when  `external_file` is used in `ImageSeries` a new check should be added which ensures that
a) `starting_frame` is provided if `external_file` `is` used,
b) `starting_frame` has the same length as `external_file`. 

Currently `starting_frame` is set to `[0]` by default, which is fine for a single external file but is incorrect for multiple external files. 

_No response_

A new check method in  ImageSeries.__init__ that checks if
a) `starting_frame` is provided if external_file is used
b) `starting_frame` has the same length as external_file. 

Yes.

### Code of Conduct

weiglszonja:
## Motivation

This PR is a continuation of the issue described in #1510 by raising a `ValueError` when `starting_frame` does not have the same length as `external_file`. 

- Added a method to `NWBMixin` that only raises an error when a check is violated on instance creation,
  otherwise throws a warning when reading from a file. 

The new checks is used with this method to ensure that that files with invalid data can be read, but prohibits the user from creating new instances when these checks are violated. 

New checks:
- The length of `starting_frame` must match the length of `external_file` when `external_file` is specified
- Format must be `'external'` when `external_file` is specified
- `data` must be empty array when `external_file` is specified
`ValueError` is raised when a check is violated on instance creation,
but only warns when file is read from disk.

Default values:
- `starting_frame` is `[0]` when `external_file` has a length of 1
- `format` is `'external'` when format was originally None and `external_file` is specified (`DeprecationWarning` is raised)

          

Fix #1510 

## How to test the behavior?

Unittests are added in `tests/unit/test_image.py`

## Checklist

### DIFF
--- a/src/pynwb/image.py
+++ b/src/pynwb/image.py
@@ -51,8 +51,8 @@ class ImageSeries(TimeSeries):
              'doc': 'Path or URL to one or more external file(s). Field only present if format=external. '
                     'Either external_file or data must be specified (not None), but not both.', 'default': None},
             {'name': 'starting_frame', 'type': Iterable,
-             'doc': 'Each entry is a frame number that corresponds to the first frame of each file in external_file. '
-                    'This serves as an index to what frames each file contains.', 'default': None},
+             'doc': 'Each entry is a frame number that corresponds to the first frame of each file '
+                    'listed in external_file within the full ImageSeries.', 'default': None},
             {'name': 'bits_per_pixel', 'type': int, 'doc': 'DEPRECATED: Number of bits per image pixel',
              'default': None},
             {'name': 'dimension', 'type': Iterable,
--- a/src/pynwb/image.py
+++ b/src/pynwb/image.py
@@ -84,8 +84,8 @@ class ImageSeries(TimeSeries):
                 args_to_set['starting_frame'] = [0]
         if external_file_shape != starting_frame_shape:
             warnings.warn(
-                "%s '%s': The number of frame indices in 'starting_frame' should have the same length as 'external_file'."
-                % (self.__class__.__name__, name)
+                "%s '%s': The number of frame indices in 'starting_frame' should have the same length "
+                "as 'external_file'." % (self.__class__.__name__, name)
             )
 
         super().__init__(**kwargs)
--- a/src/pynwb/image.py
+++ b/src/pynwb/image.py
@@ -77,8 +77,13 @@ class ImageSeries(TimeSeries):
         if unit is None:
             kwargs['unit'] = ImageSeries.DEFAULT_UNIT
 
-        if args_to_set['external_file'] is not None and args_to_set['starting_frame'] is None:
-            args_to_set['starting_frame'] = [0] if len(args_to_set['external_file']) == 1 else None
+        if (
+            args_to_set["external_file"] is not None
+            and args_to_set["starting_frame"] is None
+        ):
+            args_to_set["starting_frame"] = (
+                [0] if len(args_to_set["external_file"]) == 1 else None
+            )
 
         super().__init__(**kwargs)
 
@@ -97,12 +102,13 @@ class ImageSeries(TimeSeries):
                 "Setting a default value for 'format' is deprecated and will be changed "
                 "to raising a ValueError in the next release."
                 % (self.__class__.__name__, self.name),
-                DeprecationWarning
+                DeprecationWarning,
             )
 
         if not self._check_external_file_format():
             raise ValueError(
-                "%s '%s': Format must be 'external' when external_file is specified." % (self.__class__.__name__, name)
+                "%s '%s': Format must be 'external' when external_file is specified."
+                % (self.__class__.__name__, name)
             )
 
         if not self._check_external_file_data():
@@ -114,7 +120,8 @@ class ImageSeries(TimeSeries):
         if not self._check_image_series_dimension():
             warnings.warn(
                 "%s '%s': Length of data does not match length of timestamps. Your data may be transposed. "
-                "Time should be on the 0th dimension" % (self.__class__.__name__, self.name)
+                "Time should be on the 0th dimension"
+                % (self.__class__.__name__, self.name)
             )
 
     def _check_external_file_starting_frame_length(self):
@@ -132,11 +139,11 @@ class ImageSeries(TimeSeries):
         Change the format to 'external' when external_file is specified.
         """
         if (
-                self.data.shape[0] == 0 and
-                self.external_file is not None and
-                self.format is None
+            self.data.shape[0] == 0
+            and self.external_file is not None
+            and self.format is None
         ):
-            self.format = 'external'
+            self.format = "external"
             return True
 
         return False
--- a/src/pynwb/image.py
+++ b/src/pynwb/image.py
@@ -96,7 +96,7 @@ class ImageSeries(TimeSeries):
         if not self._check_external_file_starting_frame_length():
             msg = ("%s '%s': The number of frame indices in 'starting_frame' should have "
                    "the same length as 'external_file'." % (self.__class__.__name__, name))
-            self._raise_error_when_check_is_violated(error_msg=msg)
+            self._error_on_new_warn_on_construct(error_msg=msg)
 
         if self._change_external_file_format():
             warnings.warn(
@@ -112,14 +112,14 @@ class ImageSeries(TimeSeries):
                 "%s '%s': Format must be 'external' when external_file is specified."
                 % (self.__class__.__name__, name)
             )
-            self._raise_error_when_check_is_violated(error_msg=msg)
+            self._error_on_new_warn_on_construct(error_msg=msg)
 
         if not self._check_external_file_data():
             msg = (
                 "%s '%s': Either external_file or data must be specified (not None), but not both."
                 % (self.__class__.__name__, name)
             )
-            self._raise_error_when_check_is_violated(error_msg=msg)
+            self._error_on_new_warn_on_construct(error_msg=msg)
 
         if not self._check_image_series_dimension():
             warnings.warn(
@@ -127,19 +127,6 @@ class ImageSeries(TimeSeries):
                 "Time should be on the 0th dimension"
                 % (self.__class__.__name__, self.name)
             )
-
-    def _raise_error_when_check_is_violated(self, error_msg):
-        """
-        Raise an error when a check is violated on instance creation.
-        To ensure backwards compatibility, this method throws a warning
-        instead of raising an error when reading from a file, ensuring that
-        files with invalid data can be read.
-        """
-        if error_msg is None:
-            return
-        if not self._in_construct_mode:
-            raise ValueError(error_msg)
-        warnings.warn(error_msg)
 
     def _check_external_file_starting_frame_length(self):
         """
--- a/src/pynwb/image.py
+++ b/src/pynwb/image.py
@@ -115,6 +115,8 @@ class ImageSeries(TimeSeries):
         """
         Check that the number of frame indices in 'starting_frame' matches the number of files in 'external_file'.
         """
+        if self.external_file is None:
+            return True
         external_file_shape = get_data_shape(self.external_file)
         starting_frame_shape = get_data_shape(self.starting_frame)
         return external_file_shape == starting_frame_shape
--- a/src/pynwb/image.py
+++ b/src/pynwb/image.py
@@ -107,7 +107,8 @@ class ImageSeries(TimeSeries):
 
         if not self._check_external_file_data():
             raise ValueError(
-                "%s '%s': Either external_file or data must be specified (not None), but not both." % (self.__class__.__name__, name)
+                "%s '%s': Either external_file or data must be specified (not None), but not both."
+                % (self.__class__.__name__, name)
             )
 
         if not self._check_image_series_dimension():
--- a/src/pynwb/image.py
+++ b/src/pynwb/image.py
@@ -82,7 +82,7 @@ class ImageSeries(TimeSeries):
         if external_file_shape is not None and starting_frame_shape is None:
             if len(external_file_shape) == 1:
                 args_to_set['starting_frame'] = [0]
-        if external_file_shape != starting_frame_shape:
+        if external_file_shape != get_data_shape(args_to_set['starting_frame']):
             warnings.warn(
                 "%s '%s': The number of frame indices in 'starting_frame' should have the same length "
                 "as 'external_file'." % (self.__class__.__name__, name)
--- a/src/pynwb/image.py
+++ b/src/pynwb/image.py
@@ -77,27 +77,33 @@ class ImageSeries(TimeSeries):
         if unit is None:
             kwargs['unit'] = ImageSeries.DEFAULT_UNIT
 
-        external_file_shape = get_data_shape(args_to_set['external_file'])
-        starting_frame_shape = get_data_shape(args_to_set['starting_frame'])
-        if external_file_shape is not None and starting_frame_shape is None:
-            if len(external_file_shape) == 1:
-                args_to_set['starting_frame'] = [0]
-        if external_file_shape != get_data_shape(args_to_set['starting_frame']):
-            raise ValueError(
-                "%s '%s': The number of frame indices in 'starting_frame' should have the same length "
-                "as 'external_file'." % (self.__class__.__name__, name)
-            )
+        if args_to_set['external_file'] is not None and args_to_set['starting_frame'] is None:
+            args_to_set['starting_frame'] = [0] if len(args_to_set['external_file']) == 1 else None
 
         super().__init__(**kwargs)
 
         for key, val in args_to_set.items():
             setattr(self, key, val)
 
+        if not self._check_external_file_starting_frame_length():
+            raise ValueError(
+                "%s '%s': The number of frame indices in 'starting_frame' should have the same length "
+                "as 'external_file'." % (self.__class__.__name__, name)
+            )
+
         if not self._check_image_series_dimension():
             warnings.warn(
                 "%s '%s': Length of data does not match length of timestamps. Your data may be transposed. "
                 "Time should be on the 0th dimension" % (self.__class__.__name__, self.name)
             )
+
+    def _check_external_file_starting_frame_length(self):
+        """
+        Check that the number of frame indices in 'starting_frame' matches the number of files in 'external_file'.
+        """
+        external_file_shape = get_data_shape(self.external_file)
+        starting_frame_shape = get_data_shape(self.starting_frame)
+        return external_file_shape == starting_frame_shape
 
     def _check_time_series_dimension(self):
         """Override _check_time_series_dimension to do nothing.
--- a/src/pynwb/image.py
+++ b/src/pynwb/image.py
@@ -79,8 +79,6 @@ class ImageSeries(TimeSeries):
 
         if args_to_set['external_file'] is not None and args_to_set['starting_frame'] is None:
             args_to_set['starting_frame'] = [0] if len(args_to_set['external_file']) == 1 else None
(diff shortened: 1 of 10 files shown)


weiglszonja:
There are tests in `tests/unit/test_ophys.py` that are affected by this change,
for instance the `ImageSeries` used in many tests (actually in integration tests as well) define a single `external_file` but the corresponding `starting_frame` has a different length:
https://github.com/NeurodataWithoutBorders/pynwb/blob/b11bffcb752870b810952355affe2c4bf542d91e/tests/unit/test_ophys.py#L46-L55

@oruebel, should I change all of these or what is the best way to proceed?

oruebel:
> should I change all of these or what is the best way to proceed?

Yes, I think updating the tests is the right approach. Looking at those tests, I believe the starting_frame is indeed set incorrectly in those tests.

src/pynwb/image.py
```diff
@@ -65,17 +71,25 @@ def __init__(self, **kwargs):
...
         if unit is None:
             kwargs['unit'] = ImageSeries.DEFAULT_UNIT
 
+        external_file_shape = get_data_shape(args_to_set['external_file'])
+        starting_frame_shape = get_data_shape(args_to_set['starting_frame'])
+        if external_file_shape is not None and starting_frame_shape is None:
+            if len(external_file_shape) == 1:
+                args_to_set['starting_frame'] = [0]
+        if external_file_shape != get_data_shape(args_to_set['starting_frame']):
+            warnings.warn(
```
oruebel: I would suggest to change this to an error, rather than just a warning

```suggestion
            raise ValueError(
```

tests/unit/test_image.py
```diff
@@ -117,6 +125,39 @@ def test_dimension_warning_external_file_with_rate(self):
...
+    def test_external_files_starting_frame_warning(self):
+        """Test that a warning is raised when the length of starting_frame
+        is not the same as the length of external_file."""
+        for starting_frame in [None, [0], [1, 2, 3]]:
+            with self.subTest():
+                msg = (
+                    "ImageSeries 'test_iS': The number of frame indices in "
+                    "'starting_frame' should have the same length as 'external_file'."
+                )
+                with self.assertWarnsWith(UserWarning, msg):
```
oruebel:
```suggestion
                with self.assertRaisesWith(ValueError, msg):
```

src/pynwb/image.py
```diff
@@ -65,17 +71,25 @@ def __init__(self, **kwargs):
...
+                "%s '%s': The number of frame indices in 'starting_frame' should have the same length "
+                "as 'external_file'." % (self.__class__.__name__, name)
+            )
```
weiglszonja: comments from @bendichter :
- this should be moved to a private function
- when `starting_frame` is `None`, we should have `[0] * len(args_to_set['external_file'])` (setting default not just for a signle external file)
- we should also have a check that ensures that `external_file` is only used when `format` is `"external"`

oruebel: > * when `starting_frame` is `None`, we should have `[0] * len(args_to_set['external_file'])` (setting default not just for a signle external file)

Unfortunately I don't think this is possible for multiple external files. Setting ``starting_frame = [0] * len(args_to_set['external_file'])``  (i.e., setting it to ``[0, 0, 0, 0, ...]``)  is unfortunately not the correct value in almost all cases as it would mean that none of the external files contain any frames. To set the ``starting_frame`` correctly for multiple external files we would need to know how many frames are in each external file and then you could calculate starting_frame as the cummulative sum (with leading zero) of those values. 

If we wanted to do this, we could add an optional argument along the lines of ``frames_per_external_file`` so that we can calculate the ``starting_frame`` for the user. Specifying the number of frames per external file rather than having to calculate the starting_frame could be more intuitive for users. 

> this should be moved to a private function

sounds good

>  we should also have a check that ensures that `external_file` is only used when `format` is `"external"`

sounds good

bendichter: > Setting starting_frame = [0] * len(args_to_set['external_file']) (i.e., setting it to [0, 0, 0, 0, ...]) is unfortunately not the correct value in almost all cases as it would mean that none of the external files contain any frames.

wait, what? @oruebel I think you and I have different ideas of what `starting_frame` means

oruebel: According to the schema, ``starting_frame`` is described as:

```
Each external image may contain one or more consecutive frames 
of the full ImageSeries. This attribute serves as an index to indicate
 which frames each file contains, to faciliate random access. 
The ‘starting_frame’ attribute, hence, contains a list of frame numbers 
within the full ImageSeries of the first frame of each file listed in the
parent ‘external_file’ dataset. Zero-based indexing is used (hence, 
the first element will always be zero). For example, if the ‘external_file’ 
dataset has three paths to files and the first file has 5 frames, the second 
file has 10 frames, and the third file has 20 frames, then this attribute will 
have values [0, 5, 15]. If there is a single external file that holds all of the 
frames of the ImageSeries (and so there is a single element in the 
‘external_file’ dataset), then this attribute should have value [0].
```

I.e., ``starting_frame`` describes for each ``external_file`` what index the first frame in that file corresponds to in the whole image series. In this way, if we want to access a particular frame,  ``starting_frame`` allows us to identify which ``external_file`` contains that frame and what the index of that frame within the external file is. Without this information one would need to open all external files to figure out how many frames each contains in order to find a particular frame.

bendichter: oh, I see. OK I guess I was mistaken about what `starting_frame` was doing.

tests/integration/hdf5/test_image.py
```diff
@@ -15,8 +15,8 @@ def setUpContainer(self):
             data=np.ones((3, 3, 3)),
             unit='unit',
             external_file=['external_file'],
-            starting_frame=[1, 2, 3],
-            format='tiff',
+            starting_frame=[0],
+            format='external',
             timestamps=[1., 2., 3.],
             device=self.dev1,
         )
```
weiglszonja: I would also remove the `data` argument, as there should be either `external_file` or `data` specified but not both (right?), so far we only check if none of them is supplied. should I add a check for that too, or are there cases when this is correct?

oruebel: > I would also remove the `data` argument, as there should be either `external_file` or `data`

Agreed

> should I add a check for that too, or are there cases when this is correct?

I would add a check for this as well.

weiglszonja:
@oruebel Seems like back compatibility tests are failing, how can I resolve them?

src/pynwb/image.py
```diff
@@ -79,6 +79,8 @@ def __init__(self, **kwargs):
 
         if args_to_set['external_file'] is not None and args_to_set['starting_frame'] is None:
             args_to_set['starting_frame'] = [0] if len(args_to_set['external_file']) == 1 else None
+        if args_to_set['external_file'] is not None and args_to_set['format'] is None:
+            args_to_set['format'] = 'external'
```
weiglszonja: I'm not sure that we want a default value this case.

oruebel: As long as ``data`` is ``None`` and ``external_file`` is ``not None``, this behavior should be Ok. However, I agree that the more consistent behavior would be to just raise a ValueError here. However, since this changes behavior of the API, what I would suggest for now is to just add a warning that indicates: 1) that the value for format has been changed, and b) that this behavior is deprecated and will be changed in the next released. We can then change the behavior to a ValueError in the next main release after that.

oruebel:
> @oruebel Seems like back compatibility tests are failing, how can I resolve them?

I'll need to have a look at the tests. @rly thoughts?

weiglszonja:
@oruebel, These are the back compatibility tests that are failing:
```
test_read (back_compat.test_read.TestReadOldVersions) (file='1.5.1_imageseries_no_data.nwb')
test_read (back_compat.test_read.TestReadOldVersions) (file='1.5.1_imageseries_no_unit.nwb')
test_read_imageseries_no_data (back_compat.test_read.TestReadOldVersions)
test_read_imageseries_no_unit (back_compat.test_read.TestReadOldVersions)
```
They all fail due to:
`ImageSeries 'test_imageseries': The number of frame indices in 'starting_frame' should have the same length as 'external_file'.`

Is it an option to recreate these two files with a correct `ImageSeries`? Or it breaks the point of back compatibility?

rly:
> Or it breaks the point of back compatibility?
Yes, it breaks the point and is in part why we have these tests. The latest pynwb should be able to read older NWB 2 files that fail this check. I suggest changing the error to a warning.

oruebel:
> Yes, it breaks the point and is in part why we have these tests. The latest pynwb should be able to read older NWB 2 files that fail this check. I suggest changing the error to a warning.

@rly is it possible to have this as a warning on read and as an error when creating a new file (e.g., by having the BuildManager or ObjectMapper set a flag)? To be honest, old files that raise this error means that those files are incorrect and while we should still support reading those files, it would be ideal if we could enforce that new files can't be created in such a state (i.e., ideally we should warn on read and  error on write).

weiglszonja:
@oruebel, I think this would be great, but maybe what we can also do to avoid creating new files with incorrect data is (following @CodyCBakerPhD suggestion): (1) warning instead of error here, (2) adding a check to `nwbinspector` too an error is raised?
but if this warn on read/error on write can be also done, it would probably be better for also those cases when users are not using the inspector to validate their files.

oruebel:
> adding a check to `nwbinspector` too an error is raised?

Adding a check to the inspector I think is a good idea either way.

> but if this warn on read/error on write can be also done, it would probably be better

Yes, I agree. It should be possible, we just need to decide on the correct design. One way I could see is to have a constructor argument to define whether to warn or raise an error with a default setting to error. In the ObjectMapper we could then set the constructor argument to warn instead. In this way, if a user explicitly creates the object, the default behavior is to error and any time the object mapper is involved (i.e., on read) the behavior would be to just warn. Using this approach a user could still chose to turn the error into a warning, but maybe having this as an explicit behavior is Ok. What do you think @rly? I thought we did something like this already for another class (or at least considered doing this) but I can't remember which class it is.

oruebel:
> but if this warn on read/error on write can be also done, it would probably be better

We working to support this as part of https://github.com/hdmf-dev/hdmf/issues/750 We could either make this a warning for now and move to this approach once it is available or wait until it is available in HDMF. @rly what would you suggest to do?

rly:
Let's wait until this is available in HDMF.

oruebel:
@weiglszonja with https://github.com/hdmf-dev/hdmf/pull/751 merged and released in HDMF 3.4 as of yesterday we can now implement the variant to warn when reading from file and raise error when creating new data. To do this you will need update your HDMF and then you can simply do the following:

```
if self._in_construct_mode:  # Reading data from disk, i.e., BuildManger creates the instance
    warnings.warn("....) 
else:  # User creates the instance
    raise ValueError("....")
```

The ``_in_construct_mode`` variable is automatically set to ``True`` by the ``BuildManager`` before ``__init__`` is being called and reset back to ``False`` after ``__init__`` is done, i.e., you can just use the ``_in_construct_mode`` as a regular instance variable but you don't need to modify it.

src/pynwb/image.py
```diff
@@ -73,19 +79,104 @@ def __init__(self, **kwargs):
...
+                % (self.__class__.__name__, self.name)
             )
 
+    def _raise_error_when_check_is_violated(self, error_msg):
+        """Raise an error when a check is violated on instance creation.
+        To ensure backwards compatibility, this method throws a warning
+        instead of raising an error when reading from a file."""
+        if not self._in_construct_mode:
+            raise ValueError(error_msg)
+        warnings.warn(error_msg)
```
weiglszonja: @oruebel let me know what you think about this, I added tests in [tests/back_compat/test_read.py](https://github.com/NeurodataWithoutBorders/pynwb/pull/1516/files#diff-7f35c9a018fa89011cfaf801cfcc8747edbeefc94ab48e6ab0862469407f57e2) to see if warnings are raised. I'm not sure how to fix coverage for L136, I thought the back compat tests are sufficient for testing the warnings.

src/pynwb/image.py
```diff
@@ -73,19 +79,104 @@ def __init__(self, **kwargs):
         if unit is None:
             kwargs['unit'] = ImageSeries.DEFAULT_UNIT
 
+        if (
```
oruebel, line 83:
```suggestion
        # If a single external_file is given then set starting_frame  to [0] for backward compatibility 
        if (
```

src/pynwb/image.py
```diff
@@ -73,19 +79,104 @@ def __init__(self, **kwargs):
...
                 "%s '%s': Length of data does not match length of timestamps. Your data may be transposed. "
-                "Time should be on the 0th dimension" % (self.__class__.__name__, self.name)
+                "Time should be on the 0th dimension"
+                % (self.__class__.__name__, self.name)
             )
 
+    def _raise_error_when_check_is_violated(self, error_msg):
+        """Raise an error when a check is violated on instance creation.
+        To ensure backwards compatibility, this method throws a warning
+        instead of raising an error when reading from a file."""
```
oruebel:
```suggestion
        """
        Raise an error when a check is violated on instance creation.
        To ensure backwards compatibility, this method throws a warning
        instead of raising an error when reading from a file, ensuring that
        files with invalid data can be read.
        """
```

src/pynwb/image.py
```diff
@@ -73,19 +79,104 @@ def __init__(self, **kwargs):
...
+
         if not self._check_image_series_dimension():
             warnings.warn(
                 "%s '%s': Length of data does not match length of timestamps. Your data may be transposed. "
-                "Time should be on the 0th dimension" % (self.__class__.__name__, self.name)
+                "Time should be on the 0th dimension"
+                % (self.__class__.__name__, self.name)
             )
 
+    def _raise_error_when_check_is_violated(self, error_msg):
```
oruebel: I think it would make sense to move this function to the ``NWBMixin``, as this pattern of warning on read and error on new creation will be more broadly useful.

https://github.com/NeurodataWithoutBorders/pynwb/blob/2de33c032eebc46800191bc24a72976bca30aa6f/src/pynwb/core.py#L23-L33

oruebel: Instead of ``_raise_error_when_check_is_violated`` I would suggest renaming the function to ``_error_on_new_warn_on_construct`` just to be more explicit on the purpose of the function. T

src/pynwb/image.py
oruebel:
```suggestion
        if error_msg is None:
            return
        if not self._in_construct_mode:
            raise ValueError(error_msg)
        warnings.warn(error_msg)
```

src/pynwb/image.py
```diff
@@ -73,19 +79,104 @@ def __init__(self, **kwargs):
...
+            )
+            self._raise_error_when_check_is_violated(error_msg=msg)
+
+        if not self._check_external_file_data():
+            msg = (
+                "%s '%s': Either external_file or data must be specified (not None), but not both."
+                % (self.__class__.__name__, name)
+            )
+            self._raise_error_when_check_is_violated(error_msg=msg)
+
```
oruebel: This pattern feels a bit verbose, in that it requires a lot of if statements with checks, generation of messages and calls to raise. To make ``__init__`` more compact and to have the checks more self-contained, I think one option would be to:

1. Have the ``_check`` functions return the message string if an error should be raised and return None if the check is passed
2. In `` self._raise_error_when_check_is_violated(error_msg=msg)`` we can then check if ``error_msg is None ``and simply return if it is ``None`` (I added a corresponding suggestion below)

With these changes the code above would simplify to:

```
self._raise_error_when_check_is_violated(error_msg=self._check_external_file_starting_frame_length())
self._raise_error_when_check_is_violated(error_msg=self._check_external_file_format())
self._raise_error_when_check_is_violated(error_msg=self._check_external_file_data())
```

Or one could in this way even write this a loop, but I think that only makes sense when there are a very large number of checks.

```
data_checks = [self._check_external_file_starting_frame_length, self._check_external_file_format, self._check_external_file_data]
for check in data_checks:
    self._raise_error_when_check_is_violated(error_msg=check)
```

src/pynwb/image.py
oruebel: It would be nice to have also a test for Line 136 to check that warning are raised on read. Creating invalid files to test all these options may be too involved just to test this behavior, but I think we could just copy the pattern from the ObjectMapper itself in the test:

https://github.com/hdmf-dev/hdmf/blob/7be40229d778dea940134753de957f2457a11e1f/src/hdmf/build/objectmapper.py#L1257-L1262

I.e., call ``__new__`` explicitly to set ``in_construct_mode=True`` and then call ``__init__`` with the invalid data to check if the warnings are raised as expected. I.e., a test would look something like:

```
def test_warn_on_bad_starting_frame():
     obj = ImageSeries.__new__(ImageSeries, container_source=None, parent=None, object_id="test", in_construct_mode=True)
      with self.assertWarnsWith(ValueError,  "...message..."):
           obj.__init__(........arguments for ImageSeries.....)
     obj._in_construct_mode = False 
```

Setting ``obj._in_construct_mode = False`` should only be necessary of you continue to the use the object.

oruebel: Ok, looking at the tests, it looks like you are already testing this with actual files. Maybe the backward compatibility tests are not counted in codecov? @rly

@weiglszonja so I don't think we need to add more tests, but it would still be interesting to know why codecov indicates that Line 136 ids not covered when it seems like it should be based on the tests you added.

oruebel:
@weiglszonja overall this PR looks good. Normally, I would have just approved this PR as is, however, since this is the first example where we distinguish between warning on read and error on new, I think it will be useful to spent the extra effort to have the pattern we use for this in the code as clean as possible as others will likely follow the pattern we set here in the future. I think the suggestions I made should only require some minor reshuffling of lines of code, so hopefully the extra effort should not be too much. Thanks a lot for your patience and help!

weiglszonja:
@oruebel Thank you for the suggestions, I completely agree with your comments and I hope now this PR is in the right shape. This [codecov](https://app.codecov.io/gh/NeurodataWithoutBorders/pynwb/compare/1516/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr%20comments&utm_term=NeurodataWithoutBorders#diff-c3JjL3B5bndiL2NvcmUucHk=) is still bugging me, maybe adding another test wouldn't hurt?

src/pynwb/image.py
rly: Backward compatibility tests are currently not counted in codecov, but now that we are including integration tests in codecov, I think it makes sense to include the backward compatibility tests too.

oruebel:
> This [codecov](https://app.codecov.io/gh/NeurodataWithoutBorders/pynwb/compare/1516/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr%20comments&utm_term=NeurodataWithoutBorders#diff-c3JjL3B5bndiL2NvcmUucHk=) is still bugging me, maybe adding another test wouldn't hurt?

I just added variants of the three relevant unit test cases that you had added to run the same test also in construct mode. I think that should take care of the codecov. codecov is I believe running the different tests suits (unit, integration, back_combat etc.) separately and my guess is that for some reason it is only considering the unit tests here. Adding tests in the unit test suit should, therefore, hopefully fix this.

oruebel (dismissed):

oruebel (approved):

src/pynwb/image.py
oruebel: Thanks @rly for the clarification. That makes sense. I added a couple of extra unit-tests anyways to cover this case there as well. I think those tests are useful also because they show how we can write unittests that mimic the use-case were we need to distinguish between construct mode on read and regular user mode.

Issue #1449, opened 2022-04-07
[Bug]: EpochTags


TomDonoghue:
### What happened?

In the `NWBFile.add_epoch` method, the description notes that the `tags` input can be str, list or tuple. 

Tags description:

tags (:py:class:`~str` or :py:class:`~list` or :py:class:`~tuple`): user-defined tags used throughout time intervals

However, if one actually passes a str, then the `epoch_tags` attribute doesn't preserve this as a string, but splits it up into individual elements (characters). In this situation, it makes the `epoch_tags` field not very useful. As far as I can tell, this is simply an issue with however `epoch_tags` samples the tags from the `epochs` data, since examining the tags within the epochs structure all looks correct. 

Options:
- I presume however `epoch_tags` is sampling from epochs is doing some kind of type conversion (calling `set` perhaps, though I'm quite sure where) which is going wrong, and this could be updated to better handle strings.
- Alternately, perhaps the tags input is not really supposed to accept string inputs, in which case the description should be updated, and potentially a typecheck warning or error could be raised. 

### Steps to Reproduce

```python
from datetime import datetime
from dateutil.tz import tzlocal
from pynwb import NWBFile

nwbfile = NWBFile(session_description='test',
                  identifier='000',
                  session_start_time=datetime.now(tzlocal()))

nwbfile.add_epoch(0., 10., tags='label1')

# This stores the tags properly in epochs
print(nwbfile.epochs.tags[:])

# However, it gets stored wrong in `epoch_tags`
print(nwbfile.epoch_tags)
# The printed output I see from this is:
# {'e', 'l', 'a', '1', 'b'}
```

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.8

### Package Versions

I'm using pynwb version 2.0.1

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/.github/CODE_OF_CONDUCT.rst)
- [X] Have you checked the [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [X] Have you ensured this bug was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues)?

bendichter:
[...]

PR #1521, opened 2022-07-26
[BUG] - Fix update of epoch_tags store

TomDonoghue:
## Motivation

This addresses #1449 

The way that `epoch_tags` was updated before didn't play well with strings, which should be a supported input. Since `epoch_tags` is a set, adding a string gets treated as an iterable, which then separates and adds each character. 

This PR fixes this section to check the types of the tags, and deal with them if they are a string. 

Note that the string processing line is a direct copy of `pynwb.epoch` line 47, to make sure strings are processed the same, and that epoch_tags properly supports the possible input of something like `'label1, label2'`. 

## How to test the behavior?

There is example code that shows the problem in #1449. I check that this same code when tested against this PR works as it should.

## Checklist

### DIFF
--- a/src/pynwb/file.py
+++ b/src/pynwb/file.py
@@ -590,6 +590,7 @@ class NWBFile(MultiContainerInterface):
         self.__check_epochs()
         if kwargs['tags'] is not None:
             # If a str is passed into epoch_tags directly, it gets split into characters
+            #   This processing needs to match tags parsing in `epoch.TimeIntervals.add_interval`
             tmp = kwargs['tags']
             if isinstance(kwargs['tags'], str):
                 tmp = [s.strip() for s in kwargs['tags'].split(",") if not s.isspace()]
--- a/src/pynwb/file.py
+++ b/src/pynwb/file.py
@@ -589,7 +589,11 @@ class NWBFile(MultiContainerInterface):
         """
         self.__check_epochs()
         if kwargs['tags'] is not None:
-            self.epoch_tags.update(kwargs['tags'])
+            # If a str is passed into epoch_tags directly, it gets split into characters
+            tmp = kwargs['tags']
+            if isinstance(kwargs['tags'], str):
+                tmp = [s.strip() for s in kwargs['tags'].split(",") if not s.isspace()]
+            self.epoch_tags.update(tmp)
         self.epochs.add_interval(**kwargs)
 
     def __check_electrodes(self):
--- a/src/pynwb/epoch.py
+++ b/src/pynwb/epoch.py
@@ -33,8 +33,10 @@ class TimeIntervals(DynamicTable):
 
     @docval({'name': 'start_time', 'type': 'float', 'doc': 'Start time of epoch, in seconds'},
             {'name': 'stop_time', 'type': 'float', 'doc': 'Stop time of epoch, in seconds'},
-            {'name': 'tags', 'type': (str, list, tuple), 'doc': 'user-defined tags used throughout time intervals',
-             'default': None},
+            {'name': 'tags', 'type': (str, list, tuple),
+             'doc': ('User-defined tags used throughout time intervals. If list or tuple each element is treated as '
+                     'a separate tag. If str, can be a single tag, or comma-separated entries that will be parsed '
+                     'into individual tags.'), 'default': None},
             {'name': 'timeseries', 'type': (list, tuple, TimeSeries), 'doc': 'the TimeSeries this epoch applies to',
              'default': None},
             allow_extra=True)


src/pynwb/file.py
```diff
@@ -589,7 +589,11 @@ def add_epoch(self, **kwargs):
         """
         self.__check_epochs()
         if kwargs['tags'] is not None:
-            self.epoch_tags.update(kwargs['tags'])
+            # If a str is passed into epoch_tags directly, it gets split into characters
+            tmp = kwargs['tags']
+            if isinstance(kwargs['tags'], str):
+                tmp = [s.strip() for s in kwargs['tags'].split(",") if not s.isspace()]
```
oruebel, line 598: If I understand this correctly, then if `tags='tag1, tag2, tag3'` this would split the string to make a list `['tag1'. 'tag2', 'tag3']`. It's not clear to me what the use case for allowing implicit definitions of lists via strings is. I agree that the current behavior of splitting a string into characters is not useful, but I think that is largely a bug because the string is treated like a list. Instead, I would suggest treating the data types for what they are. That is, if the user passes a string we should treat it as a single tag, and if the user passes a list we should treat it as a list of tags. Long-story-short, I would suggest changing the code as follows: 


```suggestion
            if isinstance(kwargs['tags'], str):
                tmp = [kwargs['tags'], ]  
```

TomDonoghue: This makes sense to me in terms of a clear / consistent behaviour, but I think would require a slightly broader change to handling tags in `pynwb.epoch` as well as the change here. 

What I currently wrote is based on making sure `epoch_tags`, which I understand to keep a list of all observed tags is consistent with the tags that get added to the epochs table. To keep this consistent, the strings should be processed in the same way in `pynwb.epoch` line 47. This would overall update the behavior of string tags for anyone currently doing this (the bug here is only about `epoch_tags`, and if anyone was currently using a `tag1, tag2` strategy, these tags are currently split into a list and added to the epochs table).

oruebel: That makes sense. Thanks for the pointer to https://github.com/NeurodataWithoutBorders/pynwb/blob/ed9eb6d16e5031cb1d19f74272364394f1ab3188/src/pynwb/epoch.py#L46-L47 I was not aware of this behavior. In this case, what you have makes perfect sense. 

I would suggest to updated the doc string for tags here:

https://github.com/NeurodataWithoutBorders/pynwb/blob/ed9eb6d16e5031cb1d19f74272364394f1ab3188/src/pynwb/epoch.py#L36-L37

to describe how strings are being treated. I.e., describe that if a string is passed in that the string is expected to be a comma-separated list and will be parsed accordingly. Currently, I don't think this behavior is clear from the docs. Could you add this change on this PR. 

We could consider consider changing how strings are being treated here (i.e., either always require lists or treat strings as a single tag) but I would recommend to do this as a separate issue. For this PR, making sure that the behavior is consistent seems the right thing to do. Thanks for your help!

TomDonoghue: This sounds good / makes sense to me! I have updated the docstring in `pynwb.src.pynwb.epoch` to describe the behaviour for how it treats strings. I also added a note in `file` that processing should match what is done in `epoch`.

I think that should address what's needed for this PR, to fix the bug and update the docs - and then if the behaviour is to be changed at some point, that could be a separate PR.

oruebel (dismissed):

oruebel:
Thanks for the bug fix!

oruebel:
Merging of this PR is currently blocked because flake8 checks are failing in file not modified by this PR. This appears to be due to tests running a newer version of flake8. I've created #1524 to fix those issues. Once that PR has been merged we can sync this PR with dev and merge.

rly:
Looking at `epoch_tags` with a closer eye, I do not fully understand why this instance variable exists and whether it works as intended. As far as I can tell, `epoch_tags` is a cache of the unique tag values in the NWBFile.epochs table. That may be convenient but
1. it can become out of sync, especially if a user calls `nwbfile.epochs.add_interval(...)` instead of `nwbfile.add_epoch(...)`, and 
2. it is not populated when loading data from a file -- so it is only cached on write. 

Separately, the `update` method is called on `epoch_tags` which works only if `epoch_tags` is a set, but the docval for `epoch_tags` allows the variable to be a tuple or a list.

@TomDonoghue are you using this field, and if so, how are you using it?

TomDonoghue:
@rly - I am not actively using this field - I stumbled onto this bug a while ago when I was exploring different ways to store event information (I ended up not actively using epochs). As far as I understand how it's working (having not gone too deep into this), I agree with your notes here that `epoch_tags` is a potentially useful idea, but that it's not guaranteed to be consistent or populated. 

In terms of type, I'm not clear on if / how `epoch_tags` could end up anything other than a set in practice and calling `update` is done in the current version of the code in both `add_epoch` and `add_epoch_column`, so it seems to me that for this the code expects a set and the documentation is perhaps wrong?

rly:
OK. Good to know, thanks. 

In `NWBFile.__init__`: 
https://github.com/NeurodataWithoutBorders/pynwb/blob/d75cc825fe4aaeae18feeac3cbcb3f70651a4c09/src/pynwb/file.py#L317-L318

`epoch_tags` is allowed to be a tuple or list, but as you mention, the code expects a set, so at the very least we need to update this to allow only sets and rewrite the docstring.

However, given its limited usefulness and potential confusion, I am inclined to raise a warning when it is used to say that 1) this variable is not stored on disk, 2) it is not populated on read, 3) it may be incorrect if epochs are added not through methods on `NWBFile`, and 4) that it will be deprecated in a future version of PyNWB. @oruebel @bendichter what do you think?

TomDonoghue:
@rly if `epoch_tags` is supposed to be a shortcut for getting all unique tags that are present, it sorta feels like this might be better accomplished by a property, something like:

```
@property
def epoch_tags(self):
    return set(self.epochs.tags[:])
```

I think the above would address both of the concerns with the current approach?
It does read `tags` every time it's called, but I'm assuming that the list of epochs / tags is unlikely to get so big that this becomes slow / an issue.

TomDonoghue:
@rly - just wanted to follow up here - is this still relevant / useful, or should I close this PR as out of date?

stephprince:
I believe this was addressed with this PR: https://github.com/NeurodataWithoutBorders/pynwb/pull/1935. Please reopen if not.

PR #1043, opened 2019-08-19
Replace float in docval with 'float' to prevent py3.5 on windows issue

rly:
helps deal with #1015

PR #1523, opened 2022-07-30
[DOC] - Update docval doctypes for those that render weird on docsite

TomDonoghue:
## Motivation

On the docsite, when type annotations are provided as a string, they render weird on the docsite, for example:
Screen Shot 2022-07-30 at 10 49 10 AM

This is usually the case for `float` - searching for why this is points back to this PR:
https://github.com/NeurodataWithoutBorders/pynwb/pull/1043

^The original use of 'float' was to address an issue on Py3.5. Since pynwb is now tagged at python >= 3.7, this workaround is no longer needed, so it seems these types can revert back to the normal way, which should fix some of the rendering on the docsite. 

The first commit is explicitly updating 'float', the second updates some miscellaneous other type listing in the same way. 

## How to test the behavior?

There are no real code changes - as long as the tests pass and the generated docsite look fine, this should be good I think.

## Checklist

### DIFF
--- a/src/pynwb/__init__.py
+++ b/src/pynwb/__init__.py
@@ -247,7 +247,7 @@ class NWBHDF5IO(_HDF5IO):
 
     @docval({'name': 'src_io', 'type': HDMFIO,
              'doc': 'the HDMFIO object (such as NWBHDF5IO) that was used to read the data to export'},
-            {'name': 'nwbfile', 'type': 'NWBFile',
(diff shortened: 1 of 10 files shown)


bendichter (approved):

rly (approved):

rly:
LGTM. Thanks @TomDonoghue !

PR #1526, opened 2022-08-04
update streaming tutorial

### LINKED ISSUE #1525: [Feature]: support fsspec / opening from a file object

ROS3 driver provides a nice way to stream data without downloading file in full, [as shown in this example](https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/streaming.html).  But it is S3 specific.

somewhat. AFAIK ros3 driver doesn't provide ability for local sparse caching etc.

 Although i have not looked into how it is implemented, I bet ros3 driver is *working* through HTTP Range requests.  Such method is generic and could work for any http/https URLs.  That relies in the basis of many solutions, such as https://github.com/fsspec/filesystem_spec providing uniform access to a wide (not just S3) range of resources, where `fsspec` provides generic `file` interface, see [e.g. examples in fsspec doc](https://filesystem-spec.readthedocs.io/en/latest/usage.html).  Another example for such use could be https://github.com/datalad/datalad-fuse/ where we provide fsspec-based FUSE file system for any datalad dataset where fsspec provides seamless, locally cached, sparse access to any remote file, nwb included.

So, by supporting opening from a file handle, it should be possible to open some fsspec'ed file. `NWBHDF5IO` seems to take path, not file handle.  Would it be "easy" to support operating with a pre-opened file handle?  If so -- might be just a matter of an example documentation to demonstrate such a cool feature.

No.

### Code of Conduct

bendichter:
* change fsspec from S3 to http protocol
* intersphinx into dandi and fsspec

fix #1525

### DIFF
--- a/docs/gallery/advanced_io/streaming.py
+++ b/docs/gallery/advanced_io/streaming.py
@@ -66,7 +66,7 @@ Streaming Method 2: fsspec
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 fsspec is another data streaming approach that is quite flexible and has several performance advantages. This library
 creates a virtual filesystem for remote stores. With this approach, a virtual file is created for the file and
-virtual filesystem layer will take care of requesting data from the s3 bucket whenever data is
+the virtual filesystem layer takes care of requesting data from the s3 bucket whenever data is
 read from the virtual file.
 
 First install ``fsspec`` and the dependencies of the :py:class:`~fsspec.implementations.http.HTTPFileSystem`:
--- a/docs/gallery/advanced_io/streaming.py
+++ b/docs/gallery/advanced_io/streaming.py
@@ -4,7 +4,7 @@
 Streaming NWB files
 ===================
 
-YOu can read specific sections within individual data files directly from remote storessuch as the
+You can read specific sections within individual data files directly from remote stores such as the
 `DANDI Archive <https://dandiarchive.org/>`_. This does not require you to download the entire large file just to
 access a small amount of datais especially useful for reading small pieces of data from a large NWB file stored
 remotely. First, you will need to get the location of the file. The code below illustrates how to do this on DANDI
--- a/docs/gallery/advanced_io/streaming.py
+++ b/docs/gallery/advanced_io/streaming.py
@@ -67,7 +67,8 @@ Streaming Method 2: fsspec
 fsspec is another data streaming approach that is quite flexible and has several performance advantages. This library
 creates a virtual filesystem for remote stores. With this approach, a virtual file is created for the file and
 the virtual filesystem layer takes care of requesting data from the S3 bucket whenever data is
-read from the virtual file.
+read from the virtual file.  Note that this implementation is completely unaware of internals of the HDF5 format
+and thus can work for **any** file, not only for the purpose of use with H5PY and PyNWB.
 
 First install ``fsspec`` and the dependencies of the :py:class:`~fsspec.implementations.http.HTTPFileSystem`:
 
--- a/docs/gallery/advanced_io/streaming.py
+++ b/docs/gallery/advanced_io/streaming.py
@@ -10,8 +10,8 @@ access a small amount of datais especially useful for reading small pieces of da
 remotely. First, you will need to get the location of the file. The code below illustrates how to do this on DANDI
 using the dandi API library.
 
-Getting file location on DANDI
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Getting the location of the file on DANDI
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The :py:class:`~dandi.dandiapi.DandiAPIClient` can be used to get the S3 URL of any NWB file stored in the DANDI
 Archive. If you have not already, install the latest release of the ``dandi`` package.
--- a/docs/gallery/advanced_io/streaming.py
+++ b/docs/gallery/advanced_io/streaming.py
@@ -99,7 +99,7 @@ Then in Python:
                 print(nwbfile.acquisition['lick_times'].time_series['lick_left_times'].data[:])
 
 
-fsspec is a protocol that can be used to access a variety of different store formats, including (at the time of
+fsspec is a library that can be used to access a variety of different store formats, including (at the time of
 writing):
 
 .. code-block:: python
--- a/docs/gallery/advanced_io/streaming.py
+++ b/docs/gallery/advanced_io/streaming.py
@@ -88,13 +88,7 @@ Then in Python:
     # caching to save accessed data to RAM.
     fs = CachingFileSystem(
         fs=fsspec.filesystem("http"),
-        # target_protocol='blockcache',
-        cache_storage="nwb-cache",
-        # cache_check=600,
-        # block_size=1024,
-        # check_files=True,
-        # expiry_times=True,
-        # same_names=True,
+        cache_storage="nwb-cache",  # Local folder for the cache
     )
 
     # next, open the file
--- a/docs/gallery/advanced_io/streaming.py
+++ b/docs/gallery/advanced_io/streaming.py
@@ -5,8 +5,8 @@ Streaming NWB files
 ===================
 
 You can read specific sections within individual data files directly from remote stores such as the
-`DANDI Archive <https://dandiarchive.org/>`_. This does not require you to download the entire large file just to
-access a small amount of datais especially useful for reading small pieces of data from a large NWB file stored
+`DANDI Archive <https://dandiarchive.org/>`_. This is especially useful for reading small pieces of data 
+from a large NWB file stored
 remotely. First, you will need to get the location of the file. The code below illustrates how to do this on DANDI
 using the dandi API library.
 
--- a/docs/gallery/advanced_io/streaming.py
+++ b/docs/gallery/advanced_io/streaming.py
@@ -1,98 +1,119 @@
 '''
 .. _streaming:
 
-Streaming from an S3 Bucket
-===========================
+Streaming NWB files
+===================
 
-It is possible to read data directly from an S3 bucket, such as data from the `DANDI Archive
-<https://dandiarchive.org/>`_. This is especially useful for reading small pieces of data
-from a large NWB file stored remotely. In fact, there are two different ways to do this supported by PyNWB.
+YOu can read specific sections within individual data files directly from remote storessuch as the
+`DANDI Archive <https://dandiarchive.org/>`_. This does not require you to download the entire large file just to
+access a small amount of datais especially useful for reading small pieces of data from a large NWB file stored
+remotely. First, you will need to get the location of the file. The code below illustrates how to do this on DANDI
+using the dandi API library.
 
-Method 1: ROS3
-~~~~~~~~~~~~~~
-ROS3 stands for "read only S3" and is a driver created by the HDF5 group that allows HDF5 to read HDF5 files
-stored on s3. Using this method requires that your HDF5 library is installed with the ROS3 driver enabled. This
-is not the default configuration, so you will need to make sure you install the right version of h5py that has this
-advanced configuration enabled. You can install HDF5 with the ROS3 driver from `conda-forge
-<https://conda-forge.org/>`_ using ``conda``. You may first need to uninstall a currently installed version of h5py.
+Getting file location on DANDI
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+The :py:class:`~dandi.dandiapi.DandiAPIClient` can be used to get the S3 URL of any NWB file stored in the DANDI
+Archive. If you have not already, install the latest release of the ``dandi`` package.
+
+
+.. code-block:: bash
+
+   pip install dandi
+
+Now you can get the url of a particular NWB file using the dandiset ID and the path of that file within the dandiset.
+
+.. code-block:: python
+
+   from dandi.dandiapi import DandiAPIClient
+
+   dandiset_id = '000006'  # ephys dataset from the Svoboda Lab
+   filepath = 'sub-anm372795/sub-anm372795_ses-20170718.nwb'  # 450 kB file
+   with DandiAPIClient() as client:
+       asset = client.get_dandiset(dandiset_id, 'draft').get_asset_by_path(filepath)
+       s3_url = asset.get_content_url(follow_redirects=1, strip_query=True)
+
+
+Streaming Method 1: ROS3
+~~~~~~~~~~~~~~~~~~~~~~~~
+ROS3 is one of the supported methods for reading data from a remote store. ROS3 stands for "read only S3" and is a
+driver created by the HDF5 Group that allows HDF5 to read HDF5 files stored remotely in s3 buckets. Using this method
+requires that your HDF5 library is installed with the ROS3 driver enabled. This is not the default configuration,
+so you will need to make sure you install the right version of ``h5py`` that has this advanced configuration enabled.
+You can install HDF5 with the ROS3 driver from `conda-forge <https://conda-forge.org/>`_ using ``conda``. You may
+first need to uninstall a currently installed version of ``h5py``.
+
+.. code-block:: bash
+
+   pip uninstall h5py
+   conda install -c conda-forge "h5py>=3.2"
+
+Now instantiate a :py:class:`~pynwb.NWBHDF5IO` object with the S3 URL and specify the driver as "ros3". This
+will download metadata about the file from the S3 bucket to memory. The values of datasets are accessed lazily,
+just like when reading an NWB file stored locally. So, slicing into a dataset will require additional time to
+download the sliced data (and only the sliced data) to memory.
+
+.. code-block:: python
+
+   from pynwb import NWBHDF5IO
+
+   with NWBHDF5IO(s3_url, mode='r', load_namespaces=True, driver='ros3') as io:
+       nwbfile = io.read()
+       print(nwbfile)
+       print(nwbfile.acquisition['lick_times'].time_series['lick_left_times'].data[:])
+
+Streaming Method 2: fsspec
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+fsspec is another data streaming approach that quite flexible and has several performance advantages. This library
+creates a virtual filesystem for remote stores. With this approach, a virtual file is created for the file and
+virtual filesystem layer will take care of requesting data from the s3 bucket whenever data is
+read from the virtual file.
+
+First install ``fsspec`` and the dependencies of the :py:class:`~fsspec.implementations.http.HTTPFileSystem`:
+
+.. code-block:: bash
+
+   pip install fsspec requests aiohttp
+
+Then in Python:
+
+.. code-block:: python
+
+    import fsspec
+    import pynwb
+    import h5py
+    from fsspec.implementations.cached import CachingFileSystem
+
+    # first, create a virtual filesystem based on the http protocol and use
+    # caching to save accessed data to RAM.
+    fs = CachingFileSystem(
+        fs=fsspec.filesystem("http"),
+        # target_protocol='blockcache',
+        cache_storage="nwb-cache",
+        # cache_check=600,
+        # block_size=1024,
+        # check_files=True,
+        # expiry_times=True,
+        # same_names=True,
+    )
+
+    # next, open the file
+    with fs.open(s3_url, "rb") as f:
+        with h5py.File(f) as file:
+            with pynwb.NWBHDF5IO(file=file, load_namespaces=True) as io:
+                nwbfile = io.read()
+                print(nwbfile.acquisition['lick_times'].time_series['lick_left_times'].data[:])
+
+
+fsspec is a protocol that can be used to access a variety of different store formats, including (at the time of
+writing):
+
+.. code-block:: python
+
+    from fsspec.registry import known_implementations
+    known_implementations.keys()
+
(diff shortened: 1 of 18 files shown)


bendichter:
@yarikoptic can you please take a look at

https://pynwb--1526.org.readthedocs.build/en/1526/tutorials/advanced_io/streaming.html#sphx-glr-tutorials-advanced-io-streaming-py

and lmk what you think?

docs/gallery/advanced_io/streaming.py
```diff
@@ -1,98 +1,119 @@
...
-# and s3fs has internal mechanisms to retry these requests automatically, whereas ROS3 does not. However, it may not
-# be available on all platforms. s3fs does not currently work for Windows.
+   with NWBHDF5IO(s3_url, mode='r', load_namespaces=True, driver='ros3') as io:
+       nwbfile = io.read()
+       print(nwbfile)
+       print(nwbfile.acquisition['lick_times'].time_series['lick_left_times'].data[:])
 
+Streaming Method 2: fsspec
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+fsspec is another data streaming approach that quite flexible and has several performance advantages. This library
```
oruebel:
```suggestion
fsspec is another data streaming approach that is quite flexible. This library
```

docs/gallery/advanced_io/streaming.py
```diff
@@ -1,98 +1,119 @@
...
+creates a virtual filesystem for remote stores. With this approach, a virtual file is created for the file and
+virtual filesystem layer will take care of requesting data from the s3 bucket whenever data is
```
oruebel:
```suggestion
the virtual filesystem layer takes care of requesting data from the s3 bucket whenever data is
```

docs/gallery/advanced_io/streaming.py
```diff
@@ -1,98 +1,119 @@
...
-#   import pynwb
-#   import h5py
-#
-#   fs = s3fs.S3FileSystem(anon=True)
-#
-#   f = fs.open("s3://dandiarchive/blobs/43b/f3a/43bf3a81-4a0b-433f-b471-1f10303f9d35", 'rb')
-#   file = h5py.File(f)
-#   io = pynwb.NWBHDF5IO(file=file, load_namespaces=True)
-#
-#   io.read()
```
oruebel, line 88: Since the ``s3fs`` builds on ``fsspec`` and since the ``s3fs`` code is simpler than using the full ``fsspec`` I think it would be nice to add a subsection to the ``Streaming Method 2: fsspec`` section that shows this code example for ``s3fs`` as well. In this way you can still keep the tutorial to the two main options but still mention the``s3fs`` option, which may be sufficient  and simpler for many users.

docs/gallery/advanced_io/streaming.py
oruebel: Can you also add a link to the ffspec documentation here.

oruebel: I found the phase ``and has several performance advantages.`` a bit confusing, mainly because it is non-specific. Instead, I think it would be useful if we could include a more specific section on the advantages/disadvantages of the two options and which option to use when.

oruebel:
Thanks @bendichter for adding the ``fsspec`` option and thanks @yarikoptic for investing and suggesting this option

docs/gallery/advanced_io/streaming.py
bendichter: So s3fs is a specific implementation of the fsspec protocol. I changed over to using the more generic fsspec library with http for three reasons:
1. This way we can use the http path as input, which is the same path that the ros3 driver takes, so it is clear how the two methods relate to each other.
2. Showing it this way makes it pretty clear how you would extend this to other backends.
3. This method also shows how to use caching, which I think is an important feature that a lot of users will want to leverage.

Do you really think this workflow is more complex than the s3fs syntax?

docs/gallery/advanced_io/streaming.py
```diff
@@ -1,98 +1,119 @@
...
 
-Streaming from an S3 Bucket
-===========================
+Streaming NWB files
+===================
 
-It is possible to read data directly from an S3 bucket, such as data from the `DANDI Archive
-<https://dandiarchive.org/>`_. This is especially useful for reading small pieces of data
-from a large NWB file stored remotely. In fact, there are two different ways to do this supported by PyNWB.
+YOu can read specific sections within individual data files directly from remote storessuch as the
```
bendichter, line 7:
```suggestion
You can read specific sections within individual data files directly from remote storessuch as the
```

bendichter:
```suggestion
You can read specific sections within individual data files directly from remote stores such as the
```

docs/gallery/advanced_io/streaming.py
oruebel: Showing the ``fsspec`` example as the main example makes sense. If you think that showing the full ``s3fs`` example in addition is too confusing, then maybe we can just add a note along the lines of "The s3fs package also builds on fsspec and can be used to access S3 in a similar manner. s3fs provides additional convenience methods for interacting with S3 more specifically while fsspec provides greater flexibility with regard to supported remote file systems. Note, when using s3fs requires the use of ``s3`` URIs instead of ``http``.  "

docs/gallery/advanced_io/streaming.py
```diff
@@ -1,98 +1,119 @@
...
+   filepath = 'sub-anm372795/sub-anm372795_ses-20170718.nwb'  # 450 kB file
+   with DandiAPIClient() as client:
+       asset = client.get_dandiset(dandiset_id, 'draft').get_asset_by_path(filepath)
+       s3_url = asset.get_content_url(follow_redirects=1, strip_query=True)
+
+
+Streaming Method 1: ROS3
+~~~~~~~~~~~~~~~~~~~~~~~~
+ROS3 is one of the supported methods for reading data from a remote store. ROS3 stands for "read only S3" and is a
+driver created by the HDF5 Group that allows HDF5 to read HDF5 files stored remotely in s3 buckets. Using this method
```
yarikoptic, line 40: HDF5 is just a format... most likely what was meant is libhdf5 library but here it doesn't really need it anyways IMHO, and consistent S3:

```suggestion
driver created by the HDF5 Group that allows to read HDF5 files stored remotely in S3 buckets. Using this method
```

docs/gallery/advanced_io/streaming.py
yarikoptic: I think the best would be if someone checks for what s3fs might be providing "on top" or in addition to pure straight fsspec. E.g. it could be some better choices of blocksize etc. I would have at least compared performance.

docs/gallery/advanced_io/streaming.py
```diff
@@ -1,98 +1,119 @@
...
+   with NWBHDF5IO(s3_url, mode='r', load_namespaces=True, driver='ros3') as io:
+       nwbfile = io.read()
+       print(nwbfile)
+       print(nwbfile.acquisition['lick_times'].time_series['lick_left_times'].data[:])
 
+Streaming Method 2: fsspec
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+fsspec is another data streaming approach that is quite flexible and has several performance advantages. This library
+creates a virtual filesystem for remote stores. With this approach, a virtual file is created for the file and
+the virtual filesystem layer takes care of requesting data from the s3 bucket whenever data is
```
yarikoptic:
```suggestion
the virtual filesystem layer takes care of requesting data from the S3 bucket whenever data is
```

docs/gallery/advanced_io/streaming.py
```diff
@@ -1,98 +1,119 @@
...
+read from the virtual file.
```
yarikoptic:
```suggestion
read from the virtual file.  Note that this implementation is completely unaware of internals of the HDF5 format
and thus can work for **any** file, not only for the purpose of use with H5PY and PyNWB.
```

docs/gallery/advanced_io/streaming.py
```diff
@@ -1,98 +1,119 @@
...
+    # caching to save accessed data to RAM.
+    fs = CachingFileSystem(
+        fs=fsspec.filesystem("http"),
+        # target_protocol='blockcache',
+        cache_storage="nwb-cache",
+        # cache_check=600,
+        # block_size=1024,
+        # check_files=True,
+        # expiry_times=True,
+        # same_names=True,
```
yarikoptic: I copy pasted those commented out options from older code in datalad-fuse. Not even sure if all of them applicable ATM, so best to just strip them away for now,

```suggestion
        cache_storage="nwb-cache",  # Local folder for the cache
```

docs/gallery/advanced_io/streaming.py
```diff
@@ -1,98 +1,119 @@
...
+
+    # next, open the file
+    with fs.open(s3_url, "rb") as f:
+        with h5py.File(f) as file:
+            with pynwb.NWBHDF5IO(file=file, load_namespaces=True) as io:
+                nwbfile = io.read()
+                print(nwbfile.acquisition['lick_times'].time_series['lick_left_times'].data[:])
+
+
+fsspec is a protocol that can be used to access a variety of different store formats, including (at the time of
```
yarikoptic:
```suggestion
fsspec is a library that can be used to access a variety of different store formats, including (at the time of
```
I think would better reflect its purpose here.

docs/gallery/general/file.py
```diff
@@ -190,7 +190,8 @@
 # The fields in the :py:class:`~pynwb.file.Subject` object are all free-form text (any format will be valid),
 # however it is recommended to follow particular conventions to help software tools interpret the data:
 #
-# * **age**: `ISO 8601 Duration format <https://en.wikipedia.org/wiki/ISO_8601#Durations>`_, e.g., ``"P90D"`` for 90 days old
+# * **age**: `ISO 8601 Duration format <https://en.wikipedia.org/wiki/ISO_8601#Durations>`_, e.g., ``"P90D"`` for 90
+# days old
```
yarikoptic: changes in this file are unrelated and thus best to be excluded from this PR. and not sure if that would not break this ReST itemized list in such form of wrapping

yarikoptic (commented):
Some initial comments / recommendations

docs/gallery/general/file.py
bendichter: thanks for pointing this out. Not sure how these changes snuck in here..

bendichter:
```suggestion
# * **age**: `ISO 8601 Duration format <https://en.wikipedia.org/wiki/ISO_8601#Durations>`_, e.g., ``"P90D"`` for 90 days old
```

docs/gallery/general/file.py
```diff
@@ -247,7 +248,8 @@
 ####################
 # :py:class:`~pynwb.base.TimeSeries` objects can be added directly to :py:class:`~pynwb.file.NWBFile` using:
 #
-# * :py:meth:`~pynwb.file.NWBFile.add_acquisition` to  add *acquisition* data (raw, acquired data that should never change),
+# * :py:meth:`~pynwb.file.NWBFile.add_acquisition` to  add *acquisition* data (raw, acquired data that should never
+# change),
```
bendichter:
```suggestion
# * :py:meth:`~pynwb.file.NWBFile.add_acquisition` to  add *acquisition* data (raw, acquired data that should never change),
```

docs/gallery/general/file.py
```diff
@@ -306,8 +308,9 @@
 
 ####################
 # To help data analysis and visualization tools know that this :py:class:`~pynwb.behavior.SpatialSeries` object
-# represents the position of the subject, store the :py:class:`~pynwb.behavior.SpatialSeries` object inside
-# of a :py:class:`~pynwb.behavior.Position` object, which can hold one or more :py:class:`~pynwb.behavior.SpatialSeries` objects.
+# represents the position of the subject, store the :py:class:`~pynwb.behavior.SpatialSeries` object inside a
+# :py:class:`~pynwb.behavior.Position` object, which can hold one or more :py:class:`~pynwb.behavior.SpatialSeries`
+# objects.
```
bendichter:
```suggestion
# represents the position of the subject, store the :py:class:`~pynwb.behavior.SpatialSeries` object inside
# of a :py:class:`~pynwb.behavior.Position` object, which can hold one or more :py:class:`~pynwb.behavior.SpatialSeries` objects.
```

docs/gallery/domain/images.py
```diff
@@ -135,14 +134,8 @@
...
-# If the sampling rate is constant, use :py:attr:`~pynwb.base.TimeSeries.rate` and :py:attr:`~pynwb.base.TimeSeries.starting_time` to specify time.
-# For irregularly sampled recordings, use :py:attr:`~pynwb.base.TimeSeries.timestamps` to specify time for each sample image.
-#
-# Each external image may contain one or more consecutive frames of the full :py:class:`~pynwb.image.ImageSeries`.
-# The :py:attr:`~pynwb.image.ImageSeries.starting_frame` attribute serves as an index to indicate which frame
-# each file contains.
-# For example, if the ``external_file`` dataset has three paths to files and the first and the second file have 2 frames,
-# and the third file has 3 frames, then this attribute will have values `[0, 2, 4]`.
+# The :py:attr:`~pynwb.image.ImageSeries.starting_frame` attribute serves as an index to indicate the starting frame of
+# each external file, allowing you to skip the beginning of videos.
```
bendichter:
```suggestion
# If the sampling rate is constant, use :py:attr:`~pynwb.base.TimeSeries.rate` and :py:attr:`~pynwb.base.TimeSeries.starting_time` to specify time.
# For irregularly sampled recordings, use :py:attr:`~pynwb.base.TimeSeries.timestamps` to specify time for each sample image.
#
# Each external image may contain one or more consecutive frames of the full :py:class:`~pynwb.image.ImageSeries`.
# The :py:attr:`~pynwb.image.ImageSeries.starting_frame` attribute serves as an index to indicate which frame
# each file contains.
# For example, if the ``external_file`` dataset has three paths to files and the first and the second file have 2 frames,
# and the third file has 3 frames, then this attribute will have values `[0, 2, 4]`.
```

docs/gallery/domain/images.py
```diff
@@ -119,7 +119,6 @@
     data=image_data,
     description="Image data of an animal moving in environment.",
     unit="n.a.",
-    format="raw",
     rate=1.0,
```
bendichter:
```suggestion
    format="raw",
    rate=1.0,
```

docs/gallery/domain/images.py
```diff
@@ -92,10 +95,7 @@
     orientation="lower left",  # required
     data=image_data,
     unit="n.a.",
-    format="raw",
-    starting_frame=[0.0],
     rate=1.0,
```
bendichter:
```suggestion
    format="raw",
    starting_frame=[0.0],
    rate=1.0,
    comments="no comments",
```

docs/gallery/domain/images.py
```diff
@@ -83,6 +83,9 @@
 # We can add it to the :py:class:`~pynwb.file.NWBFile` object as stimulus data using
 # the :py:meth:`~pynwb.file.NWBFile.add_stimulus` method.
 #
+# If the sampling rate is constant, use :py:attr:`~pynwb.base.TimeSeries.rate` and :py:attr:`~pynwb.base.TimeSeries.starting_time` to specify time.
+# For irregularly sampled recordings, use :py:attr:`~pynwb.base.TimeSeries.timestamps` to specify time for each sample image.
+#
```
bendichter:
```suggestion
```

docs/gallery/advanced_io/streaming.py
bendichter: is it true that s3fs provides additional convenience methods for interacting with S3 more specifically?

s3fs implements in [`S3FileSystem`](https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem) which inherits from fsspec's `AsyncFileSystem` abstract class. My impression is that it just implements the method of a `fsspec.FileSystem` without much else besides maybe credential controls that we don't use. You can switch to an s3 backend by simply changing `fs=fsspec.filesystem("http")` to `fs=fsspec.filesystem("s3")`, but then you'll have to supply the s3 path, not the http path.

oruebel: It's hard to tell from the documentation what the specific differences are, but my impression is that s3fs adds support for ``async``, improved logging, and S3 specific implementation of the file system functions. Purely to read an NWB file, this may not be big difference. Pending further investigation, maybe what we could do is to add a note to indicate that there are a growing number of libraries build on ffspec that provide functionality specific for different file systems (e.g., S3, Google Cloud, or Azure) and link to https://filesystem-spec.readthedocs.io/en/latest/api.html?highlight=S3#other-known-implementations. We could then post alternate examples, such as the code using S3FS, as GitHub Gists online and link to them from the same note. This would keep the docs clean and at the same time provide examples for folks that may want to use the more specific libraries build on fsspec. Given that we don't know how all the various options perform, we could also add a "disclaimer" along those lines.

docs/gallery/advanced_io/streaming.py
```diff
@@ -1,98 +1,114 @@
...
+Streaming NWB files
+===================
 
-It is possible to read data directly from an S3 bucket, such as data from the `DANDI Archive
-<https://dandiarchive.org/>`_. This is especially useful for reading small pieces of data
-from a large NWB file stored remotely. In fact, there are two different ways to do this supported by PyNWB.
+You can read specific sections within individual data files directly from remote stores such as the
+`DANDI Archive <https://dandiarchive.org/>`_. This does not require you to download the entire large file just to
+access a small amount of datais especially useful for reading small pieces of data from a large NWB file stored
+remotely. First, you will need to get the location of the file. The code below illustrates how to do this on DANDI
```
rly:
```suggestion
`DANDI Archive <https://dandiarchive.org/>`_. This is especially useful for reading small pieces of data 
from a large NWB file stored
remotely. First, you will need to get the location of the file. The code below illustrates how to do this on DANDI
```

docs/gallery/advanced_io/streaming.py
```diff
@@ -1,98 +1,114 @@
...
 
-Method 1: ROS3
-~~~~~~~~~~~~~~
-ROS3 stands for "read only S3" and is a driver created by the HDF5 group that allows HDF5 to read HDF5 files
-stored on s3. Using this method requires that your HDF5 library is installed with the ROS3 driver enabled. This
-is not the default configuration, so you will need to make sure you install the right version of h5py that has this
-advanced configuration enabled. You can install HDF5 with the ROS3 driver from `conda-forge
-<https://conda-forge.org/>`_ using ``conda``. You may first need to uninstall a currently installed version of h5py.
+Getting file location on DANDI
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
rly:
```suggestion
Getting the location of the file on DANDI
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

rly:
This looks good to me. And fsspec works on Windows!

docs/gallery/advanced_io/streaming.py
```diff
@@ -1,98 +1,114 @@
...
+fsspec is a library that can be used to access a variety of different store formats, including (at the time of
+writing):
+
+.. code-block:: python
+
+    from fsspec.registry import known_implementations
+    known_implementations.keys()
+
+file, memory, dropbox, http, https, zip, tar, gcs, gs, gdrive, sftp, ssh, ftp, hdfs, arrow_hdfs, webhdfs, s3, s3a, wandb, oci, adl, abfs, az, cached, blockcache, filecache, simplecache, dask, dbfs, github, git, smb, jupyter, jlab, libarchive, reference
+'''
```
bendichter, line 116:
```suggestion

The S3 backend, in particular, may provide additional functionality for accessing data on DANDI. See the
`fsspec documentation on known implementations <https://filesystem-spec.readthedocs.io/en/latest/api.html?highlight=S3#other-known-implementations>`_
for a full updated list of supported store formats.
'''
```

docs/gallery/advanced_io/streaming.py
```diff
@@ -1,98 +1,118 @@
...
-Streaming from an S3 Bucket
-===========================
+Streaming NWB files
+===================
 
-It is possible to read data directly from an S3 bucket, such as data from the `DANDI Archive
-<https://dandiarchive.org/>`_. This is especially useful for reading small pieces of data
-from a large NWB file stored remotely. In fact, there are two different ways to do this supported by PyNWB.
+You can read specific sections within individual data files directly from remote stores such as the
+`DANDI Archive <https://dandiarchive.org/>`_. This is especially useful for reading small pieces of data
```
bendichter:
```suggestion
`DANDI Archive <https://dandiarchive.org/>`_. This is especially useful for reading small pieces of data
```

oruebel:
HDF5 just published the following blog post https://www.hdfgroup.org/2022/08/cloud-storage-options-for-hdf5/ discussing the various cloud storage options with HDF5 and some of the trade-offs to consider. If we need a reference to point to, this may be a good link to point folks to for further reading about available options.

oruebel:
At some point it may be useful to also look at [HSDS](https://www.hdfgroup.org/solutions/highly-scalable-data-service-hsds/). From a PyNWB perspective it may be as simple as handing in a file handle from [h5pyd](https://github.com/HDFGroup/h5pyd), but I'm not sure. I've never used HSDS so I'm not sure how involved the setup on the server and client side are. @yarikoptic  has DANDI ever looked at HSDS.

oruebel (approved):

PR #1538, opened 2022-08-11
[BUG] Fix `TimeSeries` data does not match length of timestamps should raise an error when created

### LINKED ISSUE #1536: [Bug]: Timeseries can have timestamps + data that don't match in length
### What happened?

I use the convenient timeseries pointer feature, where you can just store timestamps in one timeseries and have many others point to them. This saves a lot of memory! However, I just fixed a bug that arose because some data I stored in a timeseries didn't have the same length as the timeseries (due to a processing mistake). Then I went and checked, and it turns out that timeseries' timestamps don't need to have the same length as the data at all! I consider this a bug, given that timeseries are meant as a core unit of data in NWB, but let me know if I'm just using them wrong.

### Steps to Reproduce

```python
# normal case
t = nwb.TimeSeries(name='test', data=np.arange(10), timestamps=np.arange(10), unit='sec')

# I expect a value error here
nwb.TimeSeries(name='test', data=np.arange(10), timestamps=np.arange(5), unit='sec')

# This similarly should raise an error imo
s = nwb.TimeSeries(name='test2', data=np.arange(2), timestamps=t, unit='sec')
```

### Traceback

```python
n/a
```

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.9

### Package Versions

nwb 2.0.0

### Code of Conduct

weiglszonja:
## Motivation

This is a follow up on the issue described in #1536 to prevent a `TimeSeries` being created with timestamps that does not match the length of data.

This PR changes `TimeSeries` `_check_time_series_dimension` check to warn when reading from file and raise an error when constructing new data.  

This change also affects `ImageSeries` check that will also raise a ValueError when an `ImageSeries`  is being created with timestamps that do not match the length of data. If `ImageSeries` is used with external file, then this error or warning will not be triggered.

Resolve #1536

## How to test the behavior?
```
tests/unit/test_base.py
tests/unit/test_image.py
```

## Checklist


weiglszonja:
@oruebel 
In `docs/gallery/domain/brain_observatory.py` `ImageSeries` are being created with timestamps of `[0]`, I tried to fix it but I don't have currently access to the data that is being used here to that was my guess. But it's not working yet.

rly:
@weiglszonja 

This code stores images in an ImageSeries so that they can be referenced by IndexSeries:
https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/gallery/domain/brain_observatory.py#L71-L85

This approach is no longer the recommended approach. Instead, use the `Images` and `Image` types with `IndexSeries` which was recently updated to support this. For example:
https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/tests/unit/test_image.py#L313-L325

bendichter:
@weiglszonja can you please resolve these conflicts?

bendichter:
@h-mayorquin can you take a look at these conflicts? It would be great to get this PR through

h-mayorquin:
> @weiglszonja
> 
> This code stores images in an ImageSeries so that they can be referenced by IndexSeries: https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/gallery/domain/brain_observatory.py#L71-L85

@rly 
I looked at the data and they are in fact Videos:

image

So I don't know how the Images, order of images, and friends would work very well here.

docs/gallery/domain/brain_observatory.py
```diff
@@ -71,8 +73,9 @@
         data=dataset.get_stimulus_template(stimulus),
         unit="NA",
         format="raw",
-        timestamps=[0.0],
+        rate=np.nan,
```
h-mayorquin: Timestamps does not make sense here because the images are repeated:

image

From the data, the rate is not very clear, I think it could be derived from the fluorescence timestamps but I don't really have the time to explore the dataset to that level.

h-mayorquin:
@rly @bendichter 
Looking at the inspector tests it looks that it fails because it can't create some of the files that the inspector requires on its testing. This makes sense as this PR makes that impossible. How to handle this?

rly:
> @rly @bendichter Looking at the inspector tests it looks that it fails because it can't create some of the files that the inspector requires on its testing. This makes sense as this PR makes that impossible. How to handle this?

We can create the test files in other ways in the nwbinspector. I'll create an issue/PR to adjust that.

rly:
@bendichter @stephprince It is strange that the brain observatory example uses the allensdk API to download an NWB file into a cache folder and then creates a new NWB file with the same data step by step. Do we want to maintain this example? I think everything that is demonstrated here is demonstrated in other tutorials (aside from the allensdk API).

This example also causes a minor headache with the test harness because it involves installing allensdk which is not well maintained with new versions of python, and it downloads a 450 MB file, which takes some time.

bendichter:
@rly I would be in support of removing that tutorial

bendichter:
Or at least demoting it to some sort of 3rd party untested status

stephprince:
@rly I'm also in support of removing that example

stephprince:
I think the allensdk API is demonstrated much more in depth in [their documentation](https://allensdk.readthedocs.io/en/latest/brain_observatory.html) and the pynwb examples are all demonstrated in other tutorials like you mentioned.

rly:
I opened #2026 which we can merge first to remove the Brain Observatory example.

rly (approved):

PR #1540, opened 2022-08-15
add age__reference and adjust tests

bendichter:
companion to https://github.com/NeurodataWithoutBorders/nwb-schema/pull/525


CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
 - Support explicit ordering of sphinx gallery tutorials in the docs. @oruebel (#1504), @bdichter (#1495)
 - Add developer guide on how to create a new tutorial. @oruebel (#1504)
 - Add images tutorial. @weiglszonja (#1470)
-- Add example code for s3fs in the streaming tutorial. @bdichter (#1499)
+- Add example code for fsspec in the streaming tutorial. @bdichter (#1499)
+- Add age__reference field. @bendichter ([#1540](https://github.com/NeurodataWithoutBorders/pynwb/pull/1540))
```
bendichter:
```suggestion
- Add age__reference field to Subject. @bendichter ([#1540](https://github.com/NeurodataWithoutBorders/pynwb/pull/1540))
```

src/pynwb/file.py
```diff
@@ -87,8 +95,11 @@ def __init__(self, **kwargs):
                        "date_of_birth",
                        "strain")
         args_to_set = popargs_to_dict(keys_to_set, kwargs)
-        kwargs['name'] = 'subject'
-        super().__init__(**kwargs)
+        super().__init__(name="subject", **kwargs)
+
+        if args_to_set["age__reference"] is not None:
+            if args_to_set["age__reference"] not in ("birth", "gestational"):
+                raise ValueError("age__reference must be 'birth' or 'gestational'.")
```
bendichter:
```suggestion
        if args_to_set["age__reference"] not in (None, "birth", "gestational"):
            raise ValueError("age__reference must be 'birth' or 'gestational'.")
```

src/pynwb/file.py
```diff
@@ -87,8 +95,10 @@ def __init__(self, **kwargs):
                        "date_of_birth",
                        "strain")
         args_to_set = popargs_to_dict(keys_to_set, kwargs)
-        kwargs['name'] = 'subject'
-        super().__init__(**kwargs)
+        super().__init__(name="subject", **kwargs)
+
+        if args_to_set["age__reference"] not in (None, "birth", "gestational"):
```
oruebel: For fields like this where we have an explicit limited list of allowable values, I think it could be useful to specify that list as a class variable or class method to get the list of allowable values. In this way users can more easily inspect which values are permitted.

oruebel: I'm not 100% sure, but I don't think the case ``args_to_set["age__reference"]  == None`` should ever occur. If ``age__reference`` is missing in the file, then the fact that ``birth`` is the default value means that you should see ``birth`` as the value in this case. As such, the only time it would be ``None`` is if the user explicitly sets the value to ``None``. Is this a case we should allow?

tests/unit/test_file.py
```diff
@@ -437,29 +437,35 @@ def test_multi_publications(self):
...
-                               description='An unfortunate rat',
-                               genotype='WT',
-                               sex='M',
-                               species='Rattus norvegicus',
-                               subject_id='RAT123',
-                               weight='2 kg',
-                               date_of_birth=datetime(2017, 5, 1, 12, tzinfo=tzlocal()),
-                               strain='my_strain')
+        self.subject = Subject(
+            age='P90D',
```
oruebel, line 441: Do we need a backward compatibility test for the case when age reference is missing?

https://github.com/NeurodataWithoutBorders/pynwb/tree/dev/tests/back_compat

src/pynwb/file.py
```diff
@@ -56,6 +56,13 @@ class Subject(NWBContainer):
     @docval({'name': 'age', 'type': str,
              'doc': ('The age of the subject. The ISO 8601 Duration format is recommended, e.g., "P90D" for '
                      '90 days old.'), 'default': None},
+            {
+                "name": "age__reference",
+                "type": str,
+                "doc": "Age is with reference to this event. Can be 'birth' or 'gestational'. If reference is "
+                       "omitted, 'birth' is implied.",
+                "default": "birth",
```
oruebel: For existing files where ``age.reference`` is missing, the user would, I believe, on read see that ``age.reference`` is set to ``birth`` instead of ``None``. I'm not sure if automatically setting this value to ``birth`` when it is missing in a file is desired behavior? I think it may be more intuitive to have ``age.reference`` appear as None if it is missing in the file while still having a default value of `birth` when a user creates a new file. To achieve this, we could either: 1) update the object mapper to set the constructor argument to None if it is missing (or issue a warning that it will be set to ``birth``), or 2) have the default value in ``__init__`` depend on ``self._in_construct_mode``, i.e., ``None if self._in_construct_mode else 'birth'.

src/pynwb/file.py
bendichter: @oruebel 

default already has the behavior you are looking for
```python
from datetime import datetime
from dateutil.tz import tzlocal, tzutc

from pynwb.file import NWBFile, Subject

from pynwb import NWBHDF5IO

subject = Subject(
    age='P90D',
    #age__reference="birth",
    subject_id='RAT123',
)
nwbfile = NWBFile(
    'a test session description for a test NWBFile',
    'FILE123',
    datetime(2017, 5, 1, 12, tzinfo=tzlocal()),
    session_id='test1',
    subject=subject,
)

with NWBHDF5IO("test_age_ref.nwb", "w") as io:
    io.write(nwbfile)
    
with NWBHDF5IO("test_age_ref.nwb", "r") as io:
    nwb = io.read()
    print(nwb.subject.age__reference)
```

```python
None
```

src/pynwb/file.py
bendichter: default already has the behavior you are looking for:

```python
from datetime import datetime
from dateutil.tz import tzlocal, tzutc

from pynwb.file import NWBFile, Subject

from pynwb import NWBHDF5IO

subject = Subject(
    age='P90D',
    #age__reference="birth",
    subject_id='RAT123',
)
nwbfile = NWBFile(
    'a test session description for a test NWBFile',
    'FILE123',
    datetime(2017, 5, 1, 12, tzinfo=tzlocal()),
    session_id='test1',
    subject=subject,
)

with NWBHDF5IO("test_age_ref.nwb", "w") as io:
    io.write(nwbfile)
    
with NWBHDF5IO("test_age_ref.nwb", "r") as io:
    nwb = io.read()
    print(nwb.subject.age__reference)
```
```python
None
```

tests/unit/test_file.py
rly: Yes, we should add a test just in case.

src/pynwb/file.py
oruebel: Interesting, I didn't know the ObjectMapper was by default setting it to None if missing (I suspected it would probably just leave it blank and leave the default specified for __init__). This makes sense. Thanks for the demonstration! That's good to know.

src/pynwb/file.py
```diff
@@ -87,8 +95,13 @@ def __init__(self, **kwargs):
...
         args_to_set = popargs_to_dict(keys_to_set, kwargs)
-        kwargs['name'] = 'subject'
-        super().__init__(**kwargs)
+        super().__init__(name="subject", **kwargs)
+
+        if args_to_set["age__reference"] is not None:
+            if args_to_set["age__reference"] not in ("birth", "gestational"):
+                raise ValueError("age__reference must be 'birth' or 'gestational'.")
+            if args_to_set["age"] is None:
+                raise ValueError("If age__reference is provided, age must also be provided.")
```
oruebel: Since ``age__reference`` has a default value, would this not always raise an error if ``age`` is not provided by the user?  In particular, ``age`` has a default of ``None`` so I'm wondering whether creating the object with the defaults would trigger this ValueError.

src/pynwb/file.py
```diff
@@ -87,8 +95,13 @@ def __init__(self, **kwargs):
                        "date_of_birth",
                        "strain")
         args_to_set = popargs_to_dict(keys_to_set, kwargs)
-        kwargs['name'] = 'subject'
-        super().__init__(**kwargs)
+        super().__init__(name="subject", **kwargs)
+
+        if args_to_set["age__reference"] is not None:
+            if args_to_set["age__reference"] not in ("birth", "gestational"):
+                raise ValueError("age__reference must be 'birth' or 'gestational'.")
```
oruebel:
```suggestion
            valid_age__reference_values = ("birth", "gestational")
            if args_to_set["age__reference"] not in valid_age__reference_values:
                raise ValueError("age__reference must be in %s." % str(valid_age__reference_values))
```

src/pynwb/file.py
bendichter: oh yeah good point

CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
 - Support explicit ordering of sphinx gallery tutorials in the docs. @oruebel (#1504), @bdichter (#1495)
 - Add developer guide on how to create a new tutorial. @oruebel (#1504)
 - Add images tutorial. @weiglszonja (#1470)
-- Add example code for s3fs in the streaming tutorial. @bdichter (#1499)
+- Add example code for fsspec in the streaming tutorial. @bdichter (#1499)
```
bendichter:
```suggestion
- Add example code for fsspec in the streaming tutorial. @bendichter (#1499)
```

rly:
@bendichter Can you add a roundtrip test and a backwards compatibility check that checks that None is returned if `age__reference` is not in the file?

bendichter:
@rly In writing this test, I came up against an error in hdmf:

```python
class TestSubjectAgeReference(TestSubjectIO):

    def setUpContainer(self):
        """ Return the test Subject """
        return Subject(
            age='P90D',
            age__reference=None,
            description='A rat',
            genotype='WT',
            sex='M',
            species='Rattus norvegicus',
            subject_id='RAT123',
            weight='2 kg',
            date_of_birth=datetime(1970, 1, 1, 12, tzinfo=tzutc()),
            strain='my_strain',
        )
```

```pytb
args = (<[AttributeError("'Subject' object has no attribute '_AbstractContainer__name'") raised in repr()] Subject object at 0x7fa562325130>,)
kwargs = {'age': 'P90D', 'age__reference': None, 'date_of_birth': datetime.datetime(1970, 1, 1, 12, 0, tzinfo=tzutc()), 'description': 'A rat', ...}

    def _check_args(args, kwargs):
        """Parse and check arguments to decorated function. Raise warnings and errors as appropriate."""
        # this function was separated from func_call() in order to make stepping through lines of code using pdb
        # easier
        parsed = __parse_args(
            loc_val,
            args[1:] if is_method else args,
            kwargs,
            enforce_type=enforce_type,
            enforce_shape=enforce_shape,
            allow_extra=allow_extra,
            allow_positional=allow_positional
        )
    
        parse_warnings = parsed.get('future_warnings')
        if parse_warnings:
            msg = '%s: %s' % (func.__qualname__, ', '.join(parse_warnings))
            warnings.warn(msg, FutureWarning)
    
        for error_type, ExceptionType in (('type_errors', TypeError),
                                          ('value_errors', ValueError),
                                          ('syntax_errors', SyntaxError)):
            parse_err = parsed.get(error_type)
            if parse_err:
                msg = '%s: %s' % (func.__qualname__, ', '.join(parse_err))
>               raise ExceptionType(msg)
E               TypeError: Subject.__init__: incorrect type for 'age__reference' (got 'NoneType', expected 'str or NoneType')
```

It looks like docval does not accept an input of `None`

bendichter:
It looks like there is an `allow_none` flag, but I don't see how to set it: https://github.com/hdmf-dev/hdmf/blob/bb156780017a5ca19b7f148f4452862761bef751/src/hdmf/utils.py#L70

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
 ## PyNWB 2.1.1 (September 1, 2022)
 
 ### Documentation and tutorial enhancements:
+
```
bendichter:
```suggestion
```

src/pynwb/file.py
```diff
@@ -64,6 +64,7 @@ class Subject(NWBContainer):
         'strain'
     )
 
+
```
bendichter:
```suggestion
```

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@
 - Added example code for fsspec in the streaming tutorial. @bdichter
   [#1499](https://github.com/NeurodataWithoutBorders/pynwb/pull/1499)
 - Add voxel_mask tutorial. @codycbakerphd (#1544)
+- Add `age__reference` field to Subject. @bendichter ([#1540](https://github.com/NeurodataWithoutBorders/pynwb/pull/1540))
```
bendichter:
```suggestion
```

rly:
@bendichter Can you add
1. a roundtrip test and
2. a backwards compatibility check that checks that age__reference = None if age__reference is not found in the file?

src/pynwb/file.py
bendichter: @oruebel actually I think you were right about this. Now when I read the data back I get "birth" even when it's `None` in the file.

```python
from datetime import datetime
from dateutil.tz import tzlocal, tzutc

from pynwb.file import NWBFile, Subject

from pynwb import NWBHDF5IO

subject = Subject(
    age='P90D',
    age__reference=None,
    subject_id='RAT123',
)
nwbfile = NWBFile(
    'a test session description for a test NWBFile',
    'FILE123',
    datetime(2017, 5, 1, 12, tzinfo=tzlocal()),
    session_id='test1',
    subject=subject,
)

with NWBHDF5IO("test_age_ref.nwb", "w") as io:
    io.write(nwbfile)
    
with NWBHDF5IO("test_age_ref.nwb", "r") as io:
    nwb = io.read()
    print(nwb.subject.age__reference)
```

birth

bendichter: Could you show me how to update the mapper to handle this properly?

rly:
tl;dr I think we should not allow the case where `Subject(age__reference=None)` sets the age_reference to None which means no value is written, and then on read, the value is set to the default value of "birth". Instead, you can either pass a valid value for `age__reference` or don't include it at all. I updated the PR accordingly. 

Explanation:
Based on [the spec for default_value](https://hdmf-schema-language.readthedocs.io/en/latest/description.html#default-value), if the schema says `default_value: birth` and a file with that schema is read with the attribute missing, then PyNWB should read the value as "birth". However, for old files generated when this attribute did not exist, I think PyNWB should read the value as None because it is governed by an older schema. This is mostly the behavior created by this PR.

Use cases:
1) Read NWB 2.6 file, read age__reference = "birth" (or gestational) - value is read as "birth" (or gestational) ✅ 
2) Read NWB 2.6 file, age__reference is missing - value is read as "birth" ✅ 
3) Read NWB 2.0-2.5 file, age__reference is missing - value is read as None ✅ 
4) Write NWB 2.6 file, `Subject(age__reference="birth")` (or gestational) - value is written as "birth" (or gestational) ✅ 
5) Write NWB 2.6 file, `Subject(age__reference=None)` - no value is written ✅ 
    a) In memory before write, age__reference has value None but I think either:
       A) age_reference should be set to the default value "birth". Otherwise, roundtrip tests fail, because on read, age__reference is read as "birth". Or 
       B) we should just not allow this case. If the user wants to set age__reference, then only "birth" and "gestational" are allowed. It would be confusing if the user explicitly sets the value to None and it gets changed to "birth" in A. 
    So I think we should do B and I have updated the PR accordingly.❓
6) Write NWB 2.6 file, `Subject()` - no value is written ✅ 
    a) In memory before write, age__reference has value "birth" and on read, age__reference is read as "birth" ✅ 

To implement what I think the logic should be above, then on read of an NWB file, the value of age__reference when it is not in the file depends on the schema version. See the mapper code I added.

rly (dismissed):

rly:
I will add a unit test for `get_nwb_version` later today and then this should be good to go.

rly (approved):

rly:
@bendichter This is good to go now. Please review my changes. 

Note that when a user calls `Subject(age__reference=None)` or `Subject()`, then `age__reference` will be set to the default value "birth" for reasons described above. The only way that `age__reference` can be None is if the file uses NWB schema < 2.6.0 and the attribute is missing.

PR #1546, opened 2022-08-27
Enhance LabMetaData docstring

### LINKED ISSUE #1542: [Documentation]: guidance on how to use LabMetaData

I've been looking for some documentation on how to use the `LabMetaData` object to potentially add some extra notes and pieces of information to our data files, but I'm having trouble finding clear documentation and examples for how to do so. On the pynwb documentation website, for example, the API listing doesn't really make it clear on how to use this approach to add any information. 

Based on searching for examples of people using this, I think using NWBfile.add_lab_meta_data perhaps also requires using the extensions tool to define a custom LabMetaData object to be added? This makes sense, but I don't find it very clear from looking at pywnb documentation. Based on poking around with `pynwb` it's confusing how it works, since it seems on can only really generate some kind of empty container with a name. I's not obvious that, I think, LabMetaData is some kind of base that should be inherited by some kind of extension (?). Perhaps this functionality is documented somewhere else and I'm missing it, but if not, I think there could be some updated documentation for this functionality. 

For example, some potential documentation updates:
- add some notes to the docstring of `add_lab_meta_data`, `create_lab_meta_data`, and/or `LabMetaData` about what this is / how it works, and update accepted inputs, etc
   - for example, `add_lab_meta_data`, which on the face of it seems like it could take meta data in general (rather than a particular LabMetaData object), says that lists and dicts are acceptable inputs, but I think this really means lists or dicts of LabMetaData objects? Passing in a list of a dict otherwise leads to non-obvious errors. 
- add an example of creating and using a custom lab meta data object, perhaps on the "Extending NWB" page?

Yes, but I would need guidance.

### Code of Conduct

oruebel:
## Motivation

Fix #1542

## How to test the behavior?

```
from pynwb.file import LabMetaData
help(LabMetaData)
```

## Checklist

### DIFF
--- a/src/pynwb/file.py
+++ b/src/pynwb/file.py
@@ -30,7 +30,20 @@ def _not_parent(arg):
 
 @register_class('LabMetaData', CORE_NAMESPACE)
 class LabMetaData(NWBContainer):
-    """Container for storing lab-specific meta-data"""
+    """
+    Container for storing lab-specific meta-data
+
+    The LabMetaData class serves as a base type for defining lab specific meta-data.
+    To define your own lab-specific metadata, create a Neurodata Extension (NDX) for
+    NWB that defines the data to add. Using the LabMetaData container as a base type
+    makes it easy to add your data to an NWBFile without having to modify the NWBFile
+    type itself, since adding of LabMetaData is already implemented. This allows
+    extensions based on LabMetaData to be typically fairly simple and helps avoid the
+    need for custom Container classes in many cases. For more details
+    on how to create an extension see the
+    :nwb_overview:`Extending NWB <extensions_tutorial/extensions_tutorial_home.html>`
+    tutorial.
+    """
 
     @docval({'name': 'name', 'type': str, 'doc': 'name of lab metadata'})
     def __init__(self, **kwargs):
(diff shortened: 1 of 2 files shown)


src/pynwb/file.py
```diff
@@ -30,7 +30,20 @@ def _not_parent(arg):
...
+
+    The LabMetaData class serves as a base type for defining lab specific meta-data.
+    To define your own lab-specific metadata, create a Neurodata Extension (NDX) for
+    NWB that defines the data to add. Using the LabMetaData container as a base type
+    makes it easy to add your data to an NWBFile without having to modify the NWBFile
+    type itself, since adding of LabMetaData is already implemented. This allows
+    extensions based on LabMetaData to be typically fairly simple and helps avoid the
+    need for custom Container classes in many cases. For more details
+    on how to create an extension see the
+    :nwb_overview:`Extending NWB <extensions_tutorial/extensions_tutorial_home.html>`
```
bendichter:
```suggestion
    type itself, since adding of LabMetaData is already implemented. For more details
    on how to create an extension see the
    :nwb_overview:`Extending NWB <extensions_tutorial/extensions_tutorial_home.html>`
```

bendichter: I find the last sentence confusing, since an extension of LabMetaData *would be* a custom Container class. Also, I think the point is made without this sentence

rly (dismissed):

rly:
Please add a changelog and I think this is good to go.

oruebel:
> Please add a changelog and I think this is good to go.

Done. This PR is ready. @rly can you please approve and merge.

rly (approved):

PR #1576, opened 2022-10-20
Update requirements-min.txt

bendichter:
I think min requirements should be gte. That way they don't cause conflicts with other packages.

### DIFF
--- a/requirements-min.txt
+++ b/requirements-min.txt
@@ -1,7 +1,7 @@
 # minimum versions of package dependencies for installing PyNWB
-h5py==2.10  # support for selection of datasets with list of indices added in 2.10
-hdmf==3.4.0
-numpy==1.16
-pandas==1.1.5
-python-dateutil==2.7.3
+h5py>=2.10  # support for selection of datasets with list of indices added in 2.10
+hdmf>=3.4.0
+numpy>=1.16
+pandas>=1.1.5
+python-dateutil>=2.7.3
 setuptools


rly:
This file is used only in CI. Hm. Maybe we should move it to the GitHub workflows folder so it is less confusing. The package run requirements can be found in `setup.py`

bendichter:
Ah ok. The larger is that I want pynwb to be installed with flexible reqs. Looking more closely, I see we **do** use flexible reqs here: https://github.com/NeurodataWithoutBorders/pynwb/blob/fdb0297ad1ba472ee8e3433559e2e3ef556c0cde/setup.py#L21-L28

but it's unclear to me why these are hard-coded in the `setup.py` file instead of being in the `requirements.txt` file.

rly:
`setup.py` and `requirements.txt` conventionally serve different purposes: https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/

`requirements.txt` specifies pinned dependencies for making a perfectly reproducible environment and is usually the result of running `pip freeze > requirements.txt`. Ours is a little different -- we pin just the top-level dependencies and not their dependencies. This file is typically used only for development when devs do `pip install -r requirements.txt` or in CI workflows.

`setup.py` specifies the top-level dependencies and their range of versions for running the package. This is used by general users when they do `pip install package` from PyPI or `pip install -e .` on the package locally.

bendichter:
@rly thanks for the pointer. This makes sense. I'll close the issue

PR #1581, opened 2022-11-04
Update ecephys.py

bendichter:
## Motivation

The ecephys tutorial currently generates data of shape (50, 4), but there are 12 electrodes so this fails inspection, generating two critical `InspectorMessage` objects:

```python
from nwbinspector import inspect_nwb
list(inspect_nwb("ecephys_tutorial.nwb"))
```

```python
[InspectorMessage(
     message='The second dimension of data does not match the length of electrodes. Your data may be transposed.',
     importance=<Importance.CRITICAL: 2>,
     severity=<Severity.LOW: 1>,
     check_function_name='check_electrical_series_dims',
     object_type='ElectricalSeries',
     object_name='ElectricalSeries',
     location='/processing/ecephys/LFP/ElectricalSeries',
     file_path='/Users/bendichter/dev/sandbox/ecephys_tutorial.nwb'
 ),
 InspectorMessage(
     message='The second dimension of data does not match the length of electrodes. Your data may be transposed.',
     importance=<Importance.CRITICAL: 2>,
     severity=<Severity.LOW: 1>,
     check_function_name='check_electrical_series_dims',
     object_type='ElectricalSeries',
     object_name='ElectricalSeries',
     location='/acquisition/ElectricalSeries',
     file_path='/Users/bendichter/dev/sandbox/ecephys_tutorial.nwb'
 ), ...
```

The shape needs to be changed to (50, 12), which resolves both of these messages.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # PyNWB Changelog
 
+## Upcoming
+
+### Documentation and tutorial enhancements:
+- Adjusted [ecephys tutorial](https://pynwb.readthedocs.io/en/stable/tutorials/domain/ecephys.html) to create fake data with proper dimensions @bendichter [#1581](https://github.com/NeurodataWithoutBorders/pynwb/pull/1581)
+
 ## PyNWB 2.2.0 (October 19, 2022)
 
 ### Enhancements and minor changes
--- a/docs/gallery/domain/ecephys.py
+++ b/docs/gallery/domain/ecephys.py
@@ -130,7 +130,7 @@ nwbfile.electrodes.to_dataframe()
 # .. _ec_recordings:
 #
 # Extracellular recordings
-# ------------------------------
+# ------------------------
 #
 # Raw voltage traces and local-field potential (LFP) data are stored in :py:class:`~pynwb.ecephys.ElectricalSeries`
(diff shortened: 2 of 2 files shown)


CodyCBakerPhD:
(i) do we have a tutorial that shows how to use a DynamicTableRegion to subselect from a larger electrodes table (if you have multiple probes and want to split the ElectricalSeries for some reason)

(ii) does the inspector correctly assess the shape from those subselected DynamicTableRegions?

oruebel (approved):

oruebel:
> (ii) does the inspector correctly assess the shape from those subselected DynamicTableRegions?

Since these are 1D arrays in the case of electrodes, the shape is just the length of the dataset so as long as ``ElectricSeries.electrodes.shape[0] == ElectricalSeries.data.shape[0]`` the check should be correct independent of the size of the ElectrodesTable. If you also want to check that the indices are valid you would need to  chck that ``np.min(ElectricSeries.electrodes[:]) >=0 and np.max(ElectricSeries.electrodes[:]) < len(ElectricSeries.electrodes.table)``. The latter check is not quite correct, because you want to load the indices not the actual electrodes so ``[:]`` is not correct. 


> (i) do we have a tutorial that shows how to use a DynamicTableRegion to subselect from a larger electrodes table (if you have multiple probes and want to split the ElectricalSeries for some reason)

Not sure, but that would be good to add to the ECEphys tutorial if it is missing.

PR #1587, opened 2022-11-09
fix mock tests

bendichter:
## Motivation

mock tests were not running. This fixes them.

### DIFF
--- a/tests/unit/test_mock.py
+++ b/tests/unit/test_mock.py
@@ -77,4 +77,3 @@ def test_name_generator():
 
     assert name_generator("TimeSeries") == "TimeSeries"
     assert name_generator("TimeSeries") == "TimeSeries2"
-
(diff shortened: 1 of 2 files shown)


tests/unit/test_mock.py
```diff
@@ -29,33 +34,47 @@
...
+        mock_TimeSeries,
+    ],
+)
+def test_mock(mock_function):
+    mock_function()
+
+
+def test_name_generator():
+
+    from pynwb.testing.mock.utils import name_generator_registry
```
rly: Does this import statement need to be in the function?

bendichter: yes, I think it does. it's a dictionary global variable that I need to reset for this test. I'm not sure how else to do it and I suspect this is the simplest way

rly: I get that the `clear()` needs to be here, but I am pretty sure that in this case, the import statement can be outside the function. I tested this locally and in a simplified MWE and it works fine. 

`from pynwb.testing.mock.utils import name_generator, name_generator_registry`

rly (approved):

PR #1588, opened 2022-11-09
Add driver argument to new validation functions

CodyCBakerPhD:
When working on https://github.com/NeurodataWithoutBorders/nwbinspector/pull/265 (the integration of the recent validation changes into the NWB Inspector) I ran into an issue getting the cached namespaces for files when using the streaming mode of the Inspector.

Traced it back to these particular function calls - thankfully, `NWBHDF5IO.load_namespaces(...)` already supports a `driver` keyword argument, we just need to add control and propagate it as shown here.

@rly What would your recommendations be for adding tests for this?

### DIFF
--- a/src/pynwb/validate.py
+++ b/src/pynwb/validate.py
@@ -1,7 +1,7 @@
 """Command line tool to Validate an NWB file against a namespace."""
 import sys
 from argparse import ArgumentParser
-from typing import Tuple, List, Dict
+from typing import Tuple, List, Dict, Optional
 
 from hdmf.spec import NamespaceCatalog
 from hdmf.build import BuildManager
@@ -29,7 +29,7 @@ def _validate_helper(io: HDMFIO, namespace: str = CORE_NAMESPACE) -> list:
     return validator.validate(builder)
 
 
-def _get_cached_namespaces_to_validate(path: str) -> Tuple[List[str], BuildManager, Dict[str, str]]:
+def _get_cached_namespaces_to_validate(path: str, driver: Optional[str] = None) -> Tuple[List[str], BuildManager, Dict[str, str]]:
     """
     Determine the most specific namespace(s) that are cached in the given NWBFile that can be used for validation.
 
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
I'll also add the `file` argument propagation in a separate PR for `fsspec` based streaming. That will require just a little bit more since it would need to not conflict with the `io vs. path` usage and just be an alternate usage of the `path`

rly:
This looks good. You can add to https://github.com/NeurodataWithoutBorders/pynwb/tree/dev/tests/integration/ros3 which is the only other place in the tests where we test different drivers.

src/pynwb/validate.py
```diff
@@ -107,6 +109,12 @@ def _get_cached_namespaces_to_validate(path: str) -> Tuple[List[str], BuildManag
         "doc": "Whether or not to print messages to stdout.",
         "default": False,
     },
+    {
+        "name": "driver",
+        "type": list,
```
bendichter: shouldn't "driver" be a string, not a list?

bendichter:
This looks like a fix specifically for ros3. It's good to support, but let's try to shift our default over to fsspec for all streaming stuff

src/pynwb/validate.py
CodyCBakerPhD:
```suggestion
        "type": str,
```

CodyCBakerPhD: yes

CodyCBakerPhD:
> This looks like a fix specifically for ros3. It's good to support, but let's try to shift our default over to fsspec for all streaming stuff

https://github.com/NeurodataWithoutBorders/pynwb/pull/1588#issuecomment-1309263552

> I'll also add the file argument propagation in a separate PR for fsspec based streaming. That will require just a little bit more since it would need to not conflict with the io vs. path usage and just be an alternate usage of the path

CodyCBakerPhD:
> This looks good. You can add to https://github.com/NeurodataWithoutBorders/pynwb/tree/dev/tests/integration/ros3 which is the only other place in the tests where we test different drivers.

OK - added tests for the new keyword injections

rly (approved):

CodyCBakerPhD:
Oh wow, forgot about this one. Thanks @rly

PR #1593, opened 2022-11-10
Adding OnePhotonSeries class

CodyCBakerPhD:
## Motivation

Adding the `OnePhotonSeries` class to PyNWB, which was recently added to the schema in https://github.com/NeurodataWithoutBorders/nwb-schema/pull/523

PR #1427, opened 2022-02-02
Fix miniconda 3.7 CI error

rly:
## Motivation

The CI for miniconda 3.7 on CircleCI is broken due to an incompatibility between some combination of Python 3.7, tox, virtualenv, and importlib-metadata.
```
AttributeError: 'dict' object has no attribute 'select'
```

See also https://github.com/pypa/virtualenv/issues/2174

Fixing this seems to then break the pandas installation. This has yet to be resolved.

This affects both PyNWB and HDMF.

## Checklist

- [x] Did you update CHANGELOG.md with your changes?
- [x] Have you checked our [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [x] Have you ensured the PR clearly describes the problem and the solution?
- [x] Is your contribution compliant with our coding style? This can be checked running `flake8` from the source directory.
- [x] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/NeurodataWithoutBorders/pynwb/pulls) for the same change?
- [x] Have you included the relevant issue number using "Fix #XXX" notation where XXX is the issue number? By including "Fix #XXX" you allow GitHub to close issue #XXX when the PR is merged.

PR #1611, opened 2022-12-15
Add support for NWB schema 2.6.0 and prepare PyNWB 2.3.0

rly:
NWB schema 2.6.0 will be released around January 5, 2023.

Currently, it contains the following [changes](https://github.com/NeurodataWithoutBorders/nwb-schema/blob/dev/docs/format/source/format_release_notes.rst):
- Added OnePhotonSeries.
- ``Subject.age`` has a new optional attribute, ``reference``, which can take a value of "birth" (default) or "gestational".
- Add "in seconds" to the doc of Units.spike_times.

Required PyNWB changes:
- [x] Add support for `Subject.age__reference`: #1540
- [x] Add support for `OnePhotonSeries`: #1593

When merging this branch into `dev`, do NOT squash and merge, so that individual PR contributions are kept. 

This branch should be merged after the NWB schema 2.6.0 release. A PyNWB release (2.2.0) should be made shortly after.


src/pynwb/io/utils.py
```diff
@@ -0,0 +1,27 @@
...
+from typing import Tuple
+
+from hdmf.build import Builder
+
+
+def get_nwb_version(builder: Builder) -> Tuple[int, ...]:
+    """Get the version of the NWB file from the root of the given builder, as a tuple.
+
+    If the "nwb_version" attribute on the root builder equals "2.5.1", then (2, 5, 1) is returned.
+    If the "nwb_version" attribute on the root builder equals "2.5.1-alpha", then (2, 5, 1) is returned.
```
oruebel: Why is the ``alpha`` being dropped in the version? It seems strange that "2.5.1" and "2.5.1-alpha" both translate to the same version?

rly: The version tuple is used to branch the API behavior based on the version, e.g., do X if version < 2.5.1, else Y. Version "2.5.1-alpha" should appear only in development. The features in schema "2.5.1-alpha" will be included in "2.5.1" and so for this purpose, the downstream behavior should treat "2.5.1-alpha" and "2.5.1" as equivalent. 

We could refactor this function so that a suffix like "-alpha" is also returned. In Python `(2, 5, 1, "-alpha") > (2, 5, 1)`, so the branching code above should still work.

oruebel: I would suggest to maybe add a flag (e.g., ``include_prerelease=False``) so that a user can configure whether the prerelease tag (alpha, beta etc.) should be included or not.

rly: done

mavaylon1:
@rly I fixed the sphinx issue but the conda 3.7 from #1427 is now appearing again.

mavaylon1:
Fix #1642

mavaylon1 (approved):

PR #1612, opened 2022-12-16
Check nwb_version on read

### LINKED ISSUE #1609: [Feature]: Enhance NWBHDF5IO to check NWB version

NWBHDFI5O should check (either on __init__ or when calling read) what the version of the file is an raise and error if the  NWB version is 1.x

This is related to https://github.com/NeurodataWithoutBorders/nwbinspector/issues/329 Currently NWBHDF5IO fails on read for NWB 1.x files with an error message that is hard to debug.

* Add function to get the nwb version from a file to NWBHDF5IO 
* Either on __init__ or when calling read, NWBHDF5IO should check the version and raise an error if the version is 1.x

Yes.

### Code of Conduct

oruebel:
## Motivation

Fix #1609

This PR:
* Adds the function ``NWBHDF5IO.nwb_version`` to check the nwb_version of an HDF5 file (i.e., before reading the file)
* Adds checking of nwb_version to ``NWBHDF5IO.read`` to raise a more informative error in case unsupported in case that unsupported file versions (i.e., NWB 1 files) are found
* Adds tests for the new functions

## Checklist

### DIFF
--- a/tests/integration/hdf5/test_io.py
+++ b/tests/integration/hdf5/test_io.py
@@ -418,6 +418,30 @@ class TestNWBHDF5IO(TestCase):
     def tearDown(self):
         remove_test_file(self.path)
 
+    def test_nwb_version_property(self):
+        """Test reading of files with missing nwb_version"""
+        # check empty version before write
+        with NWBHDF5IO(self.path, 'w') as io:
+            self.assertIsNone(io.nwb_version)
+        # write the example file
+        with NWBHDF5IO(self.path, 'w') as io:
+            io.write(self.nwbfile)
+        # check behavior for various different version strings
+        for ver in [("2.0.5", (2, 0, 5)),
+                    ("2.0.5-alpha", (2, 0, 5, "alpha")),
+                    ("bad_version", ("bad_version", ))]:
+            # Set version string
+            with File(self.path, mode='a') as io:
+                io.attrs['nwb_version'] = ver[0]
+            # Assert expected result for nwb_version tuple
+            with NWBHDF5IO(self.path, 'r') as io:
+                self.assertTupleEqual(io.nwb_version, ver[1])
+        # check empty version attribute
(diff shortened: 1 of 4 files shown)


oruebel:
@rly this PR is ready for review

tests/integration/hdf5/test_io.py
```diff
@@ -418,9 +418,80 @@ def setUp(self):
...
+        # write the example file
+        with NWBHDF5IO(self.path, 'w') as io:
+            io.write(self.nwbfile)
+        # remove the version attribute
+        with File(self.path, mode='a') as io:
+            io.attrs['nwb_version'] = "1.0.5"
+        # test that reading the file without a version strings fails
+        with self.assertRaisesWith(
+                TypeError,
+                "NWB version (1, 0, 5) not supported. PyNWB supports NWB files version 2 and above."):
```
rly: Can you make this print "NWB version 1.0.5 not supported. ..." instead? I think the tuple form is not intuitive to most users.

rly: Also note that in this form "1.0.4_beta" will print as `(1, 0, '4_beta')`

oruebel: Thanks for the suggestion @rly. I have now updated ``NWBHDF5IO.nwb_version`` to: 1) return both the original version string and the parsed version tuple and 2) added support for ".", "-", and "_" as delimiters so that ``"1.0.4_beta`` is now parsed correctly to ``(1, 0, 4, 'beta')``.  I also updated the error message in ``NWBHDF5IO.read`` to use the original version string as it appears in the file (rather than reconstructing the version string from the tuple) to make sure that all the delimiters of the version string also match in the error message. I also updated the tests accordingly.

rly: Looks good. Thanks

rly (approved):

PR #1630, opened 2023-01-08
Show recommended usaege for hdf5plugin in tutorial

### LINKED ISSUE #1629: How to pass dynamic filters compression options to `H5DataIO`?
The obvious way to me would be to do something like:

```python
import hdf5plugin
from hdmf.backends.hdf5.h5_utils import H5DataIO
import numpy as np

data = np.random.rand(10, 2)
data.shape

plug_in = hdf5plugin.BZip2()
filter_id = plug_in.filter_id
compression_opts = dict(blocksize=7)
wrapped_data = H5DataIO(data=data, compression=filter_id, allow_plugin_filters=True, compression_opts=compression_opts)
```
However, because [H5DataIO](https://hdmf.readthedocs.io/en/stable/hdmf.backends.hdf5.h5_utils.html#hdmf.backends.hdf5.h5_utils.H5DataIO) expects either an int or a tuple for the `conversion_opts` this throws an assertion:

```python
TypeError: H5DataIO.__init__: incorrect type for 'compression_opts' (got 'dict', expected 'int or tuple')
```

I guess that the final goal is to store this into `__iosettings`, so this is properly mapped when `create_dataset` is called at the moment of writting. From that perspective, maybe this can be done by only adding dict as a possible type on `conversion_opts`?

oruebel:
## Motivation

Fix #1629 to show recommended usage for hdf5plugin to set compression and compression_opts parameters for H5DataIO

## Checklist

### DIFF
--- a/docs/gallery/advanced_io/h5dataio.py
+++ b/docs/gallery/advanced_io/h5dataio.py
@@ -234,7 +234,7 @@ from pynwb.file import TimeSeries
 
 wrapped_data = H5DataIO(
     data=data,
-    **hdf5plugin.Zstd().filter_id,  # this sets the compression and compression_opts parameters
+    **hdf5plugin.Zstd().filter_id,  # set the compression and compression_opts parameters
     allow_plugin_filters=True,
 )
 
--- a/docs/gallery/advanced_io/h5dataio.py
+++ b/docs/gallery/advanced_io/h5dataio.py
@@ -234,7 +234,7 @@ from pynwb.file import TimeSeries
 
 wrapped_data = H5DataIO(
     data=data,
-    compression=hdf5plugin.Zstd().filter_id,
+    **hdf5plugin.Zstd().filter_id,  # this sets the compression and compression_opts parameters
     allow_plugin_filters=True,
 )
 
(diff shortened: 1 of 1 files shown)


docs/gallery/advanced_io/h5dataio.py
```diff
@@ -234,7 +234,7 @@
 
 wrapped_data = H5DataIO(
     data=data,
-    compression=hdf5plugin.Zstd().filter_id,
+    **hdf5plugin.Zstd().filter_id,  # this sets the compression and compression_opts parameters
```
oruebel:
```suggestion
    **hdf5plugin.Zstd().filter_id,  # set the compression and compression_opts parameters
```

docs/gallery/advanced_io/h5dataio.py
```diff
@@ -234,7 +234,7 @@
 
 wrapped_data = H5DataIO(
     data=data,
-    compression=hdf5plugin.Zstd().filter_id,
+    **hdf5plugin.Zstd().filter_id,  # set the compression and compression_opts parameters
```
h-mayorquin:
```suggestion
    **hdf5plugin.Zstd(clevel=3),  # set the compression and compression_opts parameters
```
The current diff gets the attribute `filter_id` so it is wrong.
I think having the options over there explicitly (clevel in this case) also helps the users to understand that this is the way that the options are meant to be passed. At least it would have helped me.

bendichter (approved):

docs/gallery/advanced_io/h5dataio.py
oruebel: Thanks for the fix

PR #1641, opened 2023-02-02
Increase stacklevel of timezone warning

CodyCBakerPhD:
## Motivation

I just ran across a confusing case during a conversion.

I'm setting both `session_start_time` and the subect's `date_of_birth`.

I intentionally set the `tzinfo` for the `datetime` of the `session_start_time`, but was very confused for while as to why I kept getting the warning from PyNWB about not having set a timezone.

Little did I know it was not coming from the `session_start_time` (which 90+% of the time is the cause of that warning), but it was from https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/file.py#L120 which does the same thing to the subject `date_of_birth`.

Using a higher `stacklevel` would have saved me a lot of time figuring this out, because it would have shown me a line in the traceback related to the field that was being parsed.

## Checklist


rly (dismissed):

rly:
Although it is not that relevant to this PR, could you please add a test for `_add_missing_timezone` in this or another PR? There is apparently no coverage of that line.

rly:
Otherwise, looks good to me. Thanks!

rly (approved):

PR #1643, opened 2023-02-08
revert from :py:code: to ``

bendichter:


rly (approved):

rly:
thanks! and the RTD build revealed the TypeError better I think

PR #1669, opened 2023-03-16
Support nwb_version as a fixed-length string

### LINKED ISSUE #1668: [Bug]: Error on parsing nwb_version attribute
### What happened?

I'm in the process of updating our CI pipeline to validate the IPNWB/MIES generated NWB.
pynwb is updated from 2.1.0 to 2.3.1.

The file passes validation.

See also https://github.com/AllenInstitute/MIES/pull/1650.

Now the problem is that the file can not be opened anymore because `/nwb_version` has an unexpected HDF5 format. This was not an issue with 2.1.0. a1f472b1 (Check nwb_version on read (#1612), 2022-12-18) seems to be culprit.

```
HDF5 "E:\projekte\mies-igor\HardwareBasic_1-V2.nwb" {
ATTRIBUTE "nwb_version" {
   DATATYPE  H5T_STRING {
      STRSIZE 5;
      STRPAD H5T_STR_NULLPAD;
      CSET H5T_CSET_UTF8;
      CTYPE H5T_C_S1;
   }
   DATASPACE  SCALAR
   DATA {
   (0): "2.2.4"
   }
}
}
```

### Steps to Reproduce

```python
import pynwb

io = pynwb.NWBHDF5IO("HardwareBasic_1-V2.nwb", 'r', load_namespaces=True)
nwbfile = io.read()
```

### Traceback

```python
C:\users\thomas\Anaconda3\lib\site-packages\hdmf\spec\namespace.py:531: UserWarning: Ignoring cached namespace 'hdmf-common' version 1.1.3 because version 1.5.1 is already loaded.
  warn("Ignoring cached namespace '%s' version %s because version %s is already loaded."
C:\users\thomas\Anaconda3\lib\site-packages\hdmf\spec\namespace.py:531: UserWarning: Ignoring cached namespace 'core' version 2.2.4 because version 2.6.0-alpha is already loaded.
  warn("Ignoring cached namespace '%s' version %s because version %s is already loaded."
  File "read.py", line 4, in <module>
    nwbfile = io.read()
  File "C:\users\thomas\Anaconda3\lib\site-packages\hdmf\utils.py", line 645, in func_call
    return func(args[0], **pargs)
  File "C:\users\thomas\Anaconda3\lib\site-packages\pynwb\__init__.py", line 280, in read
    file_version_str, file_version = self.nwb_version
  File "C:\users\thomas\Anaconda3\lib\site-packages\pynwb\__init__.py", line 262, in nwb_version
    nwb_version_parts = nwb_version_string.replace("-", ".").replace("_", ".").split(".")
TypeError: a bytes-like object is required, not 'str'
```

### Operating System

Windows

### Python Executable

Conda

### Python Version

3.8

### Package Versions

[HardwareBasic_1-V2.zip](https://github.com/NeurodataWithoutBorders/pynwb/files/10996295/HardwareBasic_1-V2.zip)

[environment_for_issue.txt](https://github.com/NeurodataWithoutBorders/pynwb/files/10996306/environment_for_issue.txt)

### Code of Conduct

oruebel:
## Motivation

Fix #1668 

In #1612 we added support to do version checks on NWB HDF5 files to improve error reporting. The new ``nwb_version`` function implictly assumed that the attribute was written as a variable-lenght string. This leads to errors for NWB files that may use fixed-length strings instead (see #1668 for detail). This PR updated the ``NWBHDF5IO.new_version`` property to make sure the resulting bytes object is being decoded to a string in case that the attribute was written as a fixed-length bytes.

## Checklist

### DIFF
--- a/src/pynwb/__init__.py
+++ b/src/pynwb/__init__.py
@@ -259,7 +259,7 @@ class NWBHDF5IO(_HDF5IO):
         except KeyError:
             return None, None
         # Other system may have written nwb_version as a fixed-length string, resulting in a numpy.bytes_ object
-        # on read, rather than a variable length string. To address this, decode the bytes if necessary.
+        # on read, rather than a variable-length string. To address this, decode the bytes if necessary.
         if not isinstance(nwb_version_string, str):
             nwb_version_string = nwb_version_string.decode()
 
--- a/src/pynwb/__init__.py
+++ b/src/pynwb/__init__.py
@@ -258,6 +258,11 @@ class NWBHDF5IO(_HDF5IO):
         #  or when the HDF5 file is not a valid NWB file
         except KeyError:
             return None, None
+        # Other system may have written nwb_version as a fixed-length string, resulting in a numpy.bytes_ object
+        # on read, rather than a variable length string. To address this, decode the bytes if necessary.
+        if not isinstance(nwb_version_string, str):
+            nwb_version_string = nwb_version_string.decode()
+
         # Parse the version string
         nwb_version_parts = nwb_version_string.replace("-", ".").replace("_", ".").split(".")
         nwb_version = tuple([int(i) if i.isnumeric() else i
--- a/tests/integration/hdf5/test_io.py
+++ b/tests/integration/hdf5/test_io.py
@@ -443,7 +443,7 @@ class TestNWBHDF5IO(TestCase):
             del io.attrs['nwb_version']
         with NWBHDF5IO(self.path, 'r') as io:
             self.assertTupleEqual(io.nwb_version, (None, None))
-        # check that it works when setting the attribute to a fixed-legnth numpy-bytes string
+        # check that it works when setting the attribute to a fixed-length numpy-bytes string
         with File(self.path, mode='a') as io:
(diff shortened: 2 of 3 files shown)


tests/integration/hdf5/test_io.py
```diff
@@ -443,6 +443,11 @@ def test_nwb_version_property(self):
             del io.attrs['nwb_version']
         with NWBHDF5IO(self.path, 'r') as io:
             self.assertTupleEqual(io.nwb_version, (None, None))
+        # check that it works when setting the attribute to a fixed-legnth numpy-bytes string
```
rly:
```suggestion
        # check that it works when setting the attribute to a fixed-length numpy-bytes string
```

Huh. I wonder why codespell did not catch this.

src/pynwb/__init__.py
```diff
@@ -258,6 +258,11 @@ def nwb_version(self):
         #  or when the HDF5 file is not a valid NWB file
         except KeyError:
             return None, None
+        # Other system may have written nwb_version as a fixed-length string, resulting in a numpy.bytes_ object
+        # on read, rather than a variable length string. To address this, decode the bytes if necessary.
```
```suggestion
        # on read, rather than a variable-length string. To address this, decode the bytes if necessary.
```

rly (approved):

PR #1684, opened 2023-04-04
NWBHDF5IO ER and ER_Manager

mavaylon1:
## Motivation

The two changes is to add the ability to set an instance of ExternalResources for the NWBFile and NWBHDF5IO.
This is the first integration of ExternalResources into pynwb. The file is able to be linked to an instance of ExternalResources, while remaining separate during read/write.

## How to test the behavior?
```
```
There is a test for resources via test_resources.py. The NWBFile operates as it previously did with added functionality to support ExternalResources.

## Checklist


src/pynwb/__init__.py
```diff
@@ -289,7 +291,8 @@ def read(self, **kwargs):
         # read the file
         if self.external_resources is not None:
             file = super().read(**kwargs)
-            file.set_external_resources(self.external_resources)
+            er_read=ExternalResources.from_flat_tsv(path=self.external_resources)
+            file.link_resources(er_read)
```
oruebel: To simplify this a bit, I would suggest pulling the common element of reading the NWBFile out of the if/else block such that the if-block only needs to deal with the reading and linking of external resources

```
        # read the file
        file = super().read(**kwargs)
        if self.external_resources is not None:     
            er_read=ExternalResources.from_flat_tsv(path=self.external_resources)
            file.link_resources(er_read)
        return file
```

src/pynwb/__init__.py
```diff
@@ -302,11 +302,11 @@ def read(self, **kwargs):
...
         # read the file
+        file = super().read(**kwargs)
         if self.external_resources is not None:
-            file = super().read(**kwargs)
             er_read=ExternalResources.from_flat_tsv(path=self.external_resources)
             file.link_resources(er_read)
-            return file
+        return file
         else:
             return super().read(**kwargs)
```
oruebel:
```suggestion
```

I believe the else part should be removed here. It has the same effect as the ``return file`` statement and  since the return is before the else, the else part should never be executed.

src/pynwb/__init__.py
```diff
@@ -297,7 +303,11 @@ def read(self, **kwargs):
                 raise TypeError("NWB version %s not supported. PyNWB supports NWB files version 2 and above." %
                                 str(file_version_str))
         # read the file
-        return super().read(**kwargs)
+        file = super().read(**kwargs)
+        if self.external_resources is not None:
+            er_read = ExternalResources.from_flat_tsv(path=self.external_resources)
+            file.link_resources(er_read)
```
oruebel: I think the logic for using `external_resources` with the I/O should probably move to HDMF,  probably at the level of `HDMFIO`. We will need the same logic also in `NWBZarrIO` and the logic should be the same for `HDF5IO` and `ZarrIO`. Similarly, `external_resources` should then be a parameter on `HDMFIO.__init__`. @rly what do you think?

oruebel (commented):
Other than deciding whether adding external_resources should be moved to HDMF, this PR looks good to me.

src/pynwb/__init__.py
mavaylon1: I'll create a ticket on hdmf and move the I/O logic there.

mavaylon1:
@oruebel I wanted to go a bit further with the ER logic in (https://github.com/hdmf-dev/hdmf/pull/895) where the external_resources variable would be set to a read instance from the path, but this is not possible due HDMFIO and HDF5IO being imported in the hdmf/common init.py for namespace validation and will throw a circular import. As a result, the solution I came up with is that the variable is set to the path in HDMFIO and HDF5IO, but instantiated here. We need to merge the hdmf ticket, do a release, and change the minimum hdmf version here. Thoughts?

oruebel:
> namespace validation and will throw a circular import.

If I understand it correctly you only need the `ExternalResources` in the `read` method. You could just do a local import of `ExternalResources` inside the `read` method itself to avoid the circular import.

mavaylon1:
> > namespace validation and will throw a circular import.
> 
> If I understand it correctly you only need the `ExternalResources` in the `read` method. You could just do a local import of `ExternalResources` inside the `read` method itself to avoid the circular import.



> > namespace validation and will throw a circular import.
> 
> If I understand it correctly you only need the `ExternalResources` in the `read` method. You could just do a local import of `ExternalResources` inside the `read` method itself to avoid the circular import.



> > namespace validation and will throw a circular import.
> 
> If I understand it correctly you only need the `ExternalResources` in the `read` method. You could just do a local import of `ExternalResources` inside the `read` method itself to avoid the circular import.

Local import makes sense. Pending approval of the hdmf PR.

mavaylon1:
@oruebel can you do one last look through? Not much changed since after your approval, but I removed the logic for linking resources since that moved to hdmf, I updated the reqs to the release of hdmf 3.7, and I added an ignore to codecov for resources.py.

rly:
> I added an ignore to codecov for resources.py.

Why ignore coverage testing of this file?

mavaylon1:
There's nothing to test and coverage was flagging it.

Edit: created a test similar to test_core.py

rly (approved):

PR #1707, opened 2023-06-24
Enh mock nwbfile

bendichter:
## Motivation

fix #1705 

* add `nwbfile` arg to mock functions, which automatically adds the neurodata instance to the provided NWBFile object.
* add typehints to mock functions throughout.

to do:
* changelog
* tests

## How to test the behavior?
```
nwbfile = mock_NWBFile()
electrical_series = mock_ElectricalSeries(nwbfile=nwbfile)
```

## Checklist

- [ ] Did you update CHANGELOG.md with your changes?
- [ ] Have you checked our [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [ ] Have you ensured the PR clearly describes the problem and the solution?
- [ ] Is your contribution compliant with our coding style? This can be checked running `flake8` from the source directory.
- [ ] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/NeurodataWithoutBorders/pynwb/pulls) for the same change?
- [ ] Have you included the relevant issue number using "Fix #XXX" notation where XXX is the issue number? By including "Fix #XXX" you allow GitHub to close issue #XXX when the PR is merged.

PR #1709, opened 2023-06-26
Mock Update for Electrophys

mavaylon1:
## Motivation

Update Mock regarding Electrophys to support setting the NWBFile.

## How to test the behavior?
```
```

## Checklist


mavaylon1:
@oruebel As a first pass (haven't touched SpikeEvents), I am still fixed on the idea that there seems to be two ways to use the mock here. 
1. Sequentially: If as user starts with making a Device-->Group--Table then we would want the same nwbfile to be used throughout, hence the conditionals.
2. Start at Series and then update the parents for Group, Table, etc.

Either way I've implemented a series of conditionals that cover most cases. I could go deeper; for example, checking the device parent in mock_electrical series. This is used for testing so I doubt we need to be this thorough.

mavaylon1:
@bendichter I saw your PR #1707 and I think we could do this in stages. Merge yours to get something out there but then use this one to check the nwbfile more thoroughly. Take a look here for an example for just the elctrophys and let me know what you think.

mavaylon1:
@oruebel @bendichter any idea on my idea for merging Ben's work as a quick fix and then me expanding something more thorough after the merge? Sorry this got lost in the weeds.

oruebel:
> any idea on my idea for merging Ben's work as a quick fix and then me expanding something more thorough after the merge

I would chat with @bendichter since they are using the mock classes most heavily right now. From my end that sounds good.

PR #1737, opened 2023-07-25
Add methods to find objects of a given neurodata type / pynwb class

### LINKED ISSUE #560: Programmatically find all the time series in any NWB file
Hello everyone.
I've been using your API to develop my [GSoC project](https://summerofcode.withgoogle.com/projects/#6057925982814208) and one of my tasks was to find a way to programmatically find all the time series present in any NWB 2.x file.
I've tried to develop this as follows:
``` python
def _get_data_interfaces(self, node):
        """Given a NWBHDF5IO returns all the data_interfaces objects presents on it."""
        data_interfaces_list = []
        for child in node.children:
            if isinstance(child, NWBDataInterface):
                data_interfaces_list.append(child)
            data_interfaces_list += self._get_data_interfaces(child)
       return data_interfaces_list

def _get_timeseries(self):
        """Given all the nwb_data_interfaces returns all of those that are timeseries objects."""
        time_series_list = []
        for data_interface in self.nwb_data_interfaces_list:
            if isinstance(data_interface, TimeSeries):
                time_series_list.append(data_interface)
return time_series_list
```
I believe this returns all the [TimeSeries objects](https://pynwb.readthedocs.io/en/latest/overview_nwbfile.html#timeseries), then I simply run over each one of them to get mono dimensional time series easily plottable:
``` python
def get_mono_dimensional_timeseries(self, values):
        """Given a timeseries object returns all mono dimensional timeseries presents on it."""
        mono_time_series_list = []
        if isinstance(values, collections.Iterable):
            try:
                data = [float(i) for i in values]
                mono_time_series_list.append(data)
            except:
                for inner_list in values:
                    mono_time_series_list += self.get_mono_dimensional_timeseries(inner_list)
return mono_time_series_list
```

I would like to get your opinion on this. 
I wasn't able to find this feature in your API. Do you think it has any value? Would you do this any other way? 
I understand that this is not a typical issue, so apologies in advance if it should be asked elsewhere. :pray:

rly:
## Motivation

A common feature request is to find objects of a given neurodata type or pynwb class. 

For example:
- get all objects with the neurodata type "EcephysSpecimen" from the namespace "ndx-aibs-ecephys"
- get all objects with the neurodata type "Subject" from the namespace "core" including subclasses/subtypes such as "EcephysSpecimen" from the namespace "ndx-aibs-ecephys"
- get all objects that are instances of a PyNWB class such as `pynwb.file.Subject`

Fix #560 

TODO
- [ ] Make tests

## Checklist


src/pynwb/__init__.py
```diff
@@ -349,6 +349,32 @@ def export(self, **kwargs):
...
+
+            from pynwb import NWBHDF5IO
+            with NWBHDF5IO(filepath, mode="r", load_namespaces=True) as io:
+                obj_list = io.get_neurodata_type_objs("Subject", "core")
+
+        """
+        pynwb_cls = self.manager.type_map.get_dt_container_cls(neurodata_type, namespace)
+
+        read_nwbfile = self.read()
+        return read_nwbfile.find_all_of_class(pynwb_cls)
```
oruebel: Could you clarify what use-case this function addresses that `NWBFile.find_all_of_class` does not?

src/pynwb/__init__.py
```diff
@@ -349,6 +349,32 @@ def export(self, **kwargs):
...
 
+    @docval({'name': 'neurodata_type', 'type': str, 'doc': 'The neurodata type to search for.'},
+            {'name': 'namespace', 'type': str, 'doc': 'The namespace of the neurodata type to search for.'},)
+    def get_neurodata_type_objs(self, neurodata_type: str, namespace: str):
+        """Return all PyNWB objects in the file that have a neurodata type from a namespace.
+
+        This works regardless of whether the extension was imported earlier in the python execution.
+
+        All objects that are instances of the class associated with the given neurodata_type in the
+        given namespace will be returned. This includes objects that are instances of a subclass.
```
```suggestion
        This method is useful for getting neurodata type objects from cached extensions where you
        do not have easy access to a python class to pass to `NWBFile.find_all_of_class`.

        All objects that are instances of the class associated with the given neurodata_type in the
        given namespace will be returned. This includes objects that are instances of a subclass.
```

src/pynwb/__init__.py
rly: I added a comment in the docstring. It says: This method is useful for getting neurodata type objects from cached extensions where you do not have easy access to a python class to pass to `NWBFile.find_all_of_class`.

oruebel: Makes sense. But it seems a bit strange that this functionality is in two different places. Also, with this we would need to duplicate this also for Zarr. Instead, I'm wondering whether `find_all_of_class` can be modified to do both, something like:

```
@docval({'name': 'neurodata_types', 'type': (type, str), 'doc': 'The PyNWB container class to search for instances of or the string with the name of the neurodata_type'},)
def find_all(self, neurodata_type):
     if isinstance(neurodata_type, str):
         pynwb_cls = self.io.manager.type_map.get_dt_container_cls(neurodata_type, namespace)
     else:
        pynwb_cls = neurodata_type
     ret = [obj for obj in self.objects.values() if isinstance(obj, pynwb_class)]
     return ret
```

rly: Yeah, that works too. I'll update it

CodyCBakerPhD:
Idea LGTM, just needs a quick test or two

bendichter:
This is a nice idea. I have two suggestions:

1. I'm not crazy about the name. I would prefer something a bit more explicit, like `find_all_of_type`
2. I would rather this live in `Container` or maybe `NWBContainer`. That way it could be used on any container, not just `NWBFile`.

bendichter (changes requested):
This is a nice idea. I have two suggestions:

1. I'm not crazy about the name. I would prefer something a bit more explicit, like `find_all_of_type`
2. I would rather this live in `Container` or maybe `NWBContainer`. That way it could be used on any container, not just `NWBFile`.

And yes, I of course agree with @CodyCBakerPhD that this will need tests

oruebel:
I'm wondering whether it would be simpler from a user perspective to have this part of the existing `all_children` method. I.e., we could add the query parameters for the types (with default value of `None`) to `all_children`. If the query parameters are set to `None` then `all_children` behaves as is and returns all the children and if the parameters are set then it would use the new logic to filter for types. 

https://github.com/NeurodataWithoutBorders/pynwb/blob/46e8878b9d9bd908dd3cd0715769ce08329531d0/src/pynwb/file.py#L530

PR #1748, opened 2023-07-28
default load_namespaces=True

### LINKED ISSUE #1143: Automagically load necessary Python modules for extensions registered in nwb-extensions
It might be that it is already intended, but found no issue yet, thus decided to be the one to "report" since I found no mentioning of nwb-extensions among issues or in the code.

ATM, `NWBHDF5IO(...).read()` would fail with an exception like `'AIBS_ecephys' not a namespace"` leaving user baffled on "what have I done wrong?" or "how do I fix that?".  If registry (I am still not sure if it captures all necessary information ATM) had provided listing of those "name spaces" and (Python) modules which need to be installed, then PyNWB would automagically `import allensdk.brain_observatory.ecephys.nwb` while `.read()`ing that file. And if `import` fails, it would provide an error message like
```
The file ... contains data of AIBS_ecephys neural data type, to load which PyNWB requires allensdk (>= X.Y.Z) to be installed.
```
Without such functionality, IMHO it would be pretty much impossible to develop generic PyNWB-based tools otherwise agnostic of underlying neural data types.

edit: in https://github.com/dandi/dandi-cli/pull/36 I for now provide a prototypical workaround only for AIBS_ecephys for now

bendichter:
Fix https://github.com/NeurodataWithoutBorders/pynwb/issues/1143

`test_outer_import_structure` tests that the imports available from the `pynwb` module are unchanged. This PR removes the `warn` library, so it has also been removed form this test. This will technically break backwards compatibility if a user is running `from pynwb import warn`, but that would be highly unusual and unlikely.


bendichter:
I don't understand why I am failing flake8. @rly , can you help?

oruebel:
> I don't understand why I am failing flake8.

The GitHub Action indicated the following flake8 issues:

```
./tests/integration/hdf5/test_ecephys.py:3:109: E231 missing whitespace after ','
./tests/unit/test_ecephys.py:5:104: E231 missing whitespace after ','
```

bendichter:
oh, I see. It's the imports

rly (dismissed):

rly:
Currently, calling `load_namespaces=True` is resulting in lots of this warning:
`UserWarning: No cached namespaces found in [...].nwb`

This comes from hdmf. I think we can remove that warning. @bendichter could you make a PR for this?

oruebel:
Fix #1143

tests/unit/test_file.py
```diff
@@ -563,9 +563,8 @@ def test_simple(self):
         with NWBHDF5IO(self.path, 'w') as io:
             io.write(nwbfile, cache_spec=False)
 
-        with self.assertWarnsWith(UserWarning, "No cached namespaces found in %s" % self.path):
-            with NWBHDF5IO(self.path, 'r', load_namespaces=True) as reader:
-                nwbfile = reader.read()
+        with NWBHDF5IO(self.path, 'r') as reader:
+            nwbfile = reader.read()
```
oruebel: What is this case testing, it does not seem to assert anything? Has the warning been removed when no namespace has been cached?

oruebel:
@bendichter just FYI, there were a few conflicts in particular in the streaming tutorial, when I updated the PR with the dev branch. Here the changes I made during the merge:

https://github.com/NeurodataWithoutBorders/pynwb/pull/1748/commits/e7023c23c65d460ed82cf9715abc2a958b699234

oruebel:
The "coverage / ubuntu-latest" action seems to show a large number of warnings and one test failure that appear to be related to the changes in this PR (see  https://github.com/NeurodataWithoutBorders/pynwb/actions/runs/6098358982/job/16548020150?pr=1748)

In particular

```
FAIL: test_validate_file_cached_ignore (validation.test_validate.TestValidateCLI.test_validate_file_cached_ignore)
Test that validating a file with cached spec against the core namespace succeeds.
  File "/home/runner/work/pynwb/pynwb/tests/validation/test_validate.py", line 139, in test_validate_file_cached_ignore
    self.assertEqual(result.stderr.decode('utf-8'), '')
AssertionError: '/opt/hostedtoolcache/Python/3.11.4/x64/li[552 chars]."\n' != ''
- /opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/hdmf/spec/namespace.py:531: UserWarning: Ignoring cached namespace 'hdmf-common' version 1.0.0 because version 1.8.0 is already loaded.
-   warn("Ignoring cached namespace '%s' version %s because version %s is already loaded."
- /opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/hdmf/spec/namespace.py:531: UserWarning: Ignoring cached namespace 'core' version 2.1.0 because version 2.6.0-alpha is already loaded.
-   warn("Ignoring cached namespace '%s' version %s because version %s is already loaded."
- 
```

Additionally, there are a lot of warnings where either no cached namespaces are found or another version of the same namespace has already been loaded: 

```
/opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/hdmf/backends/hdf5/h5tools.py:213: UserWarning: No cached namespaces found in test_RoiResponseSeries.nwb
```
```
/opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/hdmf/spec/namespace.py:531: UserWarning: Ignoring cached namespace 'core' version 2.1.0 because version 2.6.0-alpha is already loaded.
```

tests/unit/test_file.py
bendichter: @oruebel you're right, I'll remove this test

rly (dismissed):

docs/gallery/advanced_io/linking_data.py
```diff
@@ -6,57 +6,50 @@
...
+ .. tip::
 
-####################
-# Creating test data
-# ---------------------------
-#
-# In the following we are creating two :py:meth:`~pynwb.base.TimeSeries` each written to a separate file.
-# We then show how we can integrate these files into a single NWBFile.
+    In the case of :py:meth:`~pynwb.base.TimeSeries`, the uncorrected timestamps generated by the acquisition
+    system can be stored (or linked) in the *sync* group. In the NWB:N format, hardware-recorded time data
```
bendichter:
```suggestion
    system can be stored (or linked) in the *sync* group. In the NWB format, hardware-recorded time data
```

docs/gallery/advanced_io/linking_data.py
```diff
@@ -6,57 +6,50 @@
...
-# afterwards link the data into a single NWB:N file. In this case, each recording becomes represented by a
-# separate file-system object that can be set as read-only once the experiment is done.  In the following
-# we are using :py:meth:`~pynwb.base.TimeSeries` as an example, but the same approach works for other
-# NWBContainers as well.
-#
+NBWContainer classes (e.g., :py:class:`~pynwb.base.TimeSeries`) support the integration of data stored in external
+HDF5 files with NWB data files via external links. To make things more concrete, let's look at the following use
+case. We want to simultaneously record multiple data streams during data acquisition. Using the concept of external
+links allows us to save each data stream to an external HDF5 files during data acquisition and to
+afterwards link the data into a single NWB:N file. In this case, each recording becomes represented by a
```
bendichter:
```suggestion
afterwards link the data into a single NWB file. In this case, each recording becomes represented by a
```

bendichter:
@rly it looks like you are altering the NWBHDF5IO call to ignore the "no namespace found error". I guess this wouldn't really come up in practice since the NWB files are generally stored with cached schema. Is that right?

rly:
> @rly it looks like you are altering the NWBHDF5IO call to ignore the "no namespace found error". I guess this wouldn't really come up in practice since the NWB files are generally stored with cached schema. Is that right?

That was done in https://github.com/hdmf-dev/hdmf/pull/926 . But yes, that warning wouldn't really come up in practice except when reading old files where the schema was not cached by default. For those old files without the core/hdmf-common schema cached, even if the schema were cached, they would be ignored because the schema loaded by pynwb would take precedence. So it's not a useful warning.

rly (approved):

rly:
@bendichter can we merge this?

src/pynwb/__init__.py
```diff
@@ -261,15 +261,10 @@ def __init__(self, **kwargs):
...
         # Define the BuildManager to use
-        if load_namespaces:
-            if manager is not None:
-                warn("loading namespaces from file - ignoring 'manager'")
-            if extensions is not None:
-                warn("loading namespaces from file - ignoring 'extensions' argument")
-            # namespaces are not loaded when creating an NWBHDF5IO object in write mode
-            if 'w' in mode or mode == 'x':
-                raise ValueError("cannot load namespaces from file when writing to it")
+        if mode in 'wx' or manager is not None or extensions is not None:
```
h-mayorquin, line 264: I think this introduced a bug.

src/pynwb/__init__.py
```diff
@@ -261,15 +261,10 @@ def __init__(self, **kwargs):
             popargs('path', 'mode', 'manager', 'extensions', 'load_namespaces',
                     'file', 'comm', 'driver', 'herd_path', kwargs)
         # Define the BuildManager to use
-        if load_namespaces:
-            if manager is not None:
-                warn("loading namespaces from file - ignoring 'manager'")
-            if extensions is not None:
-                warn("loading namespaces from file - ignoring 'extensions' argument")
-            # namespaces are not loaded when creating an NWBHDF5IO object in write mode
-            if 'w' in mode or mode == 'x':
```
h-mayorquin, line 270: Here was correct.

PR #1755, opened 2023-08-11
codespell2: now ipynwb notebooks too!

yarikoptic:


rly:
@yarikoptic Should https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/.github/workflows/codespell.yml#L19C53-L19C53 be updated to v2?

rly (dismissed):

yarikoptic:
it could, pushed commit for that although doesn't per se relates to these changes AFAIK ;-)

rly (approved):

PR #1760, opened 2023-08-17
refactor streaming tutorial to expose Python code

bendichter:


rly (approved):

rly:
Huh. We obviously thought about this when we (I?) added this code:
https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/test.py#L80

But didn't realize it was not running as code. Good catch.

docs/gallery/advanced_io/streaming.py
```diff
@@ -23,113 +23,111 @@
...
+filepath = 'sub-anm372795/sub-anm372795_ses-20170718.nwb'  # 450 kB file
+with DandiAPIClient() as client:
+    asset = client.get_dandiset(dandiset_id, 'draft').get_asset_by_path(filepath)
+    s3_url = asset.get_content_url(follow_redirects=1, strip_query=True)
+
+##############################################
+# Streaming Method 1: fsspec
+# --------------------------
+# fsspec is another data streaming approach that is quite flexible and has several performance advantages. This library
+# creates a virtual filesystem for remote stores. With this approach, a virtual file is created for the file and
```
oruebel, line 42:
```suggestion
# The ffspec library provides a flexible and easy-to-use mechanism for creating a virtual filesystem for 
# remote stores and streaming remote NWB files.  With this approach, a virtual file is created for the file and
```

The main reason for the suggestion is that I found the "is another data streaming approach" confusing, since this is the first approach that is being listed.

docs/gallery/advanced_io/streaming.py
```diff
@@ -23,113 +23,111 @@
...
+# driver created by the HDF5 Group that allows HDF5 to read HDF5 files stored remotely in s3 buckets. Using this method
+# requires that your HDF5 library is installed with the ROS3 driver enabled. This is not the default configuration,
+# so you will need to make sure you install the right version of ``h5py`` that has this advanced configuration enabled.
+# You can install HDF5 with the ROS3 driver from `conda-forge <https://conda-forge.org/>`_ using ``conda``. You may
+# first need to uninstall a currently installed version of ``h5py``.
+#
+# .. code-block:: bash
+#
+#    pip uninstall h5py
+#    conda install -c conda-forge "h5py>=3.2"
```
oruebel, line 105: Since this has been in h5py now for 2 years, could this be shortened to just say something along the lines of "ROS3 requires a recent version of h5py >= 3.2"

oruebel: Or is installing h5py with ROS3 enabled still and issue?

docs/gallery/advanced_io/streaming.py
```diff
@@ -23,113 +23,111 @@
...
+#
+# 1. fsspec is easier to install
+# 2. fsspec supports caching, which will dramatically speed up repeated requests for the
+#    same region of data
+# 3. fsspec automatically retries when s3 fails to return.
+# 4. fsspec works with other storage backends and
+# 5. fsspec works with other types of files.
+# 6. In our hands, fsspec is faster out-of-the-box.
+#
+# For these reasons, we would recommend use fsspec for most Python users.
```
oruebel, line 133:
```suggestion
#
# From a user perspective, the :py:class:`~pynwb.file.NWBFile` once opened works the same with 
# both fsspec and ros3.  However, in general, we currently recommend using fsspec for streaming
# NWB files because it:
#
# 1. supports caching, which will dramatically speed up repeated requests for the
#    same region of data,
# 2. automatically retries when s3 fails to return, which helps avoid errors when accessing data due to
#     intermittent errors in connections with S3,
# 3. works also with other storage backends (e.g., GoogleDrive or Dropbox, not just S3) and file formats, and
# 4. in our experience appears to provide faster out-of-the-box performance than the ros3 driver.
# 
#
```

PR #1766, opened 2023-08-23
Add missing TimeSeriesMap.data_attr function to link data between timeries

### LINKED ISSUE #1765: [Bug]: Reusing data across TimeSeries objects does not seem to work
### What happened?

PyNWB documentation & code seem to suggest that, similar to sharing timestamps between TimeSeries, we can also share data between TimeSeries. Passing a TimeSeries object to the data field of another TimeSeries does not error, but when reading the file back, the data fields don't appear to be linked, though I'd expect that they would be like the timestamps.

### Steps to Reproduce

```python
import numpy as np
from pynwb import NWBHDF5IO, NWBFile, TimeSeries
from datetime import datetime
from uuid import uuid4

nwbfile = NWBFile(
    session_description="not a real session",
    identifier=str(uuid4()),
    session_start_time=datetime.now(),
)

timeseries1 = TimeSeries(
    name="timeseries1",
    data=np.random.rand(40),
    unit="n/a",
    timestamps=(np.arange(40) * 1e-2),
)

timeseries2 = TimeSeries(
    name="timeseries2",
    data=timeseries1,
    unit="n/a",
    timestamps=timeseries1,
)

nwbfile.add_acquisition(timeseries1)
nwbfile.add_acquisition(timeseries2)

with NWBHDF5IO('temp.nwb', 'w') as io:
    io.write(nwbfile)

io = NWBHDF5IO('temp.nwb', 'r')
read_nwbfile = io.read()
print(nwbfile.acquisition['timeseries1'].timestamps == nwbfile.acquisition['timeseries2'].timestamps)
print(nwbfile.acquisition['timeseries1'].data == nwbfile.acquisition['timeseries2'].data)
print(nwbfile.acquisition['timeseries1'])
print(nwbfile.acquisition['timeseries2'])
```

### Traceback

```python
True
False
timeseries1 pynwb.base.TimeSeries at 0x140171917956080
Fields:
  comments: no comments
  conversion: 1.0
  data: <HDF5 dataset "data": shape (40,), type "<f8">
  description: no description
  interval: 1
  offset: 0.0
  resolution: -1.0
  timestamp_link: (
    timeseries2 <class 'pynwb.base.TimeSeries'>
  )
  timestamps: <HDF5 dataset "timestamps": shape (40,), type "<f8">
  timestamps_unit: seconds
  unit: n/a
timeseries2 pynwb.base.TimeSeries at 0x140171917958048
Fields:
  comments: no comments
  conversion: 1.0
  data: <HDF5 dataset "data": shape (40,), type "<f8">
  description: no description
  interval: 1
  offset: 0.0
  resolution: -1.0
  timestamps: timeseries1 pynwb.base.TimeSeries at 0x140171917956080
Fields:
  comments: no comments
  conversion: 1.0
  data: <HDF5 dataset "data": shape (40,), type "<f8">
  description: no description
  interval: 1
  offset: 0.0
  resolution: -1.0
  timestamp_link: (
    timeseries2 <class 'pynwb.base.TimeSeries'>
  )
  timestamps: <HDF5 dataset "timestamps": shape (40,), type "<f8">
  timestamps_unit: seconds
  unit: n/a

  timestamps_unit: seconds
  unit: n/a
```

### Operating System

Linux

### Python Executable

Python

### Python Version

3.10

### Package Versions

(run on DANDI hub)
[issue_environment.txt](https://github.com/NeurodataWithoutBorders/pynwb/files/12420538/issue_environment.txt)

### Code of Conduct

oruebel:
## Motivation

Fix #1765

`TimeSeries.data` did not create SoftLink when set to another `TimeSeries`. It looks like the cause is that the `TimeSeriesMap` ObjectMapper does not define the corresponding `TimeSeriesMap.data_attr` (as is the case for `timestamps`) so that the link was not created. This PR adds this method.

- [x] Fix `TimeSeriesMap`
- [x] Add integration test for linking `TimeSeries.data`
- [x] Check and update existing tests 
- [x] Update CHANGELOG

Creating a SoftLink for `TimeSeries.data` triggered a secondary error in `TimeSeriesMap.unit_carg` where we may not get a `TimeSeries` instead of a `DatasetBuilder` to look up the unit. 

## How to test the behavior?
See example script in #1765

## Checklist

- [X] Did you update CHANGELOG.md with your changes?


oruebel:
Creating external links with the logic of setting `TimeSeries(data=timeseries,...)` and `TimeSeries(timestamps=timeseries, ...)` appears to be broken. Creating an external link with `HDF5IO` works:

https://github.com/NeurodataWithoutBorders/pynwb/blob/68c4f5645d00f3958471e5e2bc360c90f49a4740/tests/integration/hdf5/test_modular_storage.py#L82-L84

But using the approach with `TimeSeries(data=timeseries, ...)` 

https://github.com/NeurodataWithoutBorders/pynwb/blob/68c4f5645d00f3958471e5e2bc360c90f49a4740/tests/integration/hdf5/test_modular_storage.py#L81

results in a broken external link:

```
  ...hdmf/backends/hdf5/h5tools.py:641: BrokenLinkWarning: Path to Group altered/broken at /acquisition/test_mod_ts/timestamps
    warnings.warn('Path to Group altered/broken at ' + os.path.join(h5obj.name, k), BrokenLinkWarning)
```

This issues appears to be a separate issue that I think we simply didn't see previously because of the bug from #1765 hiding the failure in the test. I'm not sure yet what the fix for this issue is, but I think it is likely either part of `TimeSeriesMap.data_attr`  and `TimeSeriesMap.timestamps_attr` or a bug in `HDF5IO.write_link`

@rly do you have an idea what might be wrong here? The test that triggers this issue is `pytest tests/integration/hdf5/test_modular_storage.py`

oruebel:
Looking at the generated file, it appears that the path to the external file being set correctly but the path to the object is not being set correctly when using the `TimeSeries(data=timeseries,...)` approach:

```
data                     External Link {test_time_series_modular_data.nwb//data}
timestamps               External Link {test_time_series_modular_data.nwb//acquisition/data_ts/timestamps}
```

I.e. it should be ` External Link {test_time_series_modular_data.nwb//acquisition/data_ts/data}` instead.

CodyCBakerPhD:
To be clear, our only desire w.r.t. this feature are internal linkages. We have two TimeSeries internal to the same file that share the same `data` (and don't want to duplicate it) but different time bases, so this approach seemed ideal for us

Sounds like the external link thing might be a separate issue for a separate PR

oruebel:
> Sounds like the external link thing might be a separate issue for a separate PR

Yes, I believe that is correct. We could update the test to not use the problematic logic and file a separate issue.

oruebel:
> Sounds like the external link thing might be a separate issue for a separate PR

I've create #1767 as a separate issue. To get the test-suite to pass I have further:
* Updated `TestTimeSeriesModular` to only use HDF5IO to create external links instead of setting `TimeSeries(data=timeseries, ...)`
* Added `TestTimeSeriesModularLinkViaTimeSeries` to explicitly test the behavior with `TimeSeries(data=timeseries, timeseries=timeseries)`. This test is currently being skipped and should be addressed in #1767

rly:
Note to self: add test to make sure dataset is coming from the expected file

oruebel:
> Note to self: add test to make sure dataset is coming from the expected file

@rly I updated the test in `test_modular_storage.py` to check for this here:

https://github.com/NeurodataWithoutBorders/pynwb/blob/36000795fe46287cec7407f49fd53719cb47a6fc/tests/integration/hdf5/test_modular_storage.py#L138-L147

I also updated the test to make it more readable. The test was getting confusing because it uses 2 different files and it wasn't always clear from the names of the variables which file the variables belonged to and what was actually being tested by the asserts.

oruebel:
I accidentally marked #1767  as fixed in one of the commits because I thought I had found the issue and the problem was actually an issue with the test. But it turns out I had made and error in that commit and the problem still remains. I fixed the tests but #1767  should remain open for now.

oruebel:
@rly just to summarize the status of this:

* This PR fixes the issues of creating links within the same file by setting `TimeSeries(data=other_timeseries)` and is ready for review and I believe could be merged as is
*  https://github.com/NeurodataWithoutBorders/pynwb/issues/1767 describes the issue with linking to external files via `TimeSeries(data=other_timeseries)`. This PR adds `TestTimeSeriesModularLinkViaTimeSeries` to explicitly test this issue but the test is currently being skipped and should be addressed either separately or (if it is a quick fix) in this PR
    * Note: If we don't fix #1767 in this PR then we should reopen that issue after merging this PR because I accidentally declared the issue as fixed in one of the commits in this PR

rly (dismissed):

rly:
Looks good to me. Thanks for the thorough comments in the tests too.

rly (approved):

PR #1770, opened 2023-08-29
Fix bad validation check for postion in ElectrodeGroup.__init__

### LINKED ISSUE #1768: [Bug]: Writing ElectrodeGroup position saves incorrectly modified data to file
### What happened?

There seem to be some inconsistencies in writing ElectrodeGroup position to a file depending on the input type:
* For `np.ndarray` inputs, the `__init__` function checks whether position was provided using `if position` instead of `if position is not None`, which leads to the usual `The truth value of an array with more than one element is ambiguous` error
* For tuple inputs, the saved `position` data seems to be duplicated into an array of tuples like `[(x, y, z), (x, y, z), (x, y, z)]`
* For list inputs, the saved `position` data seems to be duplicated into a array of tuples like `[(x, x, x), (y, y, y), (z, z, z)]`

These last two things only happen on the write/read round trip, not before writing the file.

### Steps to Reproduce

```python
import numpy as np
import os
from uuid import uuid4
from datetime import datetime
​
nwbfile = NWBFile(
    session_description="no description.",
    session_start_time=datetime.now(),
    identifier=str(uuid4()),
)
​
device = nwbfile.create_device(name="Device", description="no description.")
​
electrode_group = nwbfile.create_electrode_group(
    name="ElectrodeGroup",
    description="no description.",
    device=device,
    location="unknown",
    # position=np.array([1., 2., 3.]),
    position=(1., 2., 3.),
    # position=[1., 2., 3.],
)
​
with NWBHDF5IO('test.nwb', 'w') as io:
    io.write(nwbfile)
    
with NWBHDF5IO('test.nwb', 'r') as io:
    read_nwbfile = io.read()
    
    print(read_nwbfile.electrode_groups['ElectrodeGroup'])
    print(read_nwbfile.electrode_groups['ElectrodeGroup'].position[()])
```

### Traceback

```python
# for position = np.array([1., 2., 3.])
  File "/home/jovyan/scripts/test_write_electrode_group_position.py", line 16, in <module>
    electrode_group = nwbfile.create_electrode_group(
  File "/opt/conda/lib/python3.10/site-packages/hdmf/utils.py", line 644, in func_call
    return func(args[0], **pargs)
  File "/opt/conda/lib/python3.10/site-packages/hdmf/container.py", line 1044, in _func
    ret = container_type(**kwargs)
  File "/opt/conda/lib/python3.10/site-packages/hdmf/utils.py", line 644, in func_call
    return func(args[0], **pargs)
  File "/opt/conda/lib/python3.10/site-packages/pynwb/ecephys.py", line 33, in __init__
    if args_to_set['position'] and len(args_to_set['position']) != 3:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

# for position=(1., 2., 3.)
ElectrodeGroup pynwb.ecephys.ElectrodeGroup at 0x140662980868848
Fields:
  description: no description.
  device: Device pynwb.device.Device at 0x140662980868752
Fields:
  description: no description.

  location: unknown
  position: <hdmf.backends.hdf5.h5_utils.ContainerH5TableDataset object at 0x7feea701fb50>

[(1., 2., 3.) (1., 2., 3.) (1., 2., 3.)]

# for position=[1., 2., 3.]
ElectrodeGroup pynwb.ecephys.ElectrodeGroup at 0x140659597409008
Fields:
  description: no description.
  device: Device pynwb.device.Device at 0x140659597408912
Fields:
  description: no description.

  location: unknown
  position: <hdmf.backends.hdf5.h5_utils.ContainerH5TableDataset object at 0x7feddd567b50>

[(1., 1., 1.) (2., 2., 2.) (3., 3., 3.)]
```

### Operating System

Linux

### Python Executable

Python

### Python Version

3.10

### Package Versions

[issue_environment.txt](https://github.com/NeurodataWithoutBorders/pynwb/files/12457610/issue_environment.txt)

### Code of Conduct

oruebel:
## Motivation

Fix #1768   `ElectrodeGroup.position` is a compound dataset of `x,y,z` positions, but `ElectrodeGroup.__init__` validates it wrongly by enforcing that the array has 3 elements instead of checking that the elements have 3 elements. This PR updates  `ElectrodeGroup.__init__` to validate `ElectrodeGroup.position` correctly

## How to test the behavior?

See the linked issue for details

## Checklist


src/pynwb/ecephys.py
```diff
@@ -26,13 +26,29 @@ class ElectrodeGroup(NWBContainer):
...
-                             % str(args_to_set['position']))
+        # position is a compound dataset, i.e., this must be an array with a compound data type of three floats
+        # or each element in the list must at least have three entries
+        if args_to_set['position'] is not None and len(args_to_set['position']) > 0:
+            position_dtype_valid = True
+            if hasattr(args_to_set['position'], 'dtype'):  # If we have a dtype, then check that it is valid
+                if len(args_to_set['position'].dtype) != 3:
+                    print("H1", len(args_to_set['position'].dtype))
+                    position_dtype_valid = False
+            if position_dtype_valid:  # If we have list of element, then check that the elements are of lenght 3
```
oruebel:
```suggestion
            if position_dtype_valid:  # If we have list of element, then check that the elements are of length 3
```

oruebel:
Just some detective work for background:

*  It seems that `ElectrodeGroup.position` was added in NWB v2.2 in #1146. I believe this bug has existed since the field was added.
* Looking at https://github.com/NeurodataWithoutBorders/nwb-schema/commit/59db28ee96cfe63f7304789cc879cd94a75632ce it seems that `ElectrodeGroup.position` was first added as an array of length 3 (which is what the current check is looking for) and then changed to a compound data type. I.e., my guess is that PyNWB didn't get updated when the change to compound happened, and the unit tests did not fully cover this field. 

Given that it has take 3 years for this issue to surface, it seems that `ElectrodeGroup.position` is probably rarely used.

src/pynwb/ecephys.py
```diff
@@ -26,13 +26,29 @@ class ElectrodeGroup(NWBContainer):
...
-        if args_to_set['position'] and len(args_to_set['position']) != 3:
-            raise ValueError('ElectrodeGroup position argument must have three elements: x, y, z, but received: %s'
-                             % str(args_to_set['position']))
+        # position is a compound dataset, i.e., this must be an array with a compound data type of three floats
+        # or each element in the list must at least have three entries
+        if args_to_set['position'] is not None and len(args_to_set['position']) > 0:
+            position_dtype_valid = True
+            if hasattr(args_to_set['position'], 'dtype'):  # If we have a dtype, then check that it is valid
+                if len(args_to_set['position'].dtype) != 3:
+                    print("H1", len(args_to_set['position'].dtype))
```
CodyCBakerPhD: Should this print statement still be here?

```suggestion
```

src/pynwb/ecephys.py
```diff
@@ -26,13 +26,29 @@ class ElectrodeGroup(NWBContainer):
...
         super().__init__(**kwargs)
-        if args_to_set['position'] and len(args_to_set['position']) != 3:
-            raise ValueError('ElectrodeGroup position argument must have three elements: x, y, z, but received: %s'
-                             % str(args_to_set['position']))
+        # position is a compound dataset, i.e., this must be an array with a compound data type of three floats
+        # or each element in the list must at least have three entries
+        if args_to_set['position'] is not None and len(args_to_set['position']) > 0:
+            position_dtype_valid = True
+            if hasattr(args_to_set['position'], 'dtype'):  # If we have a dtype, then check that it is valid
+                if len(args_to_set['position'].dtype) != 3:
```
CodyCBakerPhD: Not sure what the `len` of a `.dtype` here means; is it for structured arrays? (one type per column?)

CodyCBakerPhD: The comments above explain a little bit...

Would this be because of pre-wrapping of the user input by docval before we reach this point?

tests/unit/test_ecephys.py
```diff
@@ -173,7 +180,29 @@ def test_init_position_none(self):
...
+                           description='electrode description',
+                           location='electrode location',
+                           device=dev1,
+                           position=[(1, 2), ])
+        with self.assertRaises(ValueError):
+            ElectrodeGroup(name='elec1',
+                           description='electrode description',
+                           location='electrode location',
+                           device=dev1,
+                           position=np.array([(1., 2.)], dtype=np.dtype([('x', float), ('y', float)])))
```
CodyCBakerPhD: I see, this test makes the code clearer

tests/integration/hdf5/test_ecephys.py
```diff
@@ -25,7 +25,8 @@ def setUpContainer(self):
         eg = ElectrodeGroup(name='elec1',
                             description='a test ElectrodeGroup',
                             location='a nonexistent place',
-                            device=self.dev1)
+                            device=self.dev1,
+                            position=[(1., 2., 3.), ])
```
CodyCBakerPhD: One question - why is this a list at all? Is there a case where we would have multiple tuples in this list? What would that represent?

src/pynwb/ecephys.py
```diff
@@ -26,13 +26,29 @@ class ElectrodeGroup(NWBContainer):
...
+                    print("H1", len(args_to_set['position'].dtype))
+                    position_dtype_valid = False
+            if position_dtype_valid:  # If we have list of element, then check that the elements are of length 3
+                try:
+                    if len(args_to_set['position'][0]) != 3:
```
CodyCBakerPhD: Checking only the first element here seems to imply that  this value should only be a single tuple of length 2 or 3 and that we would not expect more than one to be passed in the list input

CodyCBakerPhD:
@oruebel Any updates on this PR?

src/pynwb/ecephys.py
oruebel: We could iterate through all elements. I believe I just wanted to avoid checking all elements to keep runtime short in case there are a lot of values .

oruebel:
@stephprince could you take a look at this PR. I think this is essentially complete but needs some final attention to get over the finish line.

rly:
@stephprince do you have a chance to take a look at this? thanks!

stephprince:
Yes sorry for the delay I'll take a look!

src/pynwb/ecephys.py
stephprince: I updated this to iterate and check the length of all the position elements

tests/integration/hdf5/test_ecephys.py
stephprince: I also wasn't sure what the use case of the list would be - I added a test case with multiple tuples for now but not sure if that is necessary?

stephprince:
@rly this is ready to review

rly:
Ok, after doing a full review of the issue and PR, I am also confused as to why `ElectrodeGroup.position` is allowed to be a list of tuples. I think @oruebel may be mistaken. `ElectrodeGroup.position` should be a **scalar** dataset with compound dtype (x, y, z). @oruebel can you review?

oruebel:
> I think @oruebel may be mistaken. `ElectrodeGroup.position` should be a **scalar** dataset with compound dtype (x, y, z)

The schema [here](https://github.com/NeurodataWithoutBorders/nwb-schema/blob/3ad7aa370907c2cc177d033065dcba52140ca5a7/core/nwb.ecephys.yaml#L242-L255) says:

```
  datasets:
  - name: position
    dtype:
    - name: x
      dtype: float32
      doc: x coordinate
    - name: y
      dtype: float32
      doc: y coordinate
    - name: z
      dtype: float32
      doc: z coordinate
    doc: stereotaxic or common framework coordinates
    quantity: '?'
```

I.e., position is a compound dataset but the shape is not specified. Looking at the schema language docs [here](https://schema-language.readthedocs.io/en/latest/description.html#shape), I believe this means that `shape: null` is assumed so that it should be scalar. So yes, it seems that this should indeed be a scalar with a compound data type. So this PR should be updated accordingly. 

Question, how does the input for scalar compound datasets need to look like? Looking at the source issue #1768 setting:

- `position=np.array([1., 2., 3.])` : results in an error due to the validation in ElectrodeGroup
- `position=(1., 2., 3.)` : is being written as `[(1., 2., 3.) (1., 2., 3.) (1., 2., 3.)]`
-  `position=[1., 2., 3.]` : is being written as `[(1., 1., 1.) (2., 2., 2.) (3., 3., 3.)]`

Specifying as `[(1, 2, 3), ])` as currently on the PR fixes the bad write. Is `position=[(1, 2, 3), ])` the right form and we need to updated the validation to check that only a single tuple is given or is `position=(1., 2., 3.)` the correct form and we need to dig deeper into write?

stephprince:
After some further investigation, it turned out that scalar datasets with a compound datatype were being written incorrectly so that the dataset was not scalar. These changes should be addressed in the upcoming hdmf release (3.14.4), but the tests will fail until that is released.

With the hdmf update, I believe `ElectrodeGroup.position` validation should check whether the position is a scalar with compound dtype, or whether it’s a single list/tuple of length 3 that we then convert to a scalar with compound dtype.

rly:
This looks good to me but I will let @stephprince review and approve to merge.

stephprince (approved):
Thanks for the updates @rly. All looks good to me.

I will fix the broken links in a separate PR.

PR #1789, opened 2023-11-16
TermSet Config

mavaylon1:
## Motivation

We want to integrate the `TermSet` class into pynwb such that we provide default termsets to validate fields within an object. The approach here is to have a config file that stores the object type, the fields within the object that uses a termset, and the path to the termset. These termsets will love within nwb-schema, along with the config file. 

On the pynwb side, `init_validation` is a method call prior to setting the fields and calling __init__ from the super class. This will go through the fields of the object and check the config file for fields that support termsets. These fields will have their values replaced with a `TermSetWrapper`. 

## How to test the behavior?
```
```

## Checklist


mavaylon1:
TODO/Discussion:

1. Make the validation step a boolean.
2. Most likely move the `init_validation` to `AbstractContainer`. I kept it here for a quick proof of concept.
3. Finalize this approach in general.

PR #1793, opened 2023-11-29
Change timeseries data checks to warn instead of error on read

### LINKED ISSUE #1721: [Bug]: Able to write a TimeSeries with zero rate
### What happened?

I'm able to run the following code successfully - whereas it seems to me that should be prohibited.

Will definitely add a check for this on the Inspector side, but what are your thoughts @rly @oruebel?

### Steps to Reproduce

```python

mock_TimeSeries(rate=0.0)
```

### Traceback

_No response_

### Operating System

Windows

### Python Executable

Conda

### Python Version

3.11

### Package Versions

_No response_

### Code of Conduct

### LINKED ISSUE #1786: [Feature]: Warn on read rather than error

In `TimeSeries` (and likely a few other places) additional error checks are performed on `__init__` to ensure the data is valid. E.g,.:

https://github.com/NeurodataWithoutBorders/pynwb/blob/eb585063b3e035d68f206c49ff19200f9f25fa43/src/pynwb/base.py#L175-L191

This makes sense when creating a new file to ensure that we don't create invalid files. However, on read it would be useful to warn instead and continue to attempt to read the file to try and give access to the data if at all possible. 

Related to https://github.com/NeurodataWithoutBorders/helpdesk/discussions/62

Use the `self._error_on_new_warn_on_construct` method instead of `ValueError` in the `TimeSeries` constructor. 

Yes.

### Code of Conduct

stephprince:
## Motivation

Fixes #1786 and #1721.

- Changes data validation checks for rate and timestamps to warnings if reading in data
- adds error for making a new time series with a non-positive rate. Will only give warning if reading in data with non-positive rate

## How to test the behavior?

See example in #1721 

## Checklist

### DIFF
--- a/src/pynwb/base.py
+++ b/src/pynwb/base.py
@@ -174,15 +174,23 @@ class TimeSeries(NWBDataInterface):
         timestamps = args_to_process['timestamps']
         if timestamps is not None:
             if self.rate is not None:
-                raise ValueError('Specifying rate and timestamps is not supported.')
+                self._error_on_new_warn_on_construct(
+                    error_msg='Specifying rate and timestamps is not supported.'
+                )
             if self.starting_time is not None:
-                raise ValueError('Specifying starting_time and timestamps is not supported.')
+                self._error_on_new_warn_on_construct(
+                    error_msg='Specifying starting_time and timestamps is not supported.'
+                )
             self.fields['timestamps'] = timestamps
             self.timestamps_unit = self.__time_unit
             self.interval = 1
             if isinstance(timestamps, TimeSeries):
                 timestamps.__add_link('timestamp_link', self)
         elif self.rate is not None:
+            if self.rate <= 0:
+                self._error_on_new_warn_on_construct(
+                    error_msg='Rate must be a positive value.'
+                )
             if self.starting_time is None:  # override default if rate is provided but not starting time
                 self.starting_time = 0.0
             self.starting_time_unit = self.__time_unit
--- a/tests/unit/test_base.py
+++ b/tests/unit/test_base.py
@@ -444,6 +444,24 @@ class TestTimeSeries(TestCase):
                 timestamps=[1, 2, 3, 4, 5]
             )
 
+    def test_file_with_starting_time_and_timestamps_in_construct_mode(self):
+        """Test that UserWarning is raised when starting_time and timestamps are both specified
+         while being in construct mode (i.e,. on data read)."""
+        obj = TimeSeries.__new__(TimeSeries,
+                                 container_source=None,
+                                 parent=None,
+                                 object_id="test",
+                                 in_construct_mode=True)
+        with self.assertWarnsWith(warn_type=UserWarning,
+                                  exc_msg='Specifying starting_time and timestamps is not supported.'):
+            obj.__init__(
+                name="test_ts",
+                data=[11, 12, 13, 14, 15],
+                unit="volts",
+                starting_time=1.0,
+                timestamps=[1, 2, 3, 4, 5]
+            )
+
 
 class TestImage(TestCase):
     def test_init(self):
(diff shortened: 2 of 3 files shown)


src/pynwb/base.py
```diff
@@ -174,15 +174,23 @@ def __init__(self, **kwargs):
         timestamps = args_to_process['timestamps']
         if timestamps is not None:
             if self.rate is not None:
-                raise ValueError('Specifying rate and timestamps is not supported.')
+                self._error_on_new_warn_on_construct(
+                    error_msg='Specifying rate and timestamps is not supported.'
+                )
             if self.starting_time is not None:
-                raise ValueError('Specifying starting_time and timestamps is not supported.')
+                self._error_on_new_warn_on_construct(
```
rly, line 181: Could you add a test for this warning as well?

rly:
Please add a new entry or two to `CHANGELOG.md` under "PyNWB 2.6.0 (Upcoming)" > "Enhancements and minor changes" with a brief description of the changes.

rly:
The `nwbinspector` tests fail because they include tests that create a non-positive rate. With this PR, those tests will raise an error. Pinging @CodyCBakerPhD - these nwbinspector tests will need to be updated after this is released in PyNWB 2.6.0.

rly (approved):

tests/unit/test_base.py
```diff
@@ -405,6 +405,63 @@ def test_get_data_in_units(self):
         ts = mock_TimeSeries(data=[1., 2., 3.])
         assert_array_equal(ts.get_data_in_units(), [1., 2., 3.])
 
+    def test_non_positive_rate(self):
+        with self.assertRaisesWith(ValueError, 'Rate must be a positive value.'):
+            TimeSeries(name='test_ts', data=list(), unit='volts', rate=-1.0)
+        with self.assertRaisesWith(ValueError, 'Rate must be a positive value.'):
+            TimeSeries(name='test_ts1', data=list(), unit='volts', rate=0.0)
```
CodyCBakerPhD, line 412: @rly @stephprince Hey there

Does this mean that on next release of PyNWB it will be impossible to specify any TimeSeries with a rate of zero, even if it only has a single element on the first axis?

That's a fairly big break to back-compatibility if so, we've recommended to several people throughout the years to use that approach for statically generated microscopy images since that is the closest available data type for their setup (static images module lacks the associated metadata, and no specific static image types in the ophys module)

Not that it's a bad thing overall, but there should be a bigger changelog note about it IMO, along with some update about what we recommend to such users as an alternative

oruebel: Thanks, I was not aware that folks were using the Ophys modules this way. Could you clarify why rate matters when there is only a single image. Wouldn't you just use an explicit timestamp array in that case? I think it is fine to change the check to only check for "self.rate < 0" instead of "self.rate <= 0" to avoid the breaking case you mentioned (an maybe have an warning for "rate == 0").

CodyCBakerPhD: I guess the difference with that proposal is then how to treat `NaN` values for `timestamps`

For static images in this case, it might not matter _when_ it was acquired w.r.t. any other data streams in the file, so the `starting_time` could be set to `NaN` while the `rate` is set to a true value of `0.0`

It's true an equivalent representation could be to use `timestamps` but then it would be an array of one element, that being `NaN`, but we'd need checks elsewhere to ensure no more than one `NaN` is set in other instances, such as a series with `timestamps=[NaN, NaN, NaN, ...]` which would make less sense, right?

In all cases, this is really because there's a lack of appropriate neurodata object, so maybe I'll just kick the bucket to ndx-microscopy to resolve this in the best way

I do think we shouldn't break back-compatibility for the exact `0.0` case for the time being however, just so there's no unintended disruption

oruebel: In this case I would suggest the following:

- change the new check to only check for `self.rate > 0`
- optionally add a warning for `self.rate == 0`
- Wait for `ndx-microscopy` to address the missing data type

oruebel: @stephprince could you create a PR to make those changes

stephprince: yes I can do that

PR #1794, opened 2023-12-06
Disable nightly dandi read workflow for now

rly:
## Motivation

We recently added a github action to read a random set of nwb files from dandi. This always fails, usually after running for 2+ hours. The errors are almost always curl errors from using the ROS3 driver. This is not very informative until we have the bandwidth to use fsspec or remfile instead. I suggest we disable this workflow for now so that we don't waste compute cycles and I stop getting workflow failure emails. 

Also Yarik raised a good point that stochastic testing is not ideal - you might miss a bug simply due to bad luck. We should test systematically instead of randomly, e.g., scan through dandisets 000001-000100 on monday, 000101-000200 on tuesday, etc.

## How to test the behavior?
```
```

## Checklist


oruebel (dismissed):
I would create an issue for future reference and add a comment in the yaml file stating why the workflow is disabled, just to leave breadcrumbs for ourselves

oruebel (approved):

PR #1815, opened 2024-01-09
Support stimulus_template as optional column in IntracellularRecordingsTable

### LINKED ISSUE #1747: [Feature]: Support stimulus_template column in IntracellularRecordingsTable

See https://github.com/NeurodataWithoutBorders/nwb-schema/pull/545

_No response_

Update StimulusTable to support the new optional column

Yes.

### Code of Conduct

stephprince:
## Motivation

Fix #1747  to support `stimulus_template` as a predefined optional column of the `IntracellularStimuliTable`
(See https://github.com/NeurodataWithoutBorders/nwb-schema/pull/545 for the related schema PR)

Since the proposed column neurodata type is `TimeSeriesReferenceVectorData`, I though it was maybe unclear how to add data to the column so I added an example in the tutorials. 

The `IntracellularRecordingsTable.add_recording` had some nice functionality to automatically compute the start_index and count values for stimulus time series inputs, as well as handle recordings with no stimulus info provided. I added similar behavior for `stimulus_template`, but I'm not sure if that's something that would be helpful or if it's best to add additional stimulus template data via the `stimulus_metadata` argument

## How to test the behavior?
```
from pynwb.base import TimeSeriesReferenceVectorData
from pynwb.testing.mock.icephys import (mock_IntracellularRecordingsTable,
                                        mock_VoltageClampStimulusSeries,
                                        mock_IntracellularElectrode,)

# adding data by row
nwbfile = mock_NWBFile()
electrode = mock_IntracellularElectrode(nwbfile=nwbfile)
stimulus = mock_VoltageClampStimulusSeries(nwbfile=nwbfile)
nwbfile.add_intracellular_recording(electrode=electrode,
                                    stimulus=stimulus,
                                    stimulus_template=stimulus)

# adding data by column
irt = mock_IntracellularRecordingsTable()
irt.add_column(name='stimulus_template',
               description='',
               category='stimuli',
               col_cls=TimeSeriesReferenceVectorData,
               data=[(0, 5, stimulus),
                     (0, 5, stimulus),
                     (0, 5, stimulus),
                     (0, 5, stimulus),
                     (0, 5, stimulus),])
```

## Checklist


docs/gallery/domain/plot_icephys.py
```diff
@@ -457,6 +458,38 @@
...
+#####################################################################
+# One predefined subcategory column is the stimulus template column in the stimuli table. This column can be used to
+# store stimulus template waveforms that were generated by the recording software. Similar to the stimulus and response
+# columns, we can specify a relevant time range.
+
+nwbfile.intracellular_recordings.add_column(
+    name="stimulus_template",
+    data=[(0, 5, stimulus),  # (start_index, index_count, stimulus_template) can be specified as a tuple
+          (1, 3, stimulus),
+          (-1, -1, stimulus)],
```
oruebel: I think it may be useful to use `TimeSeriesReference` here, instead of raw tuple to introduce the class and be consistent with how the data represented on read. 

https://github.com/NeurodataWithoutBorders/pynwb/blob/883b1b15810b4a040698c9e40d717f143b3a3248/src/pynwb/base.py#L378

```suggestion
    data=[TimeSeriesReference(0, 5, stimulus),  # (start_index, index_count, stimulus_template) can be specified as a tuple
          TimeSeriesReference(1, 3, stimulus),
          TimeSeriesReference(-1, -1, stimulus)],
```

oruebel: Just a thought, for the empty case ``(-1, -1, stimulus)`` we could maybe add a factory method to `TimeSeriesReference` to create an empty reference, e.g., `TimeSeriesReference.empty()`?

stephprince: The empty reference method sounds helpful. Would we still want a time series (e.g. stimulus) to be required as an input for that method? Or would it create an empty time series with arbitrary units.

oruebel: Good question. I think having an empty `TimeSeries` to point to makes sense. I'm not sure what the best way is to autogenerate that, since we would need to also add it to the file. For now, maybe we can still require that a user to provide the `TimeSeries` and maybe have a best practice on how do to this. Let's discuss when we meet next time how we may be able to do this without the user providing a `TimeSeries`. The `TimeSeries` is necessary because this is stored as compound data type, so all elements must be set, and we don't have a NaN value for ints and object references, so `(-1, -1, <some ref>)` is just a way of saying NaN or invalid. Technically, it really doesn't matter where the reference points to, so maybe there is a way that we can avoid needed a `TimeSeries` at all.

docs/gallery/domain/plot_icephys.py
```diff
@@ -457,6 +458,38 @@
     category="electrodes",
 )
 
+#####################################################################
+# One predefined subcategory column is the stimulus template column in the stimuli table. This column can be used to
+# store stimulus template waveforms that were generated by the recording software. Similar to the stimulus and response
+# columns, we can specify a relevant time range.
```
oruebel:
```suggestion
# One predefined subcategory column is the ``stimulus_template`` column in the stimuli table. This column is
# used to store template waveforms of stimuli in addition to the actual recorded stimulus that is stored in the
# ``stimulus`` column. Similar to the ``stimulus`` and ``response`` columns, we can specify a relevant
# time range.
```

docs/gallery/domain/plot_icephys.py
```diff
@@ -457,6 +458,38 @@
...
+    description="Column storing the reference to the stimulus template for the recording (rows).",
+    category="stimuli",
+    col_cls=TimeSeriesReferenceVectorData
+)
+
+# we can also add stimulus template data as follows
+rowindex = nwbfile.add_intracellular_recording(
+    electrode=electrode,
+    stimulus=stimulus,
+    stimulus_template=stimulus,  # the full time range of the stimulus template will be used unless specified
```
oruebel: We should not use the same TimeSeries for the `stimulus` and `stimulus_template`. I would suggest we create a new TimeSeries for the stimulus template and add it in the proper location for templates. Looking at the schema, the presented stimulus should I believe go into the `stimulus/presentation` group and the template in `stimulus/templates` group. Similarly, above we should use the template not the actual stimulus when adding data to the `stimulus_template` column.

docs/gallery/domain/plot_icephys.py
```diff
@@ -457,6 +458,38 @@
     category="electrodes",
 )
 
+#####################################################################
+# One predefined subcategory column is the stimulus template column in the stimuli table. This column can be used to
```
oruebel: Since this is a pre-defined column, I think it may be more intuitive to add this as as subsection in the `Adding an intracellular recording` section. 

I would also add a brief description on what the purpose of the column is. I.e,. the `stimulus` is the actual stimulus that has been recorded from the system as it was applied to the cell. The `stimulus_template` on the other hand is a template (i.e,. an idealized version) of the waveform that is being used as the stimulus. This is useful, e.g., to validate that the actual stimulus that was applied matches the expected waveform of the template. 

I would also add a brief note at the end about the use of timestamps in stimulus templates. Because they are templates that are being reused across many recordings, the timestamps in the templates are usually not aligned with the recording (and often not the reference time of the file either), but they often just start at 0 and are seen as relative to whatever the time of the application of the stimulus is.

src/pynwb/base.py
```diff
@@ -497,6 +497,25 @@ def data(self):
...
+        """
+        Creates an empty TimeSeriesReference class to represent missing data.
+
+        When missing data needs to be represented, NWB defines ``None`` for the complex data type ``(idx_start,
+        count, TimeSeries)`` as (-1, -1, TimeSeries) for storage. The exact timeseries object will technically not
+        matter since the empty reference is a way of indicating a NaN value. In practice, this may be used in the
+        :py:class:`~pynwb.icephys.IntracellularRecordingsTable` where only one of stimulus or response data was
+        recorded. In such cases, the timeseries object for the empty stimulus TimeSeriesReference could be set to the
+        response series or vice versa (the timeseries object for the empty response TimeSeriesReference could be set
+        to the stimulus)
```
oruebel:
```suggestion
        matter since the empty reference is a way of indicating a NaN value. in a 
        :py:class:`~pynwb.base.TimeSeriesReferenceVectorData` column. 
        
        An example, where this functionality is used is :py:class:`~pynwb.icephys.IntracellularRecordingsTable` 
        where only one of stimulus or response data was  recorded. In such cases, the timeseries object for the
        empty stimulus :py:class:`~pynwb.base.TimeSeriesReference` could be set to the response series, or 
        vice versa.
```

docs/gallery/domain/plot_icephys.py
```diff
@@ -97,6 +97,7 @@
 
 # Import additional core datatypes used in the example
 from pynwb.core import DynamicTable, VectorData
+from pynwb.base import TimeSeriesReference, TimeSeriesReferenceVectorData
```
oruebel, line 100: Just as side note, technically we could use `TimeSeriesReferenceVectorData.TIME_SERIES_REFERENCE_TUPLE` instead, but I think importing `TimeSeriesReference` directly looks cleaner.

oruebel (changes requested):
Looks overall good to me. I added a couple of comments for the tutorial part but those should be easy to fix.

docs/gallery/domain/plot_icephys.py
stephprince: I added an `Adding stimulus templates` subsection in the `Adding an intracellular recording` section. I kept it after the `adding custom columns...` section for now since it uses the `add_column` function and `category` arguments that get explained there

oruebel:
Looks good to me. @rly is doing a final review an will approve once ready.

src/pynwb/base.py
```diff
@@ -497,6 +497,26 @@ def data(self):
         # load the data from the timeseries
         return self.timeseries.data[self.idx_start: (self.idx_start + self.count)]
 
+    @classmethod
+    @docval({'name': 'timeseries', 'type': TimeSeries, 'doc': 'the timeseries object to reference.'})
+    def empty(cls, timeseries):
+        """
+        Creates an empty TimeSeriesReference class to represent missing data.
```
rly:
```suggestion
        Creates an empty TimeSeriesReference object to represent missing data.
```

src/pynwb/base.py
```diff
@@ -497,6 +497,26 @@ def data(self):
...
+
+        When missing data needs to be represented, NWB defines ``None`` for the complex data type ``(idx_start,
+        count, TimeSeries)`` as (-1, -1, TimeSeries) for storage. The exact timeseries object will technically not
+        matter since the empty reference is a way of indicating a NaN value in a
+        :py:class:`~pynwb.base.TimeSeriesReferenceVectorData` column.
+
+        An example where this functionality is used is :py:class:`~pynwb.icephys.IntracellularRecordingsTable`
+        where only one of stimulus or response data was recorded. In such cases, the timeseries object for the
+        empty stimulus :py:class:`~pynwb.base.TimeSeriesReference` could be set to the response series, or vice versa.
```
rly, line 513: This seems like an inconvenience and I think it would confuse users. What about creating a dummy TimeSeries to use here? Or using None as the object reference?

rly: The problem with creating a dummy TimeSeries is that it has to be added somewhere...

stephprince: @rly I agree it's confusing . We discussed this a bit above


> Good question. I think having an empty `TimeSeries` to point to makes sense. I'm not sure what the best way is to autogenerate that, since we would need to also add it to the file. For now, maybe we can still require that a user to provide the `TimeSeries` and maybe have a best practice on how do to this. Let's discuss when we meet next time how we may be able to do this without the user providing a `TimeSeries`. The `TimeSeries` is necessary because this is stored as compound data type, so all elements must be set, and we don't have a NaN value for ints and object references, so `(-1, -1, <some ref>)` is just a way of saying NaN or invalid. Technically, it really doesn't matter where the reference points to, so maybe there is a way that we can avoid needed a `TimeSeries` at all.

stephprince: right now `IntracellularRecordingsTable` automatically [fills in the missing stimulus / response](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/icephys.py#L575)  - so having a dummy TimeSeries might be helpful there as well?

rly: @oruebel and I discussed this briefly and decided to leave this PR be and ask @stephprince to try to set the object reference to None in a separate PR.

rly (approved):

rly:
@oruebel because you requested changes earlier, you still need to approve this

oruebel (approved):

rly (approved):

PR #1823, opened 2024-01-13
update remfile verbiage in streaming.py

magland:
@rly 

Here are some propsed updates to the remfile verbiage in pynwb.

The claims of fast initial load times (especially on a wifi network, but also on dandihub) are [supported here](https://github.com/scratchrealm/pynwb_streaming_benchmark).

Update: here are [some further tests on a GitHub codespace](https://github.com/magland/remfile/blob/main/benchmarks/README.md), including ros3 method (which is way slower than the other two)

### DIFF
--- a/docs/gallery/advanced_io/streaming.py
+++ b/docs/gallery/advanced_io/streaming.py
@@ -131,9 +131,11 @@ with NWBHDF5IO(s3_url, mode='r', driver='ros3') as io:
 ##################################################
 # Method 3: remfile
 # -----------------
-# ``remfile`` is another library that enables indexing and streaming of files in s3. remfile is simple, fast, and
-# allows for caching of data in the local filesystem. The caveats of ``remfile`` are that it is a very new project
-# that has not been tested in a variety of use-cases and caching options are limited compared to ``fsspec``.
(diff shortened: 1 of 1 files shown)


rly (approved):

rly:
Thanks @magland !

@sinha-r @oruebel please see above for some initial streaming benchmarks from @magland

rly:
@magland It's interesting to see ros3 so much slower and good to have hard numbers to back up the claims.

FYI This year as part of supplemental funding for our NWB U24 grant, we will be doing a set of comprehensive tests evaluating method (fsspec, ros3, remfile, kerchunk), environment (dandihub, nersc/home network), data type (ecephys, ophys, etc.), file settings (chunking, compression, etc.), and metrics (time of initial load, time to read various chunks of data). We haven't started running those tests yet, but when we do, it will be useful for us to compare your benchmark results with ours.

oruebel:
Thanks @magland My guess is that the large difference in the initial load for ros3 probably means that ros3 makes many more small web requests in sequence, i.e., I suspect network latency is a main factor here. Would be interesting to run these with a network profiler to get and idea of how many requests and data are being transferred with each method.

PR #1827, opened 2024-01-17
Add internal link checking to sphinx-build

### LINKED ISSUE #1731: [Feature]: Run sphinx build in nitpicky mode to check internal links

We can run `sphinx-build -n ...` to have sphinx check internal links. We currently check external links through the sphinx --linkcheck option but this does not check whether references like `pynwb.base.ProcessingModules` exists. When running this locally, I get over 100 broken internal links in docstrings, and about 30 broken internal links in tutorials.

_No response_

Update Makefile and make.bat in docs to use the sphinx option `-n`. Run `make html`. Fix all the warnings about broken links.

Yes.

### Code of Conduct

stephprince:
## Motivation

Fix #1731. 

Running sphinx-build in nitpicky mode can be used to check internal links. This PR will update the GitHub workflow / Makefile to use nitpicky mode and fix broken links/references.

A couple of notes
- Fixing all internal links will also require updates to hdmf (See related PR https://github.com/hdmf-dev/hdmf/pull/1031)
- Running sphinx-build in nitpicky mode will generate warnings by default. If we want to be stricter, we can add the `-W` flag to generate errors for broken internal links
- colons in docstrings get picked up by sphinx as references due to the sphinx Napoleon extension ([See issue here](https://github.com/sphinx-doc/sphinx/issues/9273)). For now, I replaced colons with dashes, but there are a couple options for getting colons to work if it's preferred to keep them. See the image below for examples of what they might look like:
Screenshot 2024-01-17 at 10 17 38 AM

- Using type hints with h5py data types causes internal reference warnings. I couldn't figure out the best way to approach fixing that currently. It only happened in the `get_nwbfile_version` function so I switched it to use `docval` instead
## How to test the behavior?
```
cd docs
make linkcheck
```

## Checklist


src/pynwb/__init__.py
```diff
@@ -251,7 +255,7 @@ def can_read(path: str):
              'doc': 'a path to a namespace, a TypeMap, or a list consisting paths to namespaces and TypeMaps',
              'default': None},
             {'name': 'file', 'type': [h5py.File, 'S3File'], 'doc': 'a pre-existing h5py.File object', 'default': None},
-            {'name': 'comm', 'type': "Intracomm", 'doc': 'the MPI communicator to use for parallel I/O',
+            {'name': 'comm', 'type': '~mpi4py.MPI.Intracomm', 'doc': 'the MPI communicator to use for parallel I/O',
```
oruebel:
```suggestion
            {'name': 'comm', 'type': '~mpi4py.MPI.Intracomm', 'doc': 'the MPI communicator to use for parallel I/O',
```

I'm not sure `'~mpi4py.MPI.Intracomm'` will work correctly with the `docval` decorator. `docval` uses this to check type by checking for the classname, I believe the line in HDMF is here 

https://github.com/hdmf-dev/hdmf/blob/b4bdcef10071199ab28e6ad3822b7fd47bfab4b3/src/hdmf/utils.py#L99

and think '~mpi4py.MPI.Intracomm' may not match this check. You can try and pass an mpi Intracomm object to the function and see if docval raises an error or not. When possible we list the actual class in the `type` key for docval but in this case we are just specifying the name of the class to avoid having to import mpi4py just to use it in docval.

oruebel:
> * For now, I replaced colons with dashes

I think that's fine.

src/pynwb/__init__.py
oruebel: I don't think we are currently covering mpi in our unit tests, so this probably does not get picked up by tests yet. We have a short tutorial for mpi, but that one does not actually run anything. 

https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/gallery/advanced_io/parallelio.py 

If I remember correctly, this is because testing with mpi requires a custom install for h5py to build against HDF5 with mpi enabled.

oruebel (changes requested):
Thanks for all these fixes. Aside from the issue with the mpi Intracom in docval this looks good to me.

docs/gallery/domain/images.py
```diff
@@ -190,7 +190,7 @@
 # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 #
 # :py:class:`~pynwb.image.RGBAImage` is for storing data of color image with transparency.
-# :py:attr:`~pynwb.image.RGBAImage.data` must be 3D where the first and second dimensions
+# `RGBAImage.data` must be 3D where the first and second dimensions
```
rly:
```suggestion
# ``RGBAImage.data`` must be 3D where the first and second dimensions
```

Unintuitively, Sphinx uses one backtick for italics and two for code.

rly: Can you change the other instances of this in this file and other files too?

src/pynwb/__init__.py
rly: I agree. I think this will not work unfortunately. We could update docval to allow us to specify a classname with module prefix. This would require some testing, but it could be useful in allowing us to be more explicit about which class we mean to allow. Or we could hardcode an exception in docval for interpreting "Intracomm" as ":py:class~`mpi4py.MPI.Intracomm`" when creating the sphinx docstring. That's easier. Either approach is fine with me.

rly:
That's a lot of fixes! Thank you.

src/pynwb/__init__.py
oruebel: I would just leave it as `'Intracomm'` for now here. We are using the same logic in a couple of places, e.g., just two lines above for `'S3File'`. If this is a problem for the Sphinx link checker, then we could maybe just make it a list of types  `['Intracomm', ]` to work around sphinx checking the link. 

> it could be useful in allowing us to be more explicit about which class we mean to allow

I agree that would be a nice-to-have feature, but I don't think it is a priority. 

> we could hardcode an exception in docval

That seems brittle and would require exception for other types too.

stephprince: I see, that makes sense. Yes I can keep it as `'Intracomm'` for now and add an exception/workaround for the link checker.

stephprince: Even having a list of types `['Intracomm', ]` didn't seem to fix the reference warnings. I'm not sure if the `S3File` worked because there was another valid reference in the list? I added it to the `nitpick_ignore` section of the `conf.py` file as a potential solution 

I also moved the `nitpicky=True` flag to `conf.py` so it's all in the same place and consistent across make/workflow files.

oruebel: Thanks, that solution looks reasonable to me.

oruebel (dismissed):
This looks all good to me. I would wait for @rly to approve as well, but I believe I need to approve too since I requested changes earlier.

stephprince:
did we want to raise nitpicky warnings as errors in the workflow here too?

rly:
> did we want to raise nitpicky warnings as errors in the workflow here too?

Yes, that would be great. 

It's too bad that the type hint for `h5py.File` doesn't play nicely with sphinx

rly:
Looks good to me. To be clear, the remaining warning will be fixed in the next release of HDMF, right?
```
/home/runner/work/pynwb/pynwb/src/pynwb/__init__.py:docstring of pynwb.docval.<locals>.dec.<locals>.func_call:9:py:class reference target not found: File
```

rly:
If so, let's hold off on merging until that's released.

rly (dismissed):

stephprince:
Yes the remaining warnings should be fixed in the next release of HDMF. Right now it's stopping at the first warning (there are a lot more). If we want to show the rest of the errors in the workflow run we could add the --keep-going flag.

rly:
@stephprince It looks like there is a new failure in the build - could you please look into it today if you can? If not, let me know and I'll try. I would like this in the next pynwb release today/tomorrow.

rly (approved):

PR #1829, opened 2024-01-18
Integrate TypeConfig

mavaylon1:
## Motivation

Integrate TypeConfigurator from HDMF to pynwb. Also provide documentation.

- [x] load and unload should use the load and unload from hdmf
- [x ] tutorial
- [x] restructure the default configs to match the new structure set in hdmf
- [x ] tests

## How to test the behavior?
```
```

## Checklist


src/pynwb/__init__.py
```diff
@@ -18,6 +18,35 @@
 from .spec import NWBDatasetSpec, NWBGroupSpec, NWBNamespace  # noqa E402
 from .validate import validate  # noqa: F401, E402
 
+from .termset_config import NWBTermSetConfigurator
+
+
+nwb_config = NWBTermSetConfigurator(path='src/pynwb/config/nwb_config.yaml')
```
oruebel: I know this is still in development. I think ultimately this path should be relative to the pynwb module (i.e. ``path='config/nwb_config.yml'`) not the structure in the git repo for this to work with pip package, similar to how the schema is being installed as an asset: 

https://github.com/NeurodataWithoutBorders/pynwb/blob/dfc19bd8a7576d454013ca74132babd4a42d32f4/src/pynwb/__init__.py#L31

mavaylon1: I appreciate this

mavaylon1: I just pushed an alternative approach. ``path='config/nwb_config.yml'` leads to issues importing within docs.

oruebel: I believe that should work. I think you'll  also need to update `setup.py` to pick this up when creating the python package for pip/conda. 

https://github.com/NeurodataWithoutBorders/pynwb/blob/dfc19bd8a7576d454013ca74132babd4a42d32f4/setup.py#L46

mavaylon1:
@oruebel This is just about done. The main lingering issue left really is the external links and adding the last bit of weird coverage.

1. The gallery tests/3.12 tests that are failing seem to be due to no PR related issues. I will take a look and will create issue tickets. 
2. I am not sure why the external link is failing for "_file_". @rly do you have any ides?
3. The coverage is weird because those methods are actively being called in the tests. Thoughts on why they are not counting as covered. The "or" shouldn't matter because we have the same syntax in HDMF but we don't test for both. On second thought I think it is because we do not have tests for optional requirements. That should do it.

docs/gallery/general/plot_configurator.py
```diff
@@ -0,0 +1,108 @@
...
+# As mentioned prior, the first step after creating a configuration file is
+# to load the file.
+# It is important to remember that with the configuration loaded, the fields
+# are wrapped automatically, meaning the user should proceed with creating
+# the instances normally, i.e., without wrapping directly. In this example,
+# we load the a configuration file that associates a
+# :py:class:`~hdmf.term_set.TermSet` for the ``species`` field in
+# :py:class:`~pynwb.file.Subject`.
+
+dir_path = os.path.dirname(os.path.abspath(__file__))
```
rly:
```suggestion
dir_path = os.path.dirname(os.path.abspath("plot_configurator.py"))
```
I think `__file__` does not exist when running under certain types of scripts or environments. I think this should work in its place.

rly: Another approach: We did this in the icephys tutorial:
```python
try:
    tutorial_path = os.path.abspath(__file__)  # when running as a .py
except NameError:
    tutorial_path = os.path.abspath("__file__")  # when running as a script or notebook
```

mavaylon1: What's weird is that "__file__" is what we have in hdmf and it works

rly: That is weird. I don't know. If you want to track it down, you could look into differences in sphinx versions or the link checker call.

mavaylon1:
Review Notes:
As you can see a lot is depending on the fix to have zarr being able to use 2.18.1. Every failing test is optional and is due to zarr.

docs/gallery/general/plot_configurator.py
```diff
@@ -0,0 +1,108 @@
...
+We also provide a helper method to see the configuration that has been loaded:
+:py:func:`~pynwb.get_loaded_type_config`
+
+
+How to make a Configuration File
+--------------------------------
+Before taking advantage of the all the wonders that comes with using a configuration file,
+the user needs to create one following some simple guidelines. To follow along with an example,
+please refer to ``nwb_config.yaml`` under ``src/config``.
+The configuration files is built on the foundation of the YAML syntax. The
```
rly:
```suggestion
The configuration file uses YAML syntax. The
```

docs/gallery/general/plot_configurator.py
```diff
@@ -0,0 +1,108 @@
...
+user will construct a series of nested dictionaries to encompass all the necessary information.
+
+1. The user needs to define all the relevant namespaces. Recall that each neurodata type exists within
+   a namespace, whether that is the core namespace in PyNWB or a namespace in an extension. As namespaces grow,
+   we also require a version to be recorded in the configuration file to ensure proper functionality.
+2. Within a namespace dictionary, the user will have a list of data types the want to use.
```
rly:
```suggestion
2. Within a namespace dictionary, the user will have a list of data types they want to configure.
```

docs/gallery/general/plot_configurator.py
```diff
@@ -0,0 +1,108 @@
+"""
+TypeConfigurator
+=======
+
+This is a user guide for how to take advantage of the
+:py:class:`~hdmf.term_set.TypeConfigurator` class by creating configuration files in
+order to more easily validate terms within datasets or attributes.
```
rly:
```suggestion
Configuring PyNWB to allow only certain terms
===================================

This is a user guide for how to configure PyNWB to allow only certain terms for
particular fields within datasets and attributes using a configuration file. This
approach can be easier than wrapping individual data objects with a
:py:class:`~hdmf.term_set.TermSetWrapper`.
```

rly: "Configuring PyNWB to allow only certain terms" - I don't like this title either. I think this is too broad. @mavaylon1 any ideas for a succinct but descriptive title?

mavaylon1: How about "How to Configure Term Validations" @rly

docs/gallery/general/plot_configurator.py
```diff
@@ -0,0 +1,108 @@
...
+
+Introduction
+-------------
+Users do not directly interact with the :py:class:`~hdmf.term_set.TypeConfigurator` class.
+Instead, users will create a configuration YAML file that outlines the fields (within a neurodata type)
+they want to be validated. After creating the configuration file, users will need to load the
+configuration with the :py:func:`~pynwb.load_type_config` method.
```
rly: A user should not need to know about the TypeConfigurator class, so I think we do not need to talk about it. It's basically an internal class, like the TypeMap, BuildManager, etc.

docs/gallery/general/plot_configurator.py
rly: Sounds good

docs/gallery/general/plot_configurator.py
```diff
@@ -0,0 +1,108 @@
...
+
+subject = Subject(
+    subject_id="001",
+    age="P90D",
+    description="mouse 5",
+    species="Mus musculus",
+    sex="M",
+)
+
+nwbfile.subject = subject
```
rly, line 100: Can you say something about what is happening with this particular configuration and this particular NWBFile?

src/pynwb/__init__.py
```diff
@@ -12,13 +12,48 @@
...
+
 
 CORE_NAMESPACE = 'core'
 
 from .spec import NWBDatasetSpec, NWBGroupSpec, NWBNamespace  # noqa E402
 from .validate import validate  # noqa: F401, E402
 
 
+CUR_DIR = os.path.dirname(os.path.realpath(__file__))
+path_to_config = os.path.join(CUR_DIR, 'config/nwb_config.yaml')
```
rly: How does this work if someone has run `pip install pynwb`? Where does the config file live? In that case, it probably makes sense as a file in the user's home directory like `~/.pynwb/nwb_config.yaml`

rly: Alternatively, we could specify no default, and the user must supply a path to a configuration file.

tox.ini
```diff
@@ -32,6 +32,14 @@ deps =
...
 commands = {[testenv]commands}
 
+# Test with python 3.11; optional and dev reqs (some features are not compatible with 3.12 yet)
+[testenv:py311-optional]
+basepython = python3.11
+deps =
+    -rrequirements-dev.txt
+    -rrequirements-opt.txt
+commands = {[testenv]commands}
+
```
rly, line 42: Is this workflow called?

rly:
I think we need to figure out the workflow for a user who has installed pynwb via pip. Otherwise this all looks good.

tox.ini
mavaylon1: I believe so. That's how I got the coverage to pass.

mavaylon1: Unless I am missing something

mavaylon1: Oh the run tests workflow

src/pynwb/__init__.py
mavaylon1: I will remove the default.

tox.ini
rly: Yeah, "tox -e py311-optional" needs to be added to the run tests workflow and run-all-tests workflow.

mavaylon1: Slipped my mind. Thanks

docs/gallery/general/plot_configurator.py
```diff
@@ -0,0 +1,112 @@
...
+for more information.
+
+To unload a configuration, simply call :py:func:`~pynwb.unload_type_config`.
+We also provide a helper method to see the configuration that has been loaded:
+:py:func:`~pynwb.get_loaded_type_config`
+
+
+How to make a Configuration File
+--------------------------------
+To see an example of a configuration file, please refer to ``nwb_config.yaml`` under ``src/config``.
```
rly:
```suggestion
To see an example of a configuration file, please refer to ``nwb_config.yaml`` under ``src/config``.
```

This file is not accessible if the user pip installs pynwb, which will be the vast majority of users. Can you point to a github link to this file instead? Also can you move the file to `/example` or another directory not in `/src/`since we are not using it by default?

mavaylon1: Would this link exist now or after the merge? The link pointing to it now would be on a branch.

docs/gallery/general/plot_configurator.py
```diff
@@ -0,0 +1,112 @@
...
+The configuration file uses YAML syntax. The
+user will construct a series of nested dictionaries to encompass all the necessary information.
```
rly, line 36: @mavaylon1 The writing style here is very formal, e.g., saying "the user will". Could you please adjust the writing style here and below to speak to the user directly, as if you were guiding them through this one on one? Second-person, imperative voice. e.g., "To accomplish X, do Y", or "Next, let's do X", and sometimes "You can do X", or "To accomplish X, we will do Y"? We don't have a style guide for this, but I think generally writing as if you were speaking directly to the user is more approachable and friendly for them.

docs/gallery/general/plot_configurator.py
rly: The link would exist after the merge, but you can figure out what the path would be after merge and link to that.

mavaylon1: Wouldn't that fail check external link since it won't exist. Currently, it already fails due to zarr, but in practice this wouldn't work?

rly: Yes, that would fail the external link check before merge. That's OK. The link would exist after merge, and we can confirm that it works and the link check passes after merge.

docs/gallery/general/plot_configurator.py
mavaylon1: I don't disagree, but also I don't want to rewrite this currently. I also checked my other tutorials and they also have my style that would also need to change. I can do a tutorial rewrite as a low issue ticket if that is a middle ground option.

rly: OK that's fine. Please create a ticket so we don't forget to do this eventually.

mavaylon1:
@rly issue ticket has been made, link has been added with what would be the link after merge.

docs/gallery/general/plot_configurator.py
```diff
@@ -0,0 +1,113 @@
...
+
+To unload a configuration, simply call :py:func:`~pynwb.unload_type_config`.
+We also provide a helper method to see the configuration that has been loaded:
+:py:func:`~pynwb.get_loaded_type_config`
+
+
+How to make a Configuration File
+--------------------------------
+To see an example of a configuration file, please refer to
+`<https://github.com/NeurodataWithoutBorders/pynwb/tree/dev/src/pynwb/config/nwb_config.yaml>`_.
```
rly, line 34:
```suggestion
`<https://github.com/NeurodataWithoutBorders/pynwb/tree/dev/docs/gallery/general/nwb_gallery_config.yaml>`_.
```

What do you think about pointing to this file and removing the `src/pynwb/config` folder? I think the gallery config files serve the same purpose as `src/pynwb/config` unless I am missing something.

docs/gallery/general/nwb_gallery_config.yaml
```diff
@@ -0,0 +1,10 @@
+namespaces:
+  core:
+    version: 2.7.0-alpha
```
rly: Can you remind me, when NWB schema 2.8.0 is released and used by PyNWB, will this work, or do we (and users) need to constantly keep this up to date? Will the user get an error if there is a version mismatch?

mavaylon1: As of right now, we don't do anything with the version. It is purely a best practice until we decide to enforce a check

docs/gallery/general/plot_configurator.py
mavaylon1: So the purposes of both are slightly different. The one in gallery is for the tutorial, but the one in `src` is the NWB curated one that we would recommend and maintain over time.

docs/gallery/general/nwb_gallery_config.yaml
rly: Got it. Thanks. I think we should never point to unreleased versions of the schema. Can we make this change?
```suggestion
    version: 2.7.0
```

docs/gallery/general/plot_configurator.py
rly: I see. It wasn't clear to me and I think it would not be clear to the user that we are recommending and maintaining that configuration file for users to use. I know you want to get this merged, so let's move forward, and I'll add this suggestion to the docs refactor issue.

rly (dismissed):

src/pynwb/config/nwb_config.yaml
```diff
@@ -0,0 +1,7 @@
+namespaces:
+  core:
+    version: 2.7.0-alpha
```
rly:
```suggestion
    version: 2.7.0
```

rly (dismissed):

rly:
In a separate PR, I updated the dev branch to use HDMF 3.14.0 throughout, so the ros3 test should pass now.

rly (approved):

PR #1836, opened 2024-01-30
add docs for editing NWB files

bendichter:
## Motivation

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/docs/gallery/advanced_io/plot_editing.py
+++ b/docs/gallery/advanced_io/plot_editing.py
@@ -71,8 +71,8 @@ with NWBHDF5IO("test_edit.nwb", "r+") as io:
 # created. If the dataset was created with a flexible shape, then it is possible to
 # change in-place. Creating a dataset with a flexible shape is done by specifying the
 # ``maxshape`` argument of the :py:class:`~hdmf.backends.hdf5.h5_utils.H5DataIO` class
-# constructor. Using a ``None`` value for a component of the ``maxshape`` tuple allows 
-# the size of the corresponding dimension to grow, such that is can be be reset arbitrarily long 
+# constructor. Using a ``None`` value for a component of the ``maxshape`` tuple allows
+# the size of the corresponding dimension to grow, such that is can be be reset arbitrarily long
 # in that dimension. Chunking is required for datasets with flexible shapes. Setting ``maxshape``,
 # hence,  automatically sets chunking to ``True``, if not specified.
 #
@@ -105,7 +105,7 @@ with NWBHDF5IO("test_edit2.nwb", "w") as io:
 ##############################################
 # The ``None``value  in the first component of ``maxshape`` means that the
 # the first dimension of the dataset is unlimited. By setting the second dimension
-# of ``maxshape`` to ``100``, that dimension is fixed to be no larger than ``100``. 
+# of ``maxshape`` to ``100``, that dimension is fixed to be no larger than ``100``.
 # If you do not specify a``maxshape``, then the shape of the dataset will be fixed
 # to the shape that the dataset was created with. Here, you can change the shape of
 # the first dimension of this dataset.
@@ -128,7 +128,7 @@ with NWBHDF5IO("test_edit2.nwb", "r+") as io:
 #
 # Editing groups
 # --------------
-# Editing of groups is not yet supported in PyNWB. 
+# Editing of groups is not yet supported in PyNWB.
 # To edit the attributes of a group, open the file and edit it using :py:mod:`h5py`:
 
 import h5py
--- a/docs/gallery/advanced_io/plot_editing.py
+++ b/docs/gallery/advanced_io/plot_editing.py
@@ -10,6 +10,8 @@ editing and running a validation check on the file after editing it.
 
 In-place editing with h5py
 ---------------------------
+The :py:mod:`h5py` package allows for in-place editing of HDF5 files. Where possible,
+this is the recommended way to edit HDF5 files.
 
 Editing a dataset value
 ~~~~~~~~~~~~~~~~~~~~~~~
@@ -120,7 +122,7 @@ with h5py.File("test_edit2.nwb", "r+") as f:
 #
 # Replacing a dataset in h5py
 # ----------------------------
-# There are several types of dataset edits that cannot be done in-place.
+# There are several types of dataset edits that cannot be done in-place:
 #
 #     * Changing the shape of a dataset with a fixed shape
 #     * Changing the datatype of a dataset
@@ -133,11 +135,11 @@ with h5py.File("test_edit2.nwb", "r+") as f:
 # the data from the old dataset to the new dataset, and deleting the old dataset.
 
 with h5py.File("test_edit2.nwb", "r+") as f:
-    data = f["acquisition"]["synthetic_timeseries"]["data"][:]
+    data_values = f["acquisition"]["synthetic_timeseries"]["data"][:]
     del f["acquisition"]["synthetic_timeseries"]["data"]
     f["acquisition"]["synthetic_timeseries"].create_dataset(
         name="data",
-        data=data,
+        data=data_values,
         maxshape=(None, 100),
         chunks=(100, 100),
         compression="gzip",
--- a/docs/gallery/advanced_io/plot_editing.py
+++ b/docs/gallery/advanced_io/plot_editing.py
@@ -4,10 +4,10 @@
 Editing NWB files
 =================
 
-This tutorial demonstrates how to edit NWB files in-place. For information on how to add
-or remove containers from an NWB file, see :ref:`modifying_data`. How and whether it is
-possible to edit an NWB file depends on the storage backend
-and the type of edit.
+This tutorial demonstrates how to edit NWB files in-place to make small changes to
+existing containers. To add or remove containers from an NWB file, see
+:ref:`modifying_data`. How and whether it is possible to edit an NWB file depends on the
+storage backend and the type of edit.
 
 .. warning::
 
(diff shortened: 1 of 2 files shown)


oruebel:
@mavaylon1 is also working on a tutorial for this

bendichter:
Where is it?

docs/gallery/advanced_io/plot_editing.py
```diff
@@ -0,0 +1,178 @@
+"""
+Editing NWB files
+=================
```
oruebel: The title of the tutorial is a bit misleading. "Editing NWB files" suggests that this is a regular process of modifying an existing file, but I think that is not what this tutorial is targeting. It seems this tutorial targets more advanced manual surgeries. For clarity, I think we should distinguish between "Modifying an existing NWB file with PyNWB" and "Manually modifying an NWB file with h5py".

docs/gallery/advanced_io/plot_editing.py
```diff
@@ -0,0 +1,178 @@
...
+
+This tutorial demonstrates how to edit NWB files. How and whether it is possible to edit
+an NWB file depends on the storage backend and the type of edit. Here, we go through the
+common types of edits for HDF5 files. Keep in mind that any edit to an existing NWB file
+make it no longer a valid NWB file. We call this "doing surgery" on the NWB file.
+We highly recommend making a copy before editing and running a validation check on the
+file after editing it. See :ref:`validating`.
```
oruebel:
```suggestion
This tutorial demonstrates how to edit NWB files manually (i.e., without using PyNWB). 
How and whether it is possible to edit an NWB file depends on the storage backend 
and the type of edit. Here, we go through the common types of manual edits for HDF5 files
using  :py:mod:`h5py`. We call this "doing surgery" on the NWB file.

.. warning::

     Manually editing an existing NWB file can make the file invalid if you are not careful. 
     We highly recommend making a copy before editing and running a validation check on the
     file after editing it. See :ref:`validating`.

```

docs/gallery/advanced_io/plot_editing.py
bendichter: Sure, that sounds good. We should add material on how to edit files using PyNWB. I am not sure what the capabilities of that are.

docs/gallery/advanced_io/plot_editing.py
```diff
@@ -0,0 +1,178 @@
...
+with NWBHDF5IO("test_edit.nwb", "w") as io:
+    io.write(nwbfile)
+
+
+##############################################
+# Now, let's try to edit the values of the dataset:
+import h5py
+
+with h5py.File("test_edit.nwb", "r+") as f:
+    f["acquisition"]["synthetic_timeseries"]["data"][:10] = 0.0
```
oruebel: For datasets we can do this PyNWB directly. @rly does changing values also work for attributes? If datasets and attributes can be edited directly from PyNWB, then I think we should make that the recommended way,  rather than h5py.

docs/gallery/advanced_io/plot_editing.py
```diff
@@ -0,0 +1,178 @@
...
+# The ``None`` in ``maxshape`` means that the dataset has an unlimited shape. You can
+# also use an integer to specify a fixed ``maxshape``. If you do not specify a
+# ``maxshape``, then the dataset will have a fixed shape. You can change the shape of
+# this dataset.
+
+import h5py
+
+with h5py.File("test_edit2.nwb", "r+") as f:
+    f["acquisition"]["synthetic_timeseries"]["data"].resize((200, 100))
+
```
oruebel: Same comment as above, I think we could go through PyNWB here, since it exposes the h5py datasets, i.e,. I think the following should work

```
with NWBHDF5IO("test_edit2.nwb", "a") as io:
    nwbfile = io.read()
    nwbfile.get_acquisition("synthetic_timeseries").data.resize((200, 100))
```

docs/gallery/advanced_io/plot_editing.py
```diff
@@ -0,0 +1,178 @@
...
+# Replacing a dataset in h5py
+# ----------------------------
+# There are several types of dataset edits that cannot be done in-place:
+#
+#     * Changing the shape of a dataset with a fixed shape
+#     * Changing the datatype of a dataset
+#     * Changing the compression of a dataset
+#     * Changing the chunking of a dataset
+#     * Changing the max-shape of a dataset
+#     * Changing the fill-value of a dataset
```
oruebel: I would put this text as a note at the top of the tutorial, because it clarifies what changes are not possible.

docs/gallery/advanced_io/plot_editing.py
```diff
@@ -0,0 +1,178 @@
...
+        dtype=np.float64,
+    )
+
+##############################################
+# .. note::
+#   Because of the way HDF5 works, the ``del`` action will not actually free up any
+#   space in the HDF5 file. To free up space in the file, you will need to run the
+#   ``h5repack`` command line tool. See the `h5repack documentation
+#   <https://support.hdfgroup.org/HDF5/doc/RM/Tools.html#Tools-Repack>`_ for more
+#   information.
```
oruebel: `h5repack` just makes a new copy of the file. If we are creating a copy of the file anyways, then I think it may be easier to use the export functionality in PyNWB for this. @rly is this covered in any of the docs on export? I think the steps would be to:

- Open the file in PyNWB
- Replace the TimeSeries
- Use `NWBHDF5IO.export` to export to a new file.

CodyCBakerPhD: I can't recommend strongly enough against `h5repack` - I had to do some file fixes for an old Buzsaki project and the average write speed was ~4 MB/s and has no options to parallelize

You would absolutely get faster speeds by just 'reading' from the source NWB file and 'writing' a brand new one (which I assume is what `export` does)

docs/gallery/advanced_io/plot_editing.py
bendichter: I guess technically you could do e.g.

```python
nwbfile.acquisition["time_series"].data.attrs["unit"] = "volts"
```
but this will not work:
```python
nwbfile.acquisition["time_series"].data.units = "volts"
```

nor 

```python
nwbfile.acquisition["time_series"].units = "volts"
```

docs/gallery/advanced_io/plot_editing.py
bendichter: yes that should work

docs/gallery/advanced_io/plot_editing.py
oruebel: Aside from renaming groups via `h5py.move`, I think the other items that are currently listed here are possible through PyNWB:

- For changes to h5py.datasets (e.g., changing values etc.), the PyNWB exposes the `h5py.Dataset` object directly to the user on read. I.e., if you open with `mode='a'` then any operations you do manually on f["acquisition"]["synthetic_timeseries"]["data"] in the tutorial, you should be able to do with `nwbfile.get_acquisition("synthetic_timeseries").data`. The same kind of risks of breaking things apply, but the access is through PyNWB. 
- For changing attribute values, I'm not 100% sure, but I believe you can change attribute values in PyNWB and then call write to update the file. The difference between attributes and datasets is that the datasets are updated immediately whereas attributes require an additional call to write. This is due to the fact that attributes don't have lazy read.
- For replacing a dataset, I believe we can use `export`. I.e., 1) open the file, 2) remove the `TimeSeries`, 3) add a new `TimeSeries` with the new data, 4) call export to write the updated file to a new location.

docs/gallery/advanced_io/plot_editing.py
bendichter: Huh, good to know

bendichter:
Thanks for the review, @oruebel.  I've made a bunch of these changes. I haven't yet included `export`, since I don't know its functionality as well, but that's the next step.

bendichter:
@oruebel ok, this is ready for another round of review

docs/gallery/general/add_remove_containers.py
```diff
@@ -200,7 +182,7 @@
         export_io.export(src_io=read_io, nwbfile=read_nwbfile)
 
 ###############################################################################
-# More information about export
-# ---------------------------------
 # For more information about the export functionality, see :ref:`export`
 # and the PyNWB documentation for :py:meth:`NWBHDF5IO.export <pynwb.NWBHDF5IO.export>`.
+#
+# For more options about editing a file in place, see :ref:`editing`.
```
bendichter:
```suggestion
# For more information about editing a file in place, see :ref:`editing`.
```

docs/gallery/advanced_io/plot_editing.py
```diff
@@ -0,0 +1,157 @@
...
+##############################################
+# Changing the shape of dataset
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Whether it is possible to change the shape of a dataset depends on how the dataset was
+# created. If the dataset was created with a flexible shape, then it is possible to
+# change in-place. Creating a dataset with a flexible shape is done by specifying the
+# ``maxshape`` argument of the :py:class:`~hdmf.backends.hdf5.h5_utils.H5DataIO` class
+# constructor. Using a ``None`` value for ``maxshape`` allows the dataset to be reset
+# arbitrarily long in that dimension. Chunking is required for datasets with flexible
+# shapes. Setting ``maxshape`` automatically sets chunking to ``True``, if not specified.
```
oruebel:
```suggestion
# Whether it is possible to change the shape of a dataset depends on how the dataset was
# created. If the dataset was created with a flexible shape, then it is possible to
# change in-place. Creating a dataset with a flexible shape is done by specifying the
# ``maxshape`` argument of the :py:class:`~hdmf.backends.hdf5.h5_utils.H5DataIO` class
# constructor. Using a ``None`` value for a component of the ``maxshape`` tuple allows 
# the size of the corresponding dimension to grow, such that is can be be reset arbitrarily long 
# in that dimension. Chunking is required for datasets with flexible shapes. Setting ``maxshape``,
# hence,  automatically sets chunking to ``True``, if not specified.
```

docs/gallery/advanced_io/plot_editing.py
```diff
@@ -0,0 +1,157 @@
...
+)
+
+with NWBHDF5IO("test_edit2.nwb", "w") as io:
+    io.write(nwbfile)
+
+##############################################
+# The ``None`` in ``maxshape`` means that the dataset has an unlimited shape. You can
+# also use an integer to specify a fixed ``maxshape``. If you do not specify a
+# ``maxshape``, then the dataset will have a fixed shape. You can change the shape of
+# this dataset.
```
oruebel:
```suggestion
##############################################
# The ``None``value  in the first component of ``maxshape`` means that the
# the first dimension of the dataset is unlimited. By setting the second dimension
# of ``maxshape`` to ``100``, that dimension is fixed to be no larger than ``100``. 
# If you do not specify a``maxshape``, then the shape of the dataset will be fixed
# to the shape that the dataset was created with. Here, you can change the shape of
# the first dimension of this dataset.
```

docs/gallery/advanced_io/plot_editing.py
```diff
@@ -0,0 +1,157 @@
...
+# .. note::
+#   There are several types of dataset edits that cannot be done in-place: changing the
+#   shape of a dataset with a fixed shape, or changing the datatype, compression,
+#   chunking, max-shape, or fill-value of a dataset. For any of these, we recommend using
+#   the :py:class:`pynwb.NWBHDF5IO.export` method to export the data to a new file. See
+#   :ref:`modifying_data` for more information.
+#
+# Editing groups
+# --------------
+# To edit the attributes of a group, open the file and edit it using :py:mod:`h5py`:
```
oruebel, line 132:
```suggestion
# Editing of groups is not yet supported in PyNWB. 
# To edit the attributes of a group, open the file and edit it using :py:mod:`h5py`:
```

oruebel:
> ok, this is ready for another round of review

Overall this looks good to me. I added a few suggestions for minor edits on text, but nothing critical. Feel free to accept/reject the suggestions as necessary.

@rly do you know whether it is possible to change the values of attributes directly in PyNWB. I vaguely remember that PyNWB/HDMF was tracking changes to attribute values so that one can set the values and then call ``write`` again on the file to update those values. But I'm not 100% sure whether this is working or if that was just something that we were discussing as a potential feature at some point.

rly:
> @rly do you know whether it is possible to change the values of attributes directly in PyNWB. I vaguely remember that PyNWB/HDMF was tracking changes to attribute values so that one can set the values and then call `write` again on the file to update those values. But I'm not 100% sure whether this is working or if that was just something that we were discussing as a potential feature at some point.

This is not allowed because `AbstractContainer` prevents all fields from being changed after they are set. We were discussing removing that restriction and replacing it with warnings in https://github.com/hdmf-dev/hdmf/pull/868 but that work is not complete. I ran a few tests here a few months ago to confirm: https://gist.github.com/rly/be7bee420b482b9ddcd084f57cc4115e

oruebel (approved):

oruebel:
> This is not allowed

Thanks for the clarification. In that case, I think this is good to go.

oruebel:
Thanks @bendichter ! One thing I missed, we should also make a thumbnail for this tutorial. The PowerPoint file with all the Thumbnail sources is in the repo here https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/source/figures/gallery_thumbnails.pptx

PR #1846, opened 2024-02-10
Exclude pynwb retinotopy from sphinx apidoc

rly:
Deprecating`import pynwb.retinotopy` results in an `ImportError` during sphinx docs generation. This PR tries to exclude that module from sphinx processing


rly:
The sphinx ImportError issue has now been resolved. The check-sphinx-links workflow is now failing because of an unrelated SSL certificate issue with incf.org.

oruebel (approved):

PR #1875, opened 2024-03-27
Add mock units table

h-mayorquin:
## Motivation

Add a quick method to create mock units table
Is this something that would be of interest or useful to have?

## How to test the behavior?

Should we have tests for the mock methods?

```python
from pynwb.testing.mock.ecephys import mock_Units


units_table = mock_Units()
units_table.to_dataframe()
```

Produces the following output.

|   id | unit_name   | spike_times                                                                                                                                                                        |
|-----:|:------------|:-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
|    0 | unit_0      | [0.6369616873214543, 0.9067484010853246, 0.9477219250215193]                                                                                                                       |
|    1 | unit_1      | [0.8158535541215322, 0.8185920542916802, 1.6759963308792496, 1.709581906184714, 2.439237352614658]                                                                                 |
|    2 | unit_2      | [0.028319671145462966, 0.1526029476450269, 0.8232273623386572]                                                                                                                     |
|    3 | unit_3      | [0.6884467305709401, 1.0773681545500438, 1.212464659572455]                                                                                                                        |
|    4 | unit_4      | [0.5715298307297609, 0.8933992218057031, 1.4876992520053998, 1.825610477512533, 2.2172294780406943, 3.1075038300454865, 3.3346614235788663, 3.9578485682649087, 4.041863911847294] |
|    5 | unit_5      | [0.7870983074886834, 1.0264677504816355, 1.9029519812923392, 1.9615200160975337, 2.297637076643194, 2.447916543538033, 2.89825591018732, 3.694580180474614]                        |
|    6 | unit_6      | [0.4045518398215282, 0.6030648843307835]                                                                                                                                           |
|    7 | unit_7      | [0.6291081515397092, 1.5562627046075765, 1.9966398593233605, 2.9512303530140978, 3.4511261667017448, 3.8763547915508205, 4.496568243566198, 5.491664748801522, 6.440608423739287]  |
|    8 | unit_8      | [0.7577288453082914, 1.2551515407959104, 1.784463700992681, 2.5702494017064885, 2.9849052510621594, 3.7193888228508887, 4.430531700840638, 5.3625913874540165, 5.477524020734922]  |
|    9 | unit_9      | [0.9274239286245599, 1.8953501185492063, 1.9100564235145756]                                              

## Checklist

### DIFF
--- a/src/pynwb/testing/mock/ecephys.py
+++ b/src/pynwb/testing/mock/ecephys.py
@@ -125,15 +125,13 @@ def mock_SpikeEventSeries(
 def mock_Units(
     num_units: int = 10,
     max_spikes_per_unit: int = 10,
-    seed: Optional[int] = None,
+    seed: int = 0,
     nwbfile: Optional[NWBFile] = None,
 ) -> Units:
 
     units_table = Units()
     units_table.add_column(name="unit_name", description="a readable identifier for the unit")
 
-    if seed is None:
-        seed = 0
     rng = np.random.default_rng(seed=seed)
 
     times = rng.random(size=(num_units, max_spikes_per_unit)).cumsum(axis=1)
--- a/src/pynwb/testing/mock/ecephys.py
+++ b/src/pynwb/testing/mock/ecephys.py
@@ -140,7 +140,7 @@ def mock_Units(num_units: int = 10, max_spikes_per_unit: int = 10, seed: Optiona
     for unit_index in range(num_units):
 
         # Not all units have the same number of spikes
-        spike_times = times[unit_index, : spikes_per_unit[unit_index]]
+        spike_times = times[unit_index, :spikes_per_unit[unit_index]]
         unit_name = f"unit_{unit_index}"
         units_table.add_unit(spike_times=spike_times, unit_name=unit_name)
         
--- a/src/pynwb/testing/mock/ecephys.py
+++ b/src/pynwb/testing/mock/ecephys.py
@@ -122,12 +122,15 @@ def mock_SpikeEventSeries(
     return spike_event_series
 
 
-def mock_Units(num_units: int = 10, max_spikes_per_unit: int = 10, seed: Optional[int] = None, nwbfile: NWBFile):
+def mock_Units(
+    num_units: int = 10,
+    max_spikes_per_unit: int = 10,
+    seed: Optional[int] = None,
+    nwbfile: Optional[NWBFile] = None,
+) -> Units:
 
     units_table = Units()
-    units_table.add_column(
-        name="unit_name", description="a readable identifier for the unit"
-    )
+    units_table.add_column(name="unit_name", description="a readable identifier for the unit")
 
     if seed is None:
         seed = 0
@@ -140,10 +143,10 @@ def mock_Units(num_units: int = 10, max_spikes_per_unit: int = 10, seed: Optiona
     for unit_index in range(num_units):
 
         # Not all units have the same number of spikes
-        spike_times = times[unit_index, :spikes_per_unit[unit_index]]
+        spike_times = times[unit_index, : spikes_per_unit[unit_index]]
         unit_name = f"unit_{unit_index}"
         units_table.add_unit(spike_times=spike_times, unit_name=unit_name)
-        
+
     if nwbfile is not None:
         nwbfile.units = units_table
 
--- a/src/pynwb/testing/mock/ecephys.py
+++ b/src/pynwb/testing/mock/ecephys.py
@@ -133,7 +133,7 @@ def mock_Units(num_units: int = 10, max_spikes: int = 10, seed: Optional[int] = 
     )
 
     if seed is None:
-        seed = np.pi
+        seed = 0
     rng = np.random.default_rng(seed=seed)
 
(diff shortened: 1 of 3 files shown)


oruebel:
> Should we have tests for the mock methods?

Since `pynwb.testing` is part of the main `pynwb` package (not just the test harness), I do think that having at least a basic set of unit tests that generate a mock instance and then check that the contents are as expected would be very useful, in particular since downstream testing depends on it. Also, to keep our test coverage up.

oruebel:
> Should we have tests for the mock methods?

Correction of my previous comment. Tests for the mock classes are here https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/tests/unit/test_mock.py

h-mayorquin:
Thanks @oruebel. I will take care of this.

src/pynwb/testing/mock/ecephys.py
```diff
@@ -119,3 +120,28 @@ def mock_SpikeEventSeries(
...
+    rng = np.random.default_rng(seed=seed)
+
+    times = rng.random(size=(num_units, max_spikes_per_unit)).cumsum(axis=1)
+    spikes_per_unit = rng.integers(1, max_spikes_per_unit, size=num_units)
+
+    spike_times = []
+    for unit_index in range(num_units):
+
+        # Not all units have the same number of spikes
+        spike_times = times[unit_index, : spikes_per_unit[unit_index]]
```
bendichter, line 144:
```suggestion
        spike_times = times[unit_index, :spikes_per_unit[unit_index]]
```

src/pynwb/testing/mock/ecephys.py
```diff
@@ -119,3 +120,28 @@ def mock_SpikeEventSeries(
...
+        unit_name = f"unit_{unit_index}"
+        units_table.add_unit(spike_times=spike_times, unit_name=unit_name)
+
+    return units_table
```
bendichter:
```suggestion
def mock_Units(num_units: int = 10, max_spikes_per_unit: int = 10, seed: Optional[int] = None, nwbfile: NWBFile):

    units_table = Units()
    units_table.add_column(
        name="unit_name", description="a readable identifier for the unit"
    )

    if seed is None:
        seed = 0
    rng = np.random.default_rng(seed=seed)

    times = rng.random(size=(num_units, max_spikes_per_unit)).cumsum(axis=1)
    spikes_per_unit = rng.integers(1, max_spikes_per_unit, size=num_units)

    spike_times = []
    for unit_index in range(num_units):

        # Not all units have the same number of spikes
        spike_times = times[unit_index, : spikes_per_unit[unit_index]]
        unit_name = f"unit_{unit_index}"
        units_table.add_unit(spike_times=spike_times, unit_name=unit_name)
        
    if nwbfile is not None:
        nwbfile.units = units_table

    return units_table
```

h-mayorquin: Ah, OK, this is the convention for the other objects as well.

src/pynwb/testing/mock/ecephys.py
```diff
@@ -119,3 +120,34 @@ def mock_SpikeEventSeries(
...
+    max_spikes_per_unit: int = 10,
+    seed: Optional[int] = None,
+    nwbfile: Optional[NWBFile] = None,
+) -> Units:
+
+    units_table = Units()
+    units_table.add_column(name="unit_name", description="a readable identifier for the unit")
+
+    if seed is None:
+        seed = 0
```
bendichter:
```suggestion
    seed: int = 0,
    nwbfile: Optional[NWBFile] = None,
) -> Units:

    units_table = Units()
    units_table.add_column(name="unit_name", description="a readable identifier for the unit")

```

h-mayorquin: Not that it matters much in this case but I am following numpy convention for default seeds.

https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.default_rng

I think the semantics are better and leaves the possibility of changing the initialization if the seed is not appropriate.

bendichter: `None` makes sense in their case because they are not using the 0 seed:

> If None, then fresh, unpredictable entropy will be pulled from the OS.

h-mayorquin: Good point. 

Do you think passing the seed as None directly would be better or do you prefer fix by default as you are proposing in your change?

bendichter: This library will be used mostly for testing, so it may make sense to default to 0 for reproducibility. I think defaulting to None and passing that directly would be fine too.

h-mayorquin: Let's go with your proposed change then. Thanks for the feedback.

tests/unit/test_mock.py
```diff
@@ -119,3 +121,5 @@ def test_name_generator():
 
     assert name_generator("TimeSeries") == "TimeSeries"
     assert name_generator("TimeSeries") == "TimeSeries2"
+
```
bendichter, line 124:
```suggestion
```

bendichter (approved):

rly:
Thank you for the contribution, @h-mayorquin !

PR #1877, opened 2024-03-31
update style for _add_missing_timezone

bendichter:
## Motivation

support date and datetime


rly:
This function was removed in #1886.

Issue #1882, opened 2024-04-01
What should be default  name of `pynwb.misc.Units`?


h-mayorquin:
Right now the following throws an error:

```python
from pynwb.misc import Units
from pynwb.testing.mock.file import mock_NWBFile


nwbfile = mock_NWBFile()
nwbfile.units = Units()


ValueError: Field 'units' on NWBFile must be named 'units'."
```

Units default name is "Units" as supported by [best practices for naming conventions](https://nwbinspector.readthedocs.io/en/dev/best_practices/general.html#naming-conventions):

```
assert Units().name == "Units"
```

And is defined here:
https://github.com/NeurodataWithoutBorders/pynwb/blob/2259bede338f2f202229bda0af15d7e3cea47369/src/pynwb/misc.py#L157-L158

However, the `NWBFile` has "units" as a required name for the attribute here on its fields:

https://github.com/NeurodataWithoutBorders/pynwb/blob/2259bede338f2f202229bda0af15d7e3cea47369/src/pynwb/file.py#L272-L273    
 
So I think that one of the two should give? Which one makes more sense? Maybe the nwbfile.units should accept "Units" as the name? Should it be the other way around?

rly:
The name of the `Units` object stored at the `NWBFile` level of the file must be `"units"` according to the schema:
https://github.com/NeurodataWithoutBorders/nwb-schema/blob/dev/core/nwb.file.yaml#L446-L449

This predates the best practices. Changing that to `"Units"` at the schema level would make existing NWB 2.0-2.7 files invalid to the 2.8+ schema, which is not ideal. We could modify the APIs to set `NWBFile.units` from the HDF5 group at `/Units` instead of `/units`. However, that might break existing software that does not use the APIs to read the units table (Neurosift *may* be one). It would also mean that the name of that group in HDF5 would be the only group that starts with a capital letter -- visually unappealing, but not a big deal. 

Alternatively, we could remove the fixed name `"units"` and modify the APIs to set `NWBFile.units` to the only `Units` object in the root group, if present, whatever it is named. That would allow for heterogeneity in what the root file looks like when >95% of use cases will need only a single units table, and would also break existing software that does not use the APIs.

Anyway, most people don't care what the object is named under the hood, but I hesitate to change the current naming scheme because of other software relying on the existing schema.

[...]

PR #1883, opened 2024-04-02
Add default `name="units"` to `Units` object in the `mock_Units` testing function

h-mayorquin:
Related to this discussion here:
https://github.com/NeurodataWithoutBorders/pynwb/issues/1882

This is a quality of life small change that would allow two things to work:
* This part of the code:
https://github.com/h-mayorquin/pynwb/blob/ad6b0283c49ce8c391fd1b0e6ced3d0f6907c17d/src/pynwb/testing/mock/ecephys.py#L148-L150
* The following script for testing:

```python
from pynwb.testing.mock.ecephys import mock_Units

nwbfile = mock_NWBFile()
units_table = mock_Units(num_units=5)
nwbfile.units = units_table
```

The only downside is that for `Units` object that are destined for some other places of the nwbfile (.e.g processing) this will not follow the best practices.  As these are functions that are mainly used for testing I think that this is not a big deal.


rly (approved):

rly:
Looks good. Thanks @h-mayorquin

PR #1892, opened 2024-04-19
Add `grid_spacing` and `origin_coords` to the fields of `ImagingPlane`

### LINKED ISSUE #1885: Question: why is `grid_spacing` not a field in `ImagingPlane`?
I would like to understand the logic of why some attributes of the class are fields and others are not. For the `ImagingPlane` we are missing `grid_spacing` and `origin_coords`

https://github.com/NeurodataWithoutBorders/pynwb/blob/2259bede338f2f202229bda0af15d7e3cea47369/src/pynwb/ophys.py#L33-L46

This is related to the html representation of the object. We use the fields to generate the html representation (that is very handy for exploring) and therefore some attributes are not displayed because they are not fields which I think is a big miss.

https://github.com/hdmf-dev/hdmf/blob/d85d0cbc36d2e0fdb25e8fbea14d58ba7bf24a40/src/hdmf/container.py#L685-L693

h-mayorquin:
## Motivation

This should fix #1885 

I discussed this with @rly and it seems that it was just an omission.  This PR tries to address this. So far, this seems to work for `grid_spacing` but for some reason that escapes me it seems that `origin_coords` is not added to the fields. 

image

```python
from pynwb.testing.mock.ophys import mock_ImagingPlane

imaging_plane = mock_ImagingPlane(grid_spacing=[1.0, 1.0], origin_coords=[0.0, 0.0])
imaging_plane.origin_coords
list(imaging_plane.fields.keys())

```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
   - Support `stimulus_template` as optional predefined column in `IntracellularStimuliTable`. @stephprince [#1815](https://github.com/NeurodataWithoutBorders/pynwb/pull/1815)
   - Support `NWBDataInterface` and `DynamicTable` in `NWBFile.stimulus`. @rly [#1842](https://github.com/NeurodataWithoutBorders/pynwb/pull/1842)
 - Added support for python 3.12 and upgraded dependency versions. This also includes infrastructure updates for developers. @mavaylon1 [#1853](https://github.com/NeurodataWithoutBorders/pynwb/pull/1853)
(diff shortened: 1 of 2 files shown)


rly:
> So far, this seems to work for grid_spacing but for some reason that escapes me it seems that origin_coords is not added to the fields.

@h-mayorquin I think you were missing a comma in the `__nwbfields__` list. I fixed that and I think it works now. Could you check, and if it looks good, add a changlog?

h-mayorquin:
> > So far, this seems to work for grid_spacing but for some reason that escapes me it seems that origin_coords is not added to the fields.
> 
> @h-mayorquin I think you were missing a comma in the `__nwbfields__` list. I fixed that and I think it works now. Could you check, and if it looks good, add a changlog?

My bad, I tested your change and it is working now. Thanks.

Added the changelog.

rly (approved):

rly:
Thanks for the fix @h-mayorquin !

PR #1894, opened 2024-04-25
Avoid having to set `rate=None` explitctly when passing timestamps in  `mock_ElectricalSeries`

h-mayorquin:
## Motivation

Right now you get the following error when trying to pass timestamps to the `mock_ElectricalSeries` function. 

```python
from pynwb.testing.mock.ecephys import mock_ElectricalSeries


mock_ElectricalSeries(timestamps=np.arange(10))
ValueError: Specifying rate and timestamps is not supported.
```

This is because we have `rate=30,000.0` as a default value in the function signature. To make this work, the user has to explictly set rate=None, that is:

```python
from pynwb.testing.mock.ecephys import mock_ElectricalSeries


mock_ElectricalSeries(timestamps=np.arange(10), rate=None)
```

This is too cumbersome. This PR simplifies this so this is not necessary. To do this, it moves the logic of default values inside the function.

I also added some missing typing.

## How to test the behavior?
The code above should work.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # PyNWB Changelog
 
+## PyNWB 2.8.0 (Upcoming)
+
+### Enhancements and minor changes
+- Set rate default value inside `mock_ElectricalSeries` to avoid having to set `rate=None` explicitly when passing timestamps. @h-mayorquin [#1894](https://github.com/NeurodataWithoutBorders/pynwb/pull/1894)
+
 ## PyNWB 2.7.0 (May 2, 2024)
 
 ### Enhancements and minor changes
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
   - Modified `OptogeneticSeries` to allow 2D data, primarily in extensions of `OptogeneticSeries`. @rly [#1812](https://github.com/NeurodataWithoutBorders/pynwb/pull/1812)
   - Support `stimulus_template` as optional predefined column in `IntracellularStimuliTable`. @stephprince [#1815](https://github.com/NeurodataWithoutBorders/pynwb/pull/1815)
   - Support `NWBDataInterface` and `DynamicTable` in `NWBFile.stimulus`. @rly [#1842](https://github.com/NeurodataWithoutBorders/pynwb/pull/1842)
+- Set rate default value inside `mock_ElectricalSeries` to avoid having to set `rate=None` explitctly when passing timestamps. @h-mayorquin [#1894](https://github.com/NeurodataWithoutBorders/pynwb/pull/1894)
```
stephprince:
```suggestion
- Set rate default value inside `mock_ElectricalSeries` to avoid having to set `rate=None` explicitly when passing timestamps. @h-mayorquin [#1894](https://github.com/NeurodataWithoutBorders/pynwb/pull/1894)
```

src/pynwb/testing/mock/ecephys.py
```diff
@@ -69,9 +69,9 @@ def mock_electrodes(
 def mock_ElectricalSeries(
     name: Optional[str] = None,
     description: str = "description",
-    data=None,
-    rate: float = 30000.0,
-    timestamps=None,
+    data: Optional[np.ndarray] = None,
+    rate: Optional[float] = None,
+    timestamps: Optional[np.ndarray] = None,
```
stephprince:
```suggestion
    data=None,
    rate: Optional[float] = None,
    timestamps=None,
```

I suggest not adding typing for data and timestamps here since these `ElectricalSeries` arguments can be other array data types (h5py Dataset, Zarr array, TimeSeries, etc.) and so it might be misleading to specify only numpy arrays for the `mock_ElectricalSeries`.

stephprince (commented):
Sorry for the delayed review! Added a few small suggestions but otherwise looks good to me.

src/pynwb/testing/mock/ecephys.py
h-mayorquin: Makes sense.

stephprince (approved):

stephprince:
The failing checks are due to a separate hdmf-zarr issue https://github.com/hdmf-dev/hdmf-zarr/issues/192 and are unrelated to these changes to `mock_ElectricalSeries`. So we will ignore them for now and go ahead and merge this PR.

PR #1902, opened 2024-05-13
Exclude artifacts from sdist and wheel

penguinpee:
Building the wheel from the PyPI sdist will include those files and
directories otherwise. They are also distributed in the uploaded wheel.


rly:
Thanks for the report @penguinpee , I'll take a look at this before the next release.

mavaylon1 (approved):

rly:
Some tests are failing due to unrelated issues related to the Zarr tutorial and a recent Zarr update.

PR #1907, opened 2024-05-22
Add "bounds" to SpatialSeries fields

rly:
## Motivation

I noticed that pynwb tests were passing with NWB schema 2.7.0 even though "bounds" was not added to SpatialSeries until after NWB schema 2.7.0, so really the IO tests should fail. 

Until then, on the dev branch, "bounds" in `SpatialSeries.__init__` will not be written to disk because it is missing from `__nwbfields__`.

Using the convenience test classes can hide this error. I think we should be more explicit when writing roundtrip tests to ensure that all expected fields are read correctly from the written file.

This PR merges into a new `nwb_schema_2.8.0` branch.

## Checklist


stephprince (approved):
looks good to me!

PR #1911, opened 2024-05-29
Upgrade validation methods

### LINKED ISSUE #1179: validator should be able to output machine readable results
Could be JSON using `--json` in a format like

```json
{
    "exitcode" : 0,
    "issues" : [ { "description": "", "location":"", "severity": 10}, ]
}
```

### LINKED ISSUE #1530: [Documentation]: Link to NWBInspector from the validator

The PyNWB validator checks for structural compliance of NWB files. In practice, however, most users should validate both structural compliance  with the NWB schema as well as compliance of the data itself with NWB best practice. Users should, hence, be pointed to the [NWBInspector](https://github.com/NeurodataWithoutBorders/nwbinspector), which runs both the PyNWB validator and additional data checks. To do this we should:

1) Update the tutorial on [Validating NWB Files](https://pynwb.readthedocs.io/en/stable/validation.html) to point to the nwbinspector
2) Add a note to the the [validator command line](https://github.com/NeurodataWithoutBorders/pynwb/blob/2de33c032eebc46800191bc24a72976bca30aa6f/src/pynwb/validate.py#L30-L31) to to indicate that use of the NWBInspector is recommended
3) Add a not to the docstring of the [validate method](https://github.com/NeurodataWithoutBorders/pynwb/blob/2de33c032eebc46800191bc24a72976bca30aa6f/src/pynwb/__init__.py#L193) to describe that the function only checks for compliance with the NWB schema and that users should use the NWBInspector for additional validation with NWB best practices. 

Yes.

### Code of Conduct

### LINKED ISSUE #1756: [Feature]: entry point for validate

The current CLI syntax for pynwb validate according to the docs is:

```bash
python -m pynwb.validate test.nwb
```

I think it would be better if it were something like 

```python
pynwb validate test.nwb
```

no

see above.

No.

### Code of Conduct

### LINKED ISSUE #1807: [Feature]: Change pynwb.validate(io=...) to match pynwb.validate(paths=[path])

From https://github.com/NeurodataWithoutBorders/pynwb/issues/1777 and follow-up to https://github.com/NeurodataWithoutBorders/pynwb/pull/1511

The behavior of the `pynwb.validate` function with `paths` kwarg, `pynwb.validate` function with `io` kwarg, and `pynwb.validate` CLI are inconsistent and therefore confusing. We added the first option while keeping the other two to maintain backwards compatibility, but this is causing issues downstream. These methods should all do the same basic validation behavior with the same argument names, except that the `pynwb.validate` function with paths kwargs and CLI should open each passed file using `NWBHDF5IO` (and later using the generic read function that will be added, to account for Zarr files) and then call `pynwb.validate` on each io object.

This is a breaking change requiring a major version bump.

_No response_

Desired changed features:

The core `validate` function should take as input an `HDMFIO` object and return a list of machine-readable JSON validation results (#1179).
- It validates a file against **all namespaces together** (requires https://github.com/hdmf-dev/hdmf/issues/608)
- You cannot validate against a single namespace in isolation because it doesn't really make sense. The file may contain objects of a neurodata type from a non-dependent namespace, which would not be interpretable. e.g., validating against "core" only, when the file contains a `Device` subtype defined in an extension, and when the `Device` is required by another type, results in an error because the `Device` subtype is not recognized (see #1777). 
- You can validate against either all namespaces together or all namespaces loaded in the global type map.

There should be a separate function and CLI to get a list of cached namespaces and their dependencies of a file.

Yes.

### Code of Conduct

### LINKED ISSUE #1953: [Bug]: Python warning when executing validate
### What happened?

I would expect no python warning from the below code:

### Steps to Reproduce

Input:
```python
python -m pynwb.validate --help
```

Output:
```
<frozen runpy>:128: RuntimeWarning: 'pynwb.validate' found in sys.modules after import of package 'pynwb', but prior to execution of 'pynwb.validate'; this may result in unpredictable behaviour
usage: validate.py [-h] [-lns] [-n NS] [--no-cached-namespace] paths [paths ...]

Validate an NWB file

positional arguments:
  paths                 NWB file paths

options:
  -h, --help            show this help message and exit
  -lns, --list-namespaces
                        List the available namespaces and exit.
  -n NS, --ns NS        the namespace to validate against
  --no-cached-namespace
                        Use the PyNWB loaded namespace (true) or use the cached namespace (false; default).

If --ns is not specified, validate against all namespaces in the NWB file.
```
```


### Traceback

_No response_

### Operating System

Linux

### Python Executable

Python

### Python Version

3.11

### Package Versions

```
$pip list
Package                   Version
------------------------- -----------
attrs                     24.2.0
h5py                      3.11.0
hdmf                      3.14.3
jsonschema                4.23.0
jsonschema-specifications 2023.12.1
numpy                     1.26.4
pandas                    2.2.2
pip                       23.0.1
pynwb                     2.8.1
python-dateutil           2.9.0.post0
pytz                      2024.1
referencing               0.35.1
rpds-py                   0.20.0
ruamel.yaml               0.18.6
ruamel.yaml.clib          0.2.8
scipy                     1.14.1
setuptools                66.1.1
six                       1.16.0
tzdata                    2024.1
```

### Code of Conduct

### PR DESCRIPTION
**Pull request by stephprince**
## Motivation

Addresses several issues summarized in #1808. This is a breaking change for the next major release.
* fix #1807 so that `validate` returns the same output whether `paths` or `io` is provided
* fix #1756 by adding a `pynwb-validate` entry point 
* fix #1179  by adding the option to output the validator output to a json file
* fix #1953 by moving the CLI to a separate file, allowing the user to run `python -m pynwb.validation_cli "test.nwb"` as an alternative to the entry point
* fix #1530 by updating the documentation and adding notes to the `validate` and `validation cli` docstrings to point users to NWB inspector
* partially address #1493 by moving `pynwb.validate` into `pynwb.validation` module so that `validate` is the function name and `validation` is the module name

This PR also modifies the validate method so that it 
1. accepts a single path as input (the CLI still accepts multiple paths)
2. no longer returns a status code but will return errors if the function fails in the process of performing validation. (The CLI still returns an exit code).

TODO

- [X] test with nwbinspector (as of now should only require changes to [this line](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/165c7270b13dd832ce7f9080200c1663c8cb7f4e/src/nwbinspector/_nwb_inspection.py#L280))
- [x] check test coverage
- [x] ~~finish testing with `ZarrIO` with file path validation~~ (bump to later PR) 
- [ ] when publishing the next release on conda-forge, the recipe file should also be updated to add the pynwb-validate entry point

## Checklist

### CODE REVIEW

File: src/pynwb/validate.py [hunk 1]
```diff
@@ -58,12 +61,15 @@ def _get_cached_namespaces_to_validate(
...
     )
-    namespace_dependencies = NWBHDF5IO.load_namespaces(
-        namespace_catalog=catalog,
-        path=path,
-        driver=driver,
-        aws_region=aws_region
-    )
+
+    if io is not None:
+        namespace_dependencies = io.load_namespaces(namespace_catalog=catalog,
```
stephprince: I wasn't sure if there is a better way to load the namespaces from the io object, I realize this is HDF5 specific, but it looks like the original namespace loading was as well? I was wondering how this works for Zarr validation?

tests/validation/test_validate.py
```diff
@@ -280,16 +280,55 @@ def test_validate_file_cached_no_cache_bad_ns(self):
...
+                        fake_err.getvalue(),
+                        (
+                            "The namespace 'hdmf-common' is included by the namespace 'core'. "
+                            "Please validate against that namespace instead.\n"
+                        )
+                    )
+                    self.assertEqual(fake_out.getvalue(), "")
+
+    def test_validate_io_and_path_same(self):
+        """Test that validating a file with an io object and a path return the same results."""
```
stephprince, line 283: I'm not sure if all of these are necessary to include, but wanted to double check they were all equivalent

src/pynwb/validate.py
```diff
@@ -137,65 +205,34 @@ def validate(**kwargs):
...
-                print(
-                    f"The namespace '{namespace}' could not be found in {namespace_message} as only "
-                    f"{namespaces_to_validate} is present.",
-                    file=sys.stderr,
-                )
+            status = status or path_status
+            if path_status == 1:
+                continue
 
-        if status == 1:
```
stephprince, line 195: from what I could tell, this status check was never reset to 0, so if any of the paths return a nonzero status, `_validate_helper` would not be run on any subsequent paths. Is that the intended behavior? Or maybe I am misunderstanding something.

src/pynwb/__init__.py
```diff
@@ -354,6 +354,24 @@ def get_sum(self, a, b):
...
+    backend_options = [b for b in backend_io_classes if b.can_read(path=path)]
+    if len(backend_options) == 0:
+        warn(f"Could not find an IO to read the file '{path}'." 
+             f"This may be due to an older file version or invalid file." 
+             f"Defaulting to NWBHDF5IO.", UserWarning)
+        return NWBHDF5IO
+    elif len(backend_options) > 1:
+        raise ValueError(f"Multiple backends found for file '{path}': {backend_options}")
+    else:
+        return backend_options[0]
```
stephprince: I have this temporary private method to start testing support for validation of Zarr NWB files as well. This can eventually be replaced by methods from the `read_nwb` PR

stephprince:
@rly this should be ready for review

There are a couple remaining TODO items related to validation support for NWBZarr files using the file path that could potentially be bumped to a separate PR:
 1. I started to try to implement support for NWBZarr file validation by allowing the `ZarrIO.load_namespaces` method to accept a `ZarrIO.file` object as input, but ran into this issue - https://github.com/hdmf-dev/hdmf-zarr/issues/243. 
 2. We would also want to either rename the `ZarrIO.file` or `HDF5IO._file` property for consistent naming across backends: https://github.com/NeurodataWithoutBorders/pynwb/blob/11cc2328130f9420359282cee4657536415a449b/src/pynwb/validation.py#L58-L61

CHANGELOG.md
```diff
@@ -2,13 +2,22 @@
 
 ## PyNWB 3.0.0 (Upcoming)
 
+### Breaking changes
+- The validation methods have been updated with multiple breaking changes. @stephprince [#1911](https://github.com/NeurodataWithoutBorders/pynwb/pull/1911)
+   - The behavior of `pynwb.validate(io=...)` now matches the behavior of `pynwb.validate(path=...)`. In previous pynwb versions, `pynwb.validate(io=...)` did not use the cached namespaces during validation. To obtain the same behavior as in previous versions, you can update the function call to `pynwb.validate(io=..., used_cached_namespaces=False)`
```
rly:
```suggestion
   - The behavior of `pynwb.validate(io=...)` now matches the behavior of `pynwb.validate(path=...)`. In previous pynwb versions, `pynwb.validate(io=...)` did not use the cached namespaces during validation. To obtain the same behavior as in previous versions, you can update the function call to `pynwb.validate(io=..., use_cached_namespaces=False)`
```

CHANGELOG.md
```diff
@@ -2,13 +2,22 @@
 
 ## PyNWB 3.0.0 (Upcoming)
 
+### Breaking changes
+- The validation methods have been updated with multiple breaking changes. @stephprince [#1911](https://github.com/NeurodataWithoutBorders/pynwb/pull/1911)
+   - The behavior of `pynwb.validate(io=...)` now matches the behavior of `pynwb.validate(path=...)`. In previous pynwb versions, `pynwb.validate(io=...)` did not use the cached namespaces during validation. To obtain the same behavior as in previous versions, you can update the function call to `pynwb.validate(io=..., use_cached_namespaces=False)`
+   - The validate module has been renamed to `validation.py`. The validate method can be 
+   imported using `import pynwb; pynwb.validate` or `from pynwb.validation import validate`
```
rly: I think we should not suggest importing directly from the module file. It gives us less flexibility to refactor things.

```suggestion
   - The validate module has been renamed to `validation.py`. The validate method can be 
   imported using `import pynwb; pynwb.validate` or `from pynwb import validate`
```

docs/source/validation.rst
```diff
@@ -29,35 +38,36 @@ within the ``test.nwb`` file.
...
-    -n NS, --ns NS        the namespace to validate against
     -lns, --list-namespaces
                           List the available namespaces and exit.
-    --cached-namespace    Use the cached namespace (default).
+    -n NS, --ns NS        the namespace to validate against
+    --json-file-path JSON_FILE_PATH
+                          Write json output to this location.
     --no-cached-namespace
-                          Don't use the cached namespace.
+                          Use the PyNWB loaded namespace (true) or use the cached namespace (false; default).
```
rly:
```suggestion
                          Use the namespaces installed by PyNWB (true) or use the cached namespaces (false; default).
```

src/pynwb/__init__.py
```diff
@@ -350,6 +350,27 @@ def get_sum(self, a, b):
     return __TYPE_MAP.get_dt_container_cls(neurodata_type, namespace)
 
 
+@docval({'name': 'path', 'type': str, 'doc': 'the neurodata_type to get the NWBContainer class for'},
```
```suggestion
@docval({'name': 'path', 'type': str, 'doc': 'Path to the NWB file which can be an HDF5 file or a Zarr store.'},
```

docs/source/validation.rst
```diff
@@ -29,35 +38,36 @@ within the ``test.nwb`` file.
...
 
-  optional arguments:
+  options:
     -h, --help            show this help message and exit
-    -n NS, --ns NS        the namespace to validate against
     -lns, --list-namespaces
                           List the available namespaces and exit.
-    --cached-namespace    Use the cached namespace (default).
+    -n NS, --ns NS        the namespace to validate against
+    --json-file-path JSON_FILE_PATH
```
rly:
```suggestion
    --json-output-path JSON_OUTPUT_PATH
```

I think `--json-output-path` might be more clear what this argument is for someone looking at the arguments quickly.

tests/integration/ros3/test_ros3.py
```diff
@@ -90,10 +90,9 @@ def test_dandi_get_cached_namespaces(self):
         )
 
         self.assertCountEqual(first=found_namespaces, second=expected_namespaces)
-        self.assertDictEqual(d1=expected_namespace_dependencies, d2=expected_namespace_dependencies)
+        self.assertDictEqual(d1=found_namespace_dependencies, d2=expected_namespace_dependencies)
```
rly, line 93: Nice catch!

test.py
```diff
@@ -169,7 +169,7 @@ def validate_nwbs():
                 is_family_nwb_file = False
                 try:
                     with pynwb.NWBHDF5IO(nwb, mode='r') as io:
-                        errors = pynwb.validate(io)
+                        errors = validate(io, use_cached_namespaces=False)  # previously io did not validate against cached namespaces
```
rly:
```suggestion
                        errors = validate(io, use_cached_namespaces=False),
                        errors.append(validate(io, use_cached_namespaces=True))
```
1. I think the original comment here might not make sense in the future
2. The example NWB files are all generated using the version of NWB being tested, and because pynwb caches the spec by default, there should be no difference between using cached namespaces and not. 
3. Since the `pynwb.validate` and `pynwb-validate` should be the same now, we don't really need this test anymore since we have the `pynwb-validate` test below. But since this `validate_nwbs()` function is super conservative in its testing of every combination, then for consistency, I suggest we validate with both `use_cached_namespaces=True` and `use_cached_namespaces=False`.

test.py
```diff
@@ -179,7 +179,7 @@ def validate_nwbs():
                         memb_size = 1024**2  # note: the memb_size must be the same as the one used to create the file
                         with h5py.File(filename_pattern, mode='r', driver='family', memb_size=memb_size) as f:
                             with pynwb.NWBHDF5IO(file=f, manager=None, mode='r') as io:
-                                errors = pynwb.validate(io)
+                                errors = validate(io, use_cached_namespaces=False)
```
rly, line 182:
```suggestion
                                errors = validate(io, use_cached_namespaces=False)
                                errors.append(validate(io, use_cached_namespaces=True))
```

docs/source/validation.rst
stephprince: I'm good updating for clarity! I used ``json-file-path`` originally since that is the same flag used by the nwbinspector CLI, but I don't think consistency across those CLIs is strictly necessary.

test.py
stephprince: If I set `use_cached_namespaces=True` here, I get several errors when running the `validate_nwb()` section of `test.py`. I had added `use_cached_namespaces=False` so that it matched the previous test behavior, but if it is expected that there should be no difference between using the cached namespaces and not in this particular case, then maybe these errors are indicative of another issue? 

You can replicate by running `test.py`, but the errors look like this below. Maybe the mylab extension generation needs to be updated?:

```python
2024-12-23 10:12:02,969 - INFO - Validating with pynwb.validate method.
  File "/Users/smprince/anaconda3/envs/pynwb/lib/python3.11/site-packages/hdmf/validate/validator.py", line 280, in get_validator
    return self.__validators[dt]
           ~~~~~~~~~~~~~~~~~^^^^
KeyError: 'NWBFile'

During handling of the above exception, another exception occurred:

  File "/Users/smprince/Documents/code/pynwb/test.py", line 172, in validate_nwbs
    errors = validate(io, use_cached_namespaces=True)  # previously io did not validate against cached namespaces
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/smprince/anaconda3/envs/pynwb/lib/python3.11/site-packages/hdmf/utils.py", line 672, in func_call
    return func(**pargs)
           ^^^^^^^^^^^^^
  File "/Users/smprince/Documents/code/pynwb/src/pynwb/validation.py", line 191, in validate
    validation_errors += _validate_helper(io=io, namespace=validation_namespace)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/smprince/Documents/code/pynwb/src/pynwb/validation.py", line 19, in _validate_helper
    return validator.validate(builder)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/smprince/anaconda3/envs/pynwb/lib/python3.11/site-packages/hdmf/utils.py", line 668, in func_call
    return func(args[0], **pargs)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/smprince/anaconda3/envs/pynwb/lib/python3.11/site-packages/hdmf/validate/validator.py", line 298, in validate
    validator = self.get_validator(dt)
                ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/smprince/anaconda3/envs/pynwb/lib/python3.11/site-packages/hdmf/utils.py", line 668, in func_call
    return func(args[0], **pargs)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/smprince/anaconda3/envs/pynwb/lib/python3.11/site-packages/hdmf/validate/validator.py", line 283, in get_validator
    raise ValueError(msg)
ValueError: data type 'NWBFile' not found in namespace mylab
```

stephprince: I think these lines: https://github.com/NeurodataWithoutBorders/pynwb/blob/80fc386c20d791208142f97e8cfacbce967b8aee/docs/gallery/general/extensions.py#L41 

https://github.com/NeurodataWithoutBorders/pynwb/blob/80fc386c20d791208142f97e8cfacbce967b8aee/docs/gallery/general/extensions.py#L267

could be updated to:
`ns_builder.include_namespace("core")`

to fix these errors. I think these changes more closely match the latest version of ndx-template [`create_extension_spec.py`](https://github.com/nwb-extensions/ndx-template/blob/c055c3ff5c26db418d02120edbfc7c8c29923c61/%7B%7B%20cookiecutter.namespace%20%7D%7D/src/spec/create_extension_spec.py) file. However, should `ns_builder.include_type` still work without including the entire namespace?

rly: Interesting. For now, let's update that to `ns_builder.include_namespace("core")`.

We need to think through what it means to run the pynwb validator on a file, validating against a particular non-core namespace. Should the core namespace always be included during validation, regardless of whether the extension includes the core namespace, and the choices are either use the core namespace cached in the file or the one installed with pynwb? I think so...

I think we eventually want to move toward not even allowing validation against a particular namespace. Either the file is valid given its cached namespaces (or the namespaces installed by pynwb and loaded by the user), or the file is not. Otherwise, we run into weird issues with such as:
https://github.com/hdmf-dev/hdmf/issues/608 and https://github.com/hdmf-dev/hdmf/issues/525

I would say let's make the above change for now, and iterate on these other ideas in a separate PR which does not need to make it in pynwb 3.0.0rc1.

src/pynwb/validation.py
```diff
@@ -0,0 +1,197 @@
...
+        "default": False,
+    },
+    {
+        "name": "driver",
+        "type": str,
+        "doc": "Driver for h5py to use when opening the HDF5 file.",
+        "default": None,
+    }, 
+    returns="Validation errors in the file.",
+    rtype=(list, (list, bool)),
```
rly:
```suggestion
    rtype=list,
```

src/pynwb/validation.py
```diff
@@ -0,0 +1,197 @@
...
+    is_method=False,
```
rly, line 126:
```suggestion
    is_method=False,
    allow_positional=AllowPositional.ERROR,
```
And
```
from hdmf.utils import AllowPositional
```
above.

I think we should require keyword arguments. What do you think? It's hard to make this change later.

rly: We could also set it to warn first:
```
    allow_positional=AllowPositional.WARNING,
```

CHANGELOG.md
```diff
@@ -2,13 +2,22 @@
 
 ## PyNWB 3.0.0 (Upcoming)
 
+### Breaking changes
+- The validation methods have been updated with multiple breaking changes. @stephprince [#1911](https://github.com/NeurodataWithoutBorders/pynwb/pull/1911)
+   - The behavior of `pynwb.validate(io=...)` now matches the behavior of `pynwb.validate(path=...)`. In previous pynwb versions, `pynwb.validate(io=...)` did not use the cached namespaces during validation. To obtain the same behavior as in previous versions, you can update the function call to `pynwb.validate(io=..., use_cached_namespaces=False)`
+   - The validate module has been renamed to `validation.py`. The validate method can be
```
rly, line 9:
```suggestion
   - `pynwb.validate` will return only a list of validation errors instead of a tuple: (list of validation_errors, status code)
   - The validate module has been renamed to `validation.py`. The validate method can be 
```

stephprince: thanks for catching that!

src/pynwb/validation.py
stephprince: Agreed, I think that's a good idea. We can warn first to be consistent with all the other functions we're updating to warn in this release.

test.py
stephprince: > Should the core namespace always be included during validation, regardless of whether the extension includes the core namespace, and the choices are either use the core namespace cached in the file or the one installed with pynwb? I think so...

I think so too?  But agreed it would definitely be helpful to compile and discuss all these different scenarios to get a better idea of the end goal validation behavior

> I would say let's make the above change for now, and iterate on these other ideas in a separate PR which does not need to make it in pynwb 3.0.0rc1

Sounds good!

rly (approved):

PR #1934, opened 2024-07-15
Return can_read as False if no nwbfile version is found

### LINKED ISSUE #1919: [Bug] `NWBHDF5IO.can_read()` fails for missing version
As in the title, the problem is the following, `NWBHDF5IO.can_read()`  relies on `get_nwbfile_version` and **critically** it indexes the output:

https://github.com/NeurodataWithoutBorders/pynwb/blob/90e60484415d38d3d566dc9fea63b6d99ffdbe43/src/pynwb/__init__.py#L272

However,  `get_nwbfile_version` can also return None and then the indexing fails:

https://github.com/NeurodataWithoutBorders/pynwb/blob/90e60484415d38d3d566dc9fea63b6d99ffdbe43/src/pynwb/__init__.py#L183

It seems to me that the solution should be that if `get_nwbfile_version` returns None then `NWBHDF5IO.can_read()` should return None instead of trying to index.

From the discussion in neuroconv:

https://github.com/catalystneuro/neuroconv/issues/910

stephprince:
## Motivation

Fix #1919. The `can_read` method would previously throw a TypeError if no nwbfile version was found. It now returns`False` instead.

I also added additional tests for `can_read` to improve code coverage.

## How to test the behavior?
```python
from h5py import File
from pynwb import NWBHDF5IO

nwbfile = mock_NWBFile()
path = 'test.nwb'

with NWBHDF5IO(path, 'w') as io:
    io.write(nwbfile)
    
with File(path, mode='a') as io:
    del io.attrs['nwb_version']

print(f'{NWBHDF5IO.can_read(path) = }')
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Documentation and tutorial enhancements
 - Added pre-release pull request instructions to release process documentation @stephprince [#1928](https://github.com/NeurodataWithoutBorders/pynwb/pull/1928)
 
+### Bug fixes
+- Fixed `can_read` method to return False if no nwbfile version can be found @stephprince [#1934](https://github.com/NeurodataWithoutBorders/pynwb/pull/1934)
+
 ## PyNWB 2.8.1 (July 3, 2024)
 
 ### Documentation and tutorial enhancements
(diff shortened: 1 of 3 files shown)


mavaylon1:
@stephprince It may be informative to have a distinction between an old file and an invalid file. I think having a warning plus the return false would be better for old files.

stephprince:
@mavaylon1 added some warnings for when the file can be read by h5py but not PyNWB

mavaylon1 (approved):

h-mayorquin:
Thanks for working on this guys.

PR #1959, opened 2024-09-05
Expose cache_spec option in NWBHDF5IO.export

rly:
## Motivation

Right now, `NWBHDF5IO.export` always caches the spec. The option to change that is not exposed. This PR exposes that keyword argument. Ref: https://github.com/hdmf-dev/hdmf/issues/1187

## Checklist


rly:
I don't think tests are necessary for this quick change, but I could write one if you think we should have it @stephprince

stephprince (approved):
Looks good to me! I think it is ok to not have tests for this small argument exposure.

Issue #1962, opened 2024-09-08
[Bug]: Python 3.9 tests are failing because numpy 2.1.1 is not available on 3.9


rly:
### What happened?

We recently bumped the default numpy version used for testing to 2.1.1, but 2.1.1 is not available on Python 3.9. This causes our daily workflows to fail. See https://github.com/NeurodataWithoutBorders/pynwb/actions/runs/10757220557/job/29831017541

### Steps to Reproduce

```python
Run `pip install -r requirements.txt` on Python 3.9.
```

### Operating System

Linux

### Python Executable

Python

### Python Version

3.9

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/.github/CODE_OF_CONDUCT.rst)
- [X] Have you checked the [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [X] Have you ensured this bug was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues)?

PR #1956, opened 2024-09-04
Numpy 2.0

mavaylon1:
## Motivation

What was the reasoning behind this change? Please explain the changes briefly.

Support numpy 2.0

## How to test the behavior?
```
Show how to reproduce the new behavior (can be a bug fix or a new feature)
```

## Checklist

- [x] Did you update CHANGELOG.md with your changes?
- [x] Have you checked our [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [x] Have you ensured the PR clearly describes the problem and the solution?
- [x] Is your contribution compliant with our coding style? This can be checked running `ruff check . && codespell` from the source directory.
- [x] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/NeurodataWithoutBorders/pynwb/pulls) for the same change?
- [x] Have you included the relevant issue number using "Fix #XXX" notation where XXX is the issue number? By including "Fix #XXX" you allow GitHub to close issue #XXX when the PR is merged.

PR #1963, opened 2024-09-08
Fix numpy version for py39

rly:
## Motivation

Fi #1962. Minor change on top of #1956, so I think no changelog updated is needed.

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 # pinned dependencies to reproduce an entire development environment to use PyNWB
 h5py==3.11.0
 hdmf==3.14.3
-numpy==2.1.1; python_version > 3.9  # numpy 2.1+ is not compatible with py3.9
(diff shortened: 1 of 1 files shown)


rly:
All tests pass: https://github.com/NeurodataWithoutBorders/pynwb/actions/runs/10762603214

stephprince:
Sounds good. For the `environment-ros3.yml` file, do you prefer using just `numpy==2.1.1` there or moving the numpy install to the pip section to use the same environment markers?

stephprince (approved):

rly:
> Sounds good. For the `environment-ros3.yml` file, do you prefer using just `numpy==2.1.1` there or moving the numpy install to the pip section to use the same environment markers?

The Python version in that file is pinned to Python 3.12 (so we only test ros3 support on Python 3.12), so it should be OK.

Issue #1974, opened 2024-10-23
[Feature]: Add `read_nwb` to simplify reading nwbfiles 


h-mayorquin:
### What would you like to see added to PyNWB?

This is a comment that I have gotten from various users: **reading and nwbfile is not as easy as it could be.**

At the moment, when I google "how to read and nwbfile", the results show a suggestion to use [NWBHDF5IO](https://pynwb.readthedocs.io/en/latest/pynwb.html#pynwb.NWBHDF5IO). This is also what we have on the [tutorial](https://pynwb.readthedocs.io/en/latest/tutorials/general/plot_read_basics.html#opening-an-nwb-file-with-nwbhdf5io). 

While the IO approach is flexible, powerful, and follows good practices (i.e., it can be used as a context manager) it also suffers from what I think are several problems that detract from usability:
1) It has too many options: path, file, region, mode, manager, etc.
2) It is hard to remember the name to import.
3) It does not read zarr.
4) It requires two method calls, first instantiating the IO and then using read on the IO.

### What solution would you like?

I think that we could provide users with a simple entry point that covers most cases and leave the IO approach as the power tool behind the scenes:

```python
from pynwb import read_nwb

nwbfile = read_nwb(file_path)

```
 I think the name is easier to remember and intuitive about what it does.

For simplicity, this function should:
*  Work only paths
*  Always on read mode 
* Open both zarr and hdf5.
* load_namespaces=True
*  If we are ambitious we can have a default streaming choice if a link to dandi s3 asset is passed.

Downsides?

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/.github/CODE_OF_CONDUCT.rst)
- [X] Have you checked the [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [X] Have you ensured this change was not already [requested](https://github.com/NeurodataWithoutBorders/pynwb/issues)?

stephprince:
I like this idea! For me the main downside would be any potential confusion for users about having multiple methods to read a file and when to use `read_nwb` vs. `NWBHDF5IO`, but I think this could be clarified in the docs. 

@oruebel @rly @bendichter thoughts?

oruebel:
[...]

PR #1979, opened 2024-11-07
add NWBHDF5IO.read_nwb() method

h-mayorquin:
This was discussed in #1974

This is the first part of the program described by @oruebel [here](https://github.com/NeurodataWithoutBorders/pynwb/issues/1974#issuecomment-2455440137). In a future PR and once https://github.com/hdmf-dev/hdmf-zarr/issues/225 is merged I will implement `pynwb.read_nwb` in another PR.

Meanwhile, I have two questions:
1) What streaming method should be used when an s3 path is passed?
2) For testing this, the [ros3 test](https://github.com/h-mayorquin/pynwb/blob/7c65fe2cb20f10cefa0cefeae019ce32c6b1706e/tests/integration/ros3/test_ros3.py#L10-L20) have some files that I think I could use, is that OK? 

## Checklist


oruebel:
> 2\. For testing this, the [ros3 test](https://github.com/h-mayorquin/pynwb/blob/7c65fe2cb20f10cefa0cefeae019ce32c6b1706e/tests/integration/ros3/test_ros3.py#L10-L20) have some files that I think I could use, is that OK?

Yes, that should be fine

> 1. What streaming method should be used when an s3 path is passed

I think we should probably use `fsspec` as the default https://pynwb.readthedocs.io/en/latest/tutorials/advanced_io/streaming.html#streaming-method-1-fsspec 

If installed, we could also use `remfile` https://pynwb.readthedocs.io/en/latest/tutorials/advanced_io/streaming.html#method-3-remfile  and use `fsspec` as the fall-back. `remfile` tends to be a bit faster than `fsspec`, but I think `fsspec` is more widely used. 

@bendichter what would you suggest for this?

oruebel:
Aside from adding s3 and testing, the outline you started here looks good to me.

h-mayorquin:
OK, I added s3 testing.

h-mayorquin:
Is the test failing a permission issue?
image

rly:
Seems like it was a codecov issue with detecting forks, perhaps when commits happen in quick succession: https://github.com/codecov/codecov-action/issues/1580 . I re-ran the failing runs.

tests/integration/hdf5/test_io.py
```diff
@@ -586,3 +593,38 @@ def test_can_read_file_old_version(self):
...
+        
+        file = h5py.File(self.path, 'r')
+        
+        read_nwbfile = NWBHDF5IO.read_nwb(file=file)
+        self.assertContainerEqual(read_nwbfile, self.nwbfile)
+
+        read_nwbfile.get_read_io().close()
+    
+    @unittest.skipIf(not HAVE_FSSPEC, "fsspec library not available")
+    def test_read_nwb_method_s3_path():
```
rly:
```suggestion
    def test_read_nwb_method_s3_path(self):
```

requirements-opt.txt
```diff
@@ -1,3 +1,8 @@
 linkml-runtime==1.7.4; python_version >= "3.9"
 schemasheets==0.2.1; python_version >= "3.9"
 oaklib==0.5.32; python_version >= "3.9"
+
+# for streaming tests
+fsspec==2024.10.0
+requests==2.32.3
+aiohttp==3.10.10
```
oruebel, line 8: Are ``requests`` and ``aiohttp`` needed here or these being installed by fsspec? I don't see these libraries being used directly here and if ``fsspec`` installs them then I don't think we should need to specify these separately.

rly: fsspec does not install them. they are required for http streaming

tests/integration/hdf5/test_io.py
```diff
@@ -586,3 +593,38 @@ def test_can_read_file_old_version(self):
...
+        import h5py
+        
+        file = h5py.File(self.path, 'r')
+        
+        read_nwbfile = NWBHDF5IO.read_nwb(file=file)
+        self.assertContainerEqual(read_nwbfile, self.nwbfile)
+
+        read_nwbfile.get_read_io().close()
+    
+    @unittest.skipIf(not HAVE_FSSPEC, "fsspec library not available")
```
oruebel, line 624: @rly does Codecove run this test when `fsspec` is specified as an optional requirement or would we need to add `fsspec` elsewhere too? I'm just wondering because codecov seems to report the s3 parts of the code changes as not covered above, even though this test should cover them.

oruebel (approved):
Aside from the codecov question, this looks good to me. I approve the PR, because I assume the codecov question can probably be resolved in the comment, rather than requiring additional code changes, and it seems easier for me to re-approve if there are changes, rather than holding up the PR.

tests/integration/hdf5/test_io.py
rly: It does run this. the codecov "Files Changed" annotation seems to be inconsistent with their website: https://app.codecov.io/gh/NeurodataWithoutBorders/pynwb/pull/1979?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=NeurodataWithoutBorders

oruebel: Thanks @rly for the clarification.

PR #1994, opened 2024-11-16
Add `read_nwb_method` for local paths in both hdf5 and zarr

h-mayorquin:
## Motivation
See #1974 

## How to test the behavior?
I added tests.
## Checklist


src/pynwb/__init__.py
```diff
@@ -535,6 +535,71 @@ def read_nwb(**kwargs):
...
+    backend_is_hdf5 = NWBHDF5IO.can_read(path=path)
+    if backend_is_hdf5:
+        return NWBHDF5IO.read_nwb(path=path)
+    else:
+        from hdmf_zarr import NWBZarrIO
+        backend_is_zarr = NWBZarrIO.can_read(path=path)
+        if backend_is_zarr:
+            return NWBZarrIO.read_nwb(path=path) 
+        else:
+            raise ValueError(f"Unsupported backend for file: {path}")
```
stephprince:
```suggestion
        try:
            from hdmf_zarr import NWBZarrIO
            backend_is_zarr = NWBZarrIO.can_read(path=path)
            if backend_is_zarr:
                return NWBZarrIO.read_nwb(path=path) 
            else:
                raise ValueError(f"Unsupported backend for file: {path}")     
        except ImportError:
            raise ValueError(f"Unsupported backend for file: '{path}'. If you are trying to read a Zarr file, make sure you have hdmf-zarr installed.")
```

I believe read_nwb will currently return an ImportError if the user provides an invalid path but does not have hdmf_zarr installed. Can you support this case?

tests/integration/io/test_read.py
```diff
@@ -0,0 +1,48 @@
...
+        with tempfile.TemporaryDirectory() as temp_dir:
+            path = Path(temp_dir) / "test.nwb"
+            with NWBHDF5IO(path, 'w') as io:
+                io.write(self.nwbfile)
+            
+            read_nwbfile = read_nwb(path=path)
+            self.assertContainerEqual(read_nwbfile, self.nwbfile)
+            read_nwbfile.get_read_io().close()
+            
+    @unittest.skipIf(not HAVE_NWBZarrIO, "NWBZarrIO library not available")
```
stephprince, line 35: Can you add tests for
1. the user providing a path that does not exist
2. if the user does not have hdmf_zarr installed and provides an unsupported file path

src/pynwb/__init__.py
```diff
@@ -535,6 +535,71 @@ def read_nwb(**kwargs):
...
+    --------
+    pynwb.NWBHDF5IO : Core I/O class for HDF5 files with advanced options.
+    hdmf_zarr.nwb.NWBZarrIO : Core I/O class for Zarr files with advanced options.
+
+    Notes
+    -----
+    This function uses the following defaults:
+    * Always opens in read-only mode
+    * Automatically loads namespaces
+    * Detects file format based on extension
```
stephprince:
```suggestion
    * Detects file format based on whether supported backend IO classes can read the file
```

I think in the current implementation other extensions will work since `can_read` and `IO.read_nwb` methods just check whether the IO can successfully open the file?

CHANGELOG.md
```diff
@@ -3,12 +3,13 @@
...
-  - Deprecated `EventWaveform` neurodata type. @rly [#1940](https://github.com/NeurodataWithoutBorders/pynwb/pull/1940)
-  - Deprecated `ImageMaskSeries` neurodata type. @rly [#1941](https://github.com/NeurodataWithoutBorders/pynwb/pull/1941)
+- Removed `SpatialSeries.bounds` field that was not functional. This will be fixed in a future release. @rly [#1907](https://github.com/NeurodataWithoutBorders/pynwb/pull/1907), [#1996](https://github.com/NeurodataWithoutBorders/pynwb/pull/1996)
+- Added support for `NWBFile.was_generated_by` field. @stephprince [#1924](https://github.com/NeurodataWithoutBorders/pynwb/pull/1924)
+- Added support for `model_number`, `model_name`, and `serial_number` fields to `Device`. @stephprince [#1997](https://github.com/NeurodataWithoutBorders/pynwb/pull/1997)
+- Deprecated `EventWaveform` neurodata type. @rly [#1940](https://github.com/NeurodataWithoutBorders/pynwb/pull/1940)
+- Deprecated `ImageMaskSeries` neurodata type. @rly [#1941](https://github.com/NeurodataWithoutBorders/pynwb/pull/1941)
```
stephprince:
```suggestion
  - Removed `SpatialSeries.bounds` field that was not functional. This will be fixed in a future release. @rly [#1907](https://github.com/NeurodataWithoutBorders/pynwb/pull/1907), [#1996](https://github.com/NeurodataWithoutBorders/pynwb/pull/1996)
   - Added support for `NWBFile.was_generated_by` field. @stephprince [#1924](https://github.com/NeurodataWithoutBorders/pynwb/pull/1924)
   - Added support for `model_number`, `model_name`, and `serial_number` fields to `Device`. @stephprince [#1997](https://github.com/NeurodataWithoutBorders/pynwb/pull/1997)
   - Deprecated `EventWaveform` neurodata type. @rly [#1940](https://github.com/NeurodataWithoutBorders/pynwb/pull/1940)
   - Deprecated `ImageMaskSeries` neurodata type. @rly [#1941](https://github.com/NeurodataWithoutBorders/pynwb/pull/1941)
```

These updates were indented to indicate schema 2.8.0 related changes

docs/gallery/general/plot_read_basics.py
```diff
@@ -103,14 +102,17 @@
 # read the data into a :py:class:`~pynwb.file.NWBFile` object.
 
 filepath = "sub-P11HMH_ses-20061101_ecephys+image.nwb"
-# Open the file in read mode "r",
-io = NWBHDF5IO(filepath, mode="r")
-nwbfile = io.read()
+from pynwb import read_nwb
+
+nwbfile = read_nwb(filepath)
```
stephprince, line 107: I think once the simplified read_nwb with streaming is supported we may want to go through the other tutorials that use io.read() in read only mode and update those as well.

stephprince (commented):
Sorry for the delay @h-mayorquin! Added my review comments

CHANGELOG.md
h-mayorquin: Ah, thanks.

src/pynwb/__init__.py
h-mayorquin: Thanks, I added this change on the last commit.

src/pynwb/__init__.py
h-mayorquin: This lien was strange, I changed to what I intended which is that it reads both zarr and nwb but this is unrelated to the extension of the file. Let me know what you think.

tests/integration/io/test_read.py
h-mayorquin: Done.

h-mayorquin:
Is the sphinx links fail related to this PR? Given my experience with neuroconv is seldom is and I looked at the CI log but no luck.

src/pynwb/__init__.py
```diff
@@ -531,6 +531,82 @@ def read_nwb(**kwargs):
...
+        try:
+            from hdmf_zarr import NWBZarrIO
+            backend_is_zarr = NWBZarrIO.can_read(path=path)
+            if backend_is_zarr:
+                return NWBZarrIO.read_nwb(path=path) 
+            else:
+                raise ValueError(
+                    f"Unable to read file: '{path}'. The file exists but is not recognized as "
+                    "either a valid HDF5 or Zarr NWB file. Please ensure the file contains valid NWB data."
+                )
```
stephprince:
```suggestion
                    f"Unable to read file: '{path}'. The file is not recognized as "
                    "either a valid HDF5 or Zarr NWB file. Please ensure the file exists and contains valid NWB data."
                )     
```

I think this error will occur even if the file does not exist.

src/pynwb/__init__.py
```diff
@@ -531,6 +531,82 @@ def read_nwb(**kwargs):
...
+
+    Examples
+    --------
+    Read a local NWB file:
+
+    >>> from pynwb import read_nwb
+    >>> nwbfile = read_nwb("path/to/file.nwb")
+
+    
+    """
```
stephprince: The sphinx link failures do seem to be related to the new docstring. I think docval does not play well with numpy style docstrings. I added a suggestion below that should address the failures but feel free to modify:

```suggestion
@docval({'name': 'path', 'type': (str, Path), 
         'doc': 'Path to the NWB file. Can be either a local filesystem path to '
                'an HDF5 (.nwb) or Zarr (.zarr) file.'}, 
        is_method=False)
def read_nwb(**kwargs):
    """Read an NWB file from a local path.

    High-level interface for reading NWB files. Automatically handles both HDF5 
    and Zarr formats. For advanced use cases (parallel I/O, custom namespaces), 
    use NWBHDF5IO or NWBZarrIO.

    See also 
        * :py:class:`~pynwb.NWBHDF5IO`: Core I/O class for HDF5 files with advanced options.
        * :py:class:`~hdmf_zarr.nwb.NWBZarrIO`: Core I/O class for Zarr files with advanced options.

    Notes
        This function uses the following defaults:
            * Always opens in read-only mode
            * Automatically loads namespaces
            * Reads any backend (e.g. HDF5 or Zarr) if there is an IO class available.

        Advanced features requiring direct use of IO classes (e.g. NWBHDF5IO NWBZarrIO) include:
            * Streaming data from s3
            * Custom namespace extensions
            * Parallel I/O with MPI
            * Custom build managers
            * Write or append modes
            * Pre-opened HDF5 file objects or Zarr stores
            * Remote file access configuration
 
    Example usage reading a local NWB file:

    .. code-block:: python

        from pynwb import read_nwb
        nwbfile = read_nwb("path/to/file.nwb")    

    :Returns: pynwb.NWBFile The loaded NWB file object.
    """
```

stephprince (commented):
Looks good to me, thanks! Added a few more small suggestions related to the docstrings.

src/pynwb/__init__.py
h-mayorquin: Thanks!

src/pynwb/__init__.py
h-mayorquin: Thanks a bunch.

h-mayorquin: Which part of the docstring does docval have troubles with?

src/pynwb/__init__.py
h-mayorquin: I need to modify the test though, so I am doing this into another commit.

src/pynwb/__init__.py
stephprince: It seems to be with the section titles:

```
/home/runner/work/pynwb/pynwb/src/pynwb/__init__.py:docstring of pynwb.docval.<locals>.dec.<locals>.func_call:8: CRITICAL: Unexpected section title.

Parameters
---------- [docutils]
```

h-mayorquin: Ah, thanks!

h-mayorquin:
Implemented the changes but now there is more CI trouble it seems.

stephprince:
The coverage report is saying that `read_nwb` had no test coverage. It seems the new test file is not being run with the other integration tests. 

Since we still run tests using `test.py` and haven't migrated completely to pytest (see https://github.com/NeurodataWithoutBorders/pynwb/issues/1803), I think adding  `run_test_suite("tests/integration/io", "integration io tests", verbose=verbose)` to `test.py` [here](https://github.com/NeurodataWithoutBorders/pynwb/blob/3d04646918226486a4da525693eaf0ad6c9a0091/test.py#L266) should ensure the new code is being tested in the CI.

h-mayorquin:
Thanks @stephprince . I added this.

> `run_test_suite("tests/integration/io", "integration io tests", verbose=verbose)`

Will this handle both the cases with and without `hdmf-zarr`? as it, include them in the coverage.

stephprince:
> Will this handle both the cases with and without hdmf-zarr? as it, include them in the coverage.

I think so. The coverage is run with and without optional requirements (which includes hdmf-zarr)

h-mayorquin:
Yeah, I am confused about the difference between codecov/patch and codecov/project. The latter works fine but the former indicates that the test is not run at all. It seems:

image

stephprince:
I don't think the updates to add the integration io tests to `test.py` got pushed, I added them and the `__init__.py` that I think is necessary from testing locally. 

From my understanding codecov/project is the coverage for all of pynwb when applying the PR changes, and patch is the coverage for the lines that were specifically changed. I'm not sure exactly how the coverage reports are merged and reported within the PR, maybe @rly knows in more detail?

stephprince (approved):

stephprince:
Looks like we are good now though and testing is running with and without hdmf-zarr. Thanks for the work on this!

h-mayorquin:
> I don't think the updates to add the integration io tests to `test.py` got pushed, I added them and the `__init__.py` that I think is necessary from testing locally.

My bad, I realize now -painfully- that I pushed those changes to other branch : /
Thanks for fixing my mistake here.

> From my understanding codecov/project is the coverage for all of pynwb when applying the PR changes, and patch is the coverage for the lines that were specifically changed. I'm not sure exactly how the coverage reports are merged and reported within the PR, maybe @rly knows in more detail?


Thanks for the info and all the help on this PR!

PR #1996, opened 2024-11-18
Temporarily revert fix for SpatialSeries.bounds

### LINKED ISSUE #1992: [Bug]: Cannot add "bounds" attribute to SpatialSeries/data
### What happened?

In https://github.com/NeurodataWithoutBorders/nwb-schema/pull/567 (pending NWB schema 2.8.0), we added a new optional "bounds" field to "SpatialSeries/data". It turns out that the current version of PyNWB / HDMF does not support writing (and probably also validating) additional attributes to untyped datasets like "data". I am surprised that we have not encountered this before.

This line in HDMF's ObjectMapper https://github.com/hdmf-dev/hdmf/blob/26d584cc0a3e0985d0e053ab9eeaf5bf8849cd31/src/hdmf/build/objectmapper.py#L1107
does not add new attributes from the new spec.

The issue appears to be due to resolution of the spec. I will open a ticket in HDMF.

While we resolve this in HDMF, I propose we revert the addition of "bounds" to "SpatialSeries/data" to unblock the other schema/pynwb issues.

### Steps to Reproduce

```python
import numpy as np

from pynwb.testing import AcquisitionH5IOMixin, TestCase


class TestSpatialSeriesIO(AcquisitionH5IOMixin, TestCase):

    def setUpContainer(self):
        """ Return the test TimeSeries to read/write """
        return SpatialSeries(
            name='test_sS',
            data=np.ones((3, 2)),
            data__bounds=[(-1,1),(-1,1),(-1,1)],
            reference_frame='reference_frame',
            timestamps=[1., 2., 3.]
        )
```

### Traceback

```python
# data__bounds is not written
```

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.11

### Package Versions

_No response_

### Code of Conduct

rly:
## Motivation

Fix #1992 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Enhancements and minor changes
 - Added support for NWB schema 2.8.0:
-  - Fixed support for `data__bounds` field to `SpatialSeries` to set optional boundary range (min, max) for each dimension of data. Removed `SpatialSeries.bounds` field that was not functional. @rly [#1907](https://github.com/NeurodataWithoutBorders/pynwb/pull/1907)
+  - Removed `SpatialSeries.bounds` field that was not functional. This will be fixed in a future release. @rly [#1907](https://github.com/NeurodataWithoutBorders/pynwb/pull/1907), 
 
 ## PyNWB 2.8.3 (Upcoming)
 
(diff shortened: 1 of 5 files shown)


rly:
Because we reverted commits in nwb-schema, the order of adding pynwb support for changes in nwb-schema 2.8.0 is no longer straightforward (and may not even be possible to not have issues). I will merge this and the other schema 2.8.0 PyNWB PRs into the nwb_schema_2.8.0 branch and look at what issues, if any, remain.

PR #2002, opened 2024-11-20
add EventDetection and FeatureExtraction code to ecephys tutorial

bendichter:
## Motivation

I realized that there was no code demonstrating how to use EventDetection, so I added a small snippet. I also consolidated some of the information from another paragraph

### DIFF
--- a/docs/gallery/domain/ecephys.py
+++ b/docs/gallery/domain/ecephys.py
@@ -350,10 +350,26 @@ event_detection = EventDetection(
 
 ecephys_module.add(event_detection)
 
-#######################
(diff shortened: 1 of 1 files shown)


docs/gallery/domain/ecephys.py
```diff
@@ -243,7 +243,27 @@
 )
 ecephys_module.add(lfp)
 
-####################
+#######################
+# If the derived data is filtered by not downsampled, you can store the data in an
```
oruebel:
```suggestion
# If the derived data is filtered but not downsampled, you can store the data in an
```

oruebel (approved):

PR #2004, opened 2024-11-20
Add all patch clamp types to latest icephys tutorial

bendichter:
## Motivation
I found that 3 of the patch clamp types were missing from our most recent icephys tutorial


docs/gallery/domain/plot_icephys.py
```diff
@@ -324,6 +324,49 @@
...
+    name="ccss",
+    data=[1, 2, 3, 4, 5],
+    starting_time=123.6,
+    rate=10e3,
+    electrode=electrode,
+    gain=0.02,
+    sweep_number=np.uint(16),
+)
+
+izcs = IZeroClampSeries(
```
stephprince, line 357: Should `IZeroClampSeries` be in the same section as current clamp? Might be helpful to add a small comment here explaining when this type is used.

bendichter, line 357:
```suggestion
# IZeroClampSeries is used when the current is clamped to 0.
izcs = IZeroClampSeries(
```

ehennestad:
@bendichter Should the types also be added to the file?

stephprince (approved):

PR #2018, opened 2025-01-23
update default mock_NWBFile start time

### LINKED ISSUE #2013: [Bug]:  Issue with Default session_start_time timezone in mock_NWBFile
### Description

A potential bug occurs when using a mock function for creating `NWBFile` objects. The function set `session_start_time` to  datetime(1970, 1, 1, tzinfo=tzlocal()). This results in an OSError: [Errno 22] Invalid argument when attempting to write the file using pynwb.NWBHDF5IO. The issue seems to be caused by the interaction between tzlocal() and early datetime values (e.g., the epoch time 1970-01-01), particularly in systems located in time zones earlier than UTC, where the local time would fall outside the valid range for the epoch.

### Steps to Reproduce

```python

# Create NWBFile object using default session_start_time 
# 
nwbfile = mock_NWBFile()

# Attempt to write the NWB file
with NWBHDF5IO('test.nwb', mode='w') as io:
    io.write(nwbfile)  # This triggers the error
```

### Traceback

```python
...
    return x.isoformat()  # method works for both date and datetime
..…\dateutil\tz\tz.py:291: in _naive_is_dst
    dstval = self._naive_is_dst(dt)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = tzlocal(), dt = datetime.datetime(1970, 1, 1, 0, 0, tzinfo=tzlocal())

    def _naive_is_dst(self, dt):
        timestamp = _datetime_to_timestamp(dt)
>       return time.localtime(timestamp + time.timezone).tm_isdst
E       OSError: [Errno 22] Invalid argument
```

### Operating System

Windows

### Python Executable

Conda

### Python Version

3.11

### Package Versions

_No response_

### Code of Conduct

stephprince:
## Motivation

Fix #2013. There was an error when creating a `mock_NWBFile` in specific time zones, where converting the default start time to local time would result in a time before the Unix epoch.

The default could also be updated to any later time, I left as 1970 for now in case that is used to indicate this is mock data.

I don't think these changes require a CHANGELOG update but can add if needed.

## How to test the behavior?
```python
from pynwb import NWBHDF5IO

nwb = mock_NWBFile()

with NWBHDF5IO('test.nwb', mode='w') as io:
    io.write(nwbfile)
```

## Checklist

### DIFF
--- a/src/pynwb/testing/mock/file.py
+++ b/src/pynwb/testing/mock/file.py
@@ -10,7 +10,7 @@ from .utils import name_generator
 def mock_NWBFile(
     session_description: str = 'session_description',
     identifier: Optional[str] = None,
-    session_start_time: datetime = datetime(1970, 1, 1, 12, tzinfo=tzutc()),
+    session_start_time: datetime = datetime(1970, 1, 1, tzinfo=tzutc()),
     **kwargs
 ):
     return NWBFile(
--- a/src/pynwb/testing/mock/file.py
+++ b/src/pynwb/testing/mock/file.py
@@ -1,7 +1,7 @@
 from typing import Optional
 from uuid import uuid4
 from datetime import datetime
-from dateutil.tz import tzlocal
+from dateutil.tz import tzutc
 
 from ...file import NWBFile, Subject
 from .utils import name_generator
@@ -10,7 +10,7 @@ from .utils import name_generator
 def mock_NWBFile(
     session_description: str = 'session_description',
     identifier: Optional[str] = None,
-    session_start_time: datetime = datetime(1970, 1, 1, tzinfo=tzlocal()),
+    session_start_time: datetime = datetime(1970, 1, 1, 12, tzinfo=tzutc()),
     **kwargs
 ):
     return NWBFile(


h-mayorquin:
[Edit]
Tangential comment that is now its own issue:
https://github.com/NeurodataWithoutBorders/pynwb/issues/2027#issuecomment-2620254344

bendichter:
> Hey guys, slightly related but now that python 3.9 is minimal have you considered using ZoneInfo from the standard library instead of dateutil?
> 
> https://docs.python.org/3/library/zoneinfo.html
> 
> Eventually you can eliminate another dependency.

@h-mayorquin Good idea but let's open a separate issue for that since it will probably require editing code all over the place

bendichter:
lgtm but I'm also interested in @rly 's take on this

src/pynwb/testing/mock/file.py
```diff
@@ -10,7 +10,7 @@
 def mock_NWBFile(
     session_description: str = 'session_description',
     identifier: Optional[str] = None,
-    session_start_time: datetime = datetime(1970, 1, 1, tzinfo=tzlocal()),
+    session_start_time: datetime = datetime(1970, 1, 1, 12, tzinfo=tzutc()),
```
rly: It seems from the discussion at the end of #2013 that using 0 hour and UTC timezone is OK, so I suggest using that, which is more standard and recognizable.
```suggestion
    session_start_time: datetime = datetime(1970, 1, 1, tzinfo=tzutc()),
```

bendichter (approved):

rly (approved):

PR #2021, opened 2025-01-27
Add all to init

### LINKED ISSUE #1686: [Feature]: use `__all__` for modules

Python allows you to use `__all__` to indicate the public interface of a module. This is commonly used for the `from x import *` pattern, where it limits the returned classes to those listed in the `__all__` list.

While I don't necessarily recommend using that import pattern, I still think `__all__` would be useful, because it informs the user what public classes are defined in that module. I have a current situation where I would like to import all of the neurodata types implemented in a module. Simply using the `dir()` command, I get all of the neurodata types that are imported into the module, as well as a bunch of double underscore functions, which I do not want.

```python
from pynwb import ecephys
dir(ecephys)
```

```
['CORE_NAMESPACE',
 'ClusterWaveforms',
 'Clustering',
 'DataChunkIterator',
 'Device',
 'DynamicTableRegion',
 'ElectricalSeries',
 'ElectrodeGroup',
 'EventDetection',
 'EventWaveform',
 'FeatureExtraction',
 'FilteredEphys',
 'Iterable',
 'LFP',
 'MultiContainerInterface',
 'NWBContainer',
 'NWBDataInterface',
 'SpikeEventSeries',
 'TimeSeries',
 '__builtins__',
 '__cached__',
 '__doc__',
 '__file__',
 '__loader__',
 '__name__',
 '__package__',
 '__spec__',
 'assertEqualShape',
 'docval',
 'get_data_shape',
 'get_docval',
 'popargs',
 'popargs_to_dict',
 'register_class',
 'warnings']
```

`__all__` lists are used on many popular packages, such as sci-kit learn: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/feature_extraction/image.py

The `__all__` list could be used in a number of ways, for instance to determine coverage of tests, mock classes, and tutorials.

_No response_

see above

Yes.

### Code of Conduct

bendichter:
Fix #1686: Add `__all__` to modules

# Motivation
Python allows using `__all__` to indicate the public interface of a module. This helps users understand what classes and functions are intended to be imported from each module, improves IDE auto-completion, and prevents accidental usage of internal implementation details. This change follows the pattern used by other major packages like scikit-learn.

# How to test the behavior?
```python
# Before: dir() shows all module contents including internals
from pynwb import ecephys
print(dir(ecephys))  # Shows everything including __builtins__, etc.

# After: Only public interfaces are imported with *
from pynwb import ecephys
print(ecephys.__all__)  # Shows only intended public interfaces
from pynwb.ecephys import *  # Only imports items listed in __all__
```

# Checklist
[x] Did you update CHANGELOG.md with your changes?
[x] Have you checked our [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
[x] Have you ensured the PR clearly describes the problem and the solution?
[x] Is your contribution compliant with our coding style? This can be checked running ruff check . && codespell from the source directory.
[x] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/NeurodataWithoutBorders/pynwb/pulls) for the same change?
[x] Have you included the relevant issue number using "Fix #1686" notation where XXX is the issue number?


src/pynwb/spec.py
```diff
@@ -8,6 +8,19 @@
...
+__all__ = [
+    'NWBRefSpec',
+    'NWBAttributeSpec',
+    'NWBLinkSpec',
+    'NWBDtypeSpec',
+    'NWBDatasetSpec',
+    'NWBGroupSpec',
+    'NWBNamespace',
+    'NWBNamespaceBuilder',
+    'export_spec'  # Re-exported from hdmf.spec.write
```
stephprince, line 20: Why should `export_spec` be included here?

stephprince (commented):
Looks good to me. Could you add a CHANGELOG update? 

And then I think adding the `__all__` definitions has caused some issues with the Sphinx documentation:

```
/home/runner/work/pynwb/pynwb/docs/source/pynwb.testing.icephys_testutils.rst:2: WARNING: more than one target found for cross-reference 'NWBFile': pynwb.NWBFile, pynwb.file.NWBFile [ref.python]
```

I think one solution would be to update the references to use the more specific path. Could you address those warnings?

src/pynwb/spec.py
bendichter: This was originally included here for backwards compatibility, a carry-over from when `export_spec` was defined in this module. I think we should keep it here.

stephprince (commented):
There were still some warnings in the link checker. I took a closer look and I think since the deprecated or private classes (e.g. SweepTable, BaseStorageOverride) were not included in the `__all__` definitions, they were no longer being found by sphinx autodoc and so were not being documented:

```
/home/runner/work/pynwb/pynwb/src/pynwb/file.py:docstring of pynwb.file.NWBFile:78: WARNING: py:class reference target not found: pynwb.icephys.SweepTable [ref.class]
```

I found this option that we can add to the `docs/source/conf.py` to solve these warnings, since I believe we still want the documentation for these classes to be accessible on readthedocs even if they are not included in the public interface:

```python
autodoc_default_options = {
    'ignore-module-all': True,
}
```

stephprince (approved):

PR #2029, opened 2025-01-29
Images in image mod docs

### LINKED ISSUE #2022: [Documentation]: Images object not present in the `pynwb.image` API

This works on my system:

```python
from pynwb.image import Images
```

But the object is not found in the images API:

https://pynwb.readthedocs.io/en/stable/pynwb.image.html

Instead, it is found on base:

https://pynwb.readthedocs.io/en/stable/pynwb.base.html#pynwb.base.Images

It makes sense to me that it should be on Images as I would look for that but maybe it should be duplicated to be here and in base? 

Should pynwb.base be internal?

I guess slightly related to https://github.com/NeurodataWithoutBorders/pynwb/issues/1686

No.

### Code of Conduct

bendichter:
fix #2022


.gitignore
```diff
@@ -48,6 +48,9 @@ build/
 dist/
 *.egg-info
 
+# Virtual Environment
+venv/
+
```
stephprince:
```suggestion
```

This was already added to the gitignore and duplicated when merging.

src/pynwb/image.py
```diff
@@ -1,3 +1,22 @@
...
+- :class:`~pynwb.image.IndexSeries`: Stores indices to image frames stored in an ImageSeries
+- :class:`~pynwb.image.OpticalSeries`: Image data that is presented or recorded
+- :class:`~pynwb.image.GrayscaleImage`: Single-channel (grayscale) image data
+- :class:`~pynwb.image.RGBImage`: Three-channel (RGB) image data
+- :class:`~pynwb.image.RGBAImage`: Four-channel (RGBA) image data with transparency
+- :class:`~pynwb.base.Images`: Container for storing a collection of images (imported from :mod:`pynwb.base`)
+- :class:`~pynwb.base.Image`: Base class for image data (imported from :mod:`pynwb.base`)
+
+Note: While the :class:`~pynwb.base.Image` and :class:`~pynwb.base.Images` classes are defined in :mod:`pynwb.base`, they can be imported directly from this module:
+    from pynwb.image import Image, Images
```
stephprince:
```suggestion
Note: While the :class:`~pynwb.base.Image` and :class:`~pynwb.base.Images` classes are defined in :mod:`pynwb.base`, 
they can be imported directly from this module:
    from pynwb.image import Image, Images
```

Break up line for ruff

stephprince (approved):

PR #2031, opened 2025-02-06
Enhance unit test coverage

oruebel:
## Motivation

Add unit tests to enhance overall test coverage for 
 * Added unit tests for missed logic in `core.py`, `image.py`, `spec.py`, `icephys.py`, `epoch.py`
 * Fixed bug in `IntracellularRecordingsTable`, the `IntracellularResponseTable` was not being initialized correctly when empty category tables were provided
 * Fixed shape check in `SpikeEventSeries` to handle the case where `AbstractDataChunkIterator` are used for timestamps/data

## Checklist

- [X] Did you update CHANGELOG.md with your changes?


oruebel:
@stephprince should this PR target the `dev` or the release branch for the 3.0 release?

stephprince:
> @stephprince should this PR target the dev or the release branch for the 3.0 release?

Since you mentioned there are breaking changes, I think this should target the 3.0 release branch. But I can review and take a closer look.

oruebel:
> I think this should target the 3.0 release branch. But I can review and take a closer look.

Sounds good. Feel free to modify the PR as necessary.

If you prefer, it should be pretty straight forward to just move the breaking changes to a separate PR since its just removes a few functions that were no longer used anywhere.

oruebel:
@stephprince @rly to simplify things I have moved all breaking changes to a new PR #2036. This PR is good to go from my end.

src/pynwb/ophys.py
```diff
@@ -338,7 +338,7 @@ def add_roi(self, **kwargs):
         """Add a Region Of Interest (ROI) data to this"""
         pixel_mask, voxel_mask, image_mask = popargs('pixel_mask', 'voxel_mask', 'image_mask', kwargs)
         if image_mask is None and pixel_mask is None and voxel_mask is None:
-            raise ValueError("Must provide 'image_mask' and/or 'pixel_mask'")
+            raise ValueError("Must provide at least on of 'image_mask', 'pixel_mask', or 'voxel_mask'")
```
stephprince:
```suggestion
            raise ValueError("Must provide at least one of 'image_mask', 'pixel_mask', or 'voxel_mask'")
```

src/pynwb/core.py
```diff
@@ -141,12 +154,18 @@ def __init__(self, **kwargs):
 
     @property
     def notes(self):
+        """
+        Get the notes attribute. This will be deprecated in the future.
```
stephprince:
```suggestion
        Get the notes attribute. Use of ScratchData.notes has been deprecated and will be removed in PyNWB 4.0.
```

src/pynwb/core.py
```diff
@@ -141,12 +154,18 @@ def __init__(self, **kwargs):
...
+        """
         warn('Use of ScratchData.notes will be deprecated. Use ScratchData.description instead.',
              PendingDeprecationWarning)
         return self.description
 
     @notes.setter
     def notes(self, value):
+        """
+        Set the notes attribute. This will be deprecated in the future.
```
stephprince:
```suggestion
        Set the notes attribute. Use of ScratchData.notes has been deprecated and will be removed in PyNWB 4.0.
```

tests/unit/test_ophys.py
```diff
@@ -558,6 +576,13 @@ def test_init_3d_image_mask(self):
         self.assertTrue(np.allclose(pS['image_mask'][0], img_masks[0]))
         self.assertTrue(np.allclose(pS['image_mask'][1], img_masks[1]))
 
+    def test_add_roi_missing_params(self):
+        _, _, pS = self.create_basic_plane_segmentation()
+        msg = "Must provide at least on of 'image_mask', 'pixel_mask', or 'voxel_mask'"
```
stephprince:
```suggestion
        msg = "Must provide at least one of 'image_mask', 'pixel_mask', or 'voxel_mask'"
```

tests/unit/test_epoch.py
```diff
@@ -66,6 +68,117 @@ def test_dataframe_roundtrip_drop_ts(self):
...
+        epochs = TimeIntervals(name='test_epochs')
+        epochs.add_interval(start_time=10.0, stop_time=20.0)
+        row = epochs[0]
+        self.assertEqual(row.loc[0]['start_time'], 10.0)
+        self.assertEqual(row.loc[0]['stop_time'], 20.0)
+
+    def test_add_interval_tags_string(self):
+        """Test adding interval with tags as comma-separated string"""
+        epochs = TimeIntervals(name='test_epochs')
+        epochs.add_interval(start_time=10.0, stop_time=20.0, tags='tag1, tag2, tag3')
```
stephprince, line 82: Should tags be allowed to be supplied this way? It feels like this approach could be prone to misunderstandings on how the data gets written. 

Looking at the line of code that this is testing and the tests that fail if you remove it:
https://github.com/NeurodataWithoutBorders/pynwb/blob/5564f82a536385388ecd9f723a4f6fe7063524b0/src/pynwb/epoch.py#L48

it seems this line is currently mostly important for converting the tags argument to a list if a single string is provided. (This issue might be also related to https://github.com/hdmf-dev/hdmf/issues/1211, where we discussed deprecating the ability to provide a single string for a ragged column.)

stephprince: I'll make an issue about this and this test can be removed if we change the behavior, but as of now will leave as is.

stephprince (approved):

PR #2035, opened 2025-02-07
Update documentation for streaming NWB files and add venv to .gitignore

bendichter:
## Motivation

The streaming tutorial needed an update.

Changes:
* I added context and non-context code snippets throughout, since this is a common source of confusion for users with less programming experience
* remfile:
    * moved to top
    * added caching feature
* fsspec:
    * has more supported backends
* ros3:
    * now provides automatic retries as of 2022
    * General performance of ros3 appears to be improved since we first tested it
    * Now easier to install, since the default installation for conda now supports ros3
* Choice of backend now less clear. I deleted that section for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,3 +80,5 @@ _version.py
 
 .core_typemap_version
 core_typemap.pkl
+
+venv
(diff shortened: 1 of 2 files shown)


oruebel (approved):
Looks good to me.

PR #2079, opened 2025-05-10
Add BaseImage and ExternalImage

### LINKED ISSUE #2042: [Feature]: Support ExternalImage for NWB Schema 2.9.0

See https://github.com/NeurodataWithoutBorders/nwb-schema/pull/604

_No response_

Add support for `BaseImage` and `ExternalImage` for NWB Schema 2.9.0

Yes.

### Code of Conduct

rly:
## Motivation

Fix #2042 

## Checklist


src/pynwb/base.py
```diff
@@ -422,26 +423,59 @@ def get_data_in_units(self):
         return np.asarray(self.data) * scale_factor + self.offset
 
 
+@register_class('BaseImage', CORE_NAMESPACE)
+class BaseImage(NWBData, ABC):
+    """
+    An abstract base type for image data. Parent type for Image and ExternalImage types.
+    """
+    __nwbfields__ = ('description', )
```
stephprince, line 433:
```suggestion
    __nwbfields__ = ('data', 'description', )
```

src/pynwb/io/base.py
```diff
@@ -125,3 +125,13 @@ def unit_carg(self, builder, manager):
...
+
+
+@register_map(Images)
+class ImagesMap(NWBContainerMapper):
+
+    def __init__(self, spec):
+        super().__init__(spec)
+
+        # Map the images attribute to the BaseImage neurodata type
+        self.map_spec('images', spec.get_neurodata_type('BaseImage'))
```
stephprince, line 137: Do we need tests for the mapper? These lines are showing up as not covered by codecov

stephprince (commented):
This looks good to me but I think I am not following when something should inherit from `BaseImage`, `Image`, or `ExternalImage`. Should `RGBImage`, `GrayscaleImage` and `RGBAImage` also inherit from `BaseImage` now so that they can support `ExternalImage`?

src/pynwb/base.py
rly: "data" should not have been a nwbfile in Image in the first place and should not be part of BaseImage as well. All subtypes of Data already have an internal field named `__data` that does not go through the `__nwbfields__` mechanism for generating getters and setters.

src/pynwb/io/base.py
rly: I think codecov screwed up with this flag in the PR files view. https://app.codecov.io/gh/NeurodataWithoutBorders/pynwb/pull/2079?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=NeurodataWithoutBorders shows the line being covered and if I add a breakpoint in this function and run the tests locally, I hit the breakpoint with `pytest tests/integration/hdf5/test_base.py::TestImagesIO`

rly:
We discussed this a bit in person. To restate and add to our discussion: `RGBImage` already exists and represents RGB data stored as a 3-dimensional array in some backend. `ExternalImage` represents an image stored as a file path. It does not describe what kind of image the data is. We could create an `ExternalRGBImage` and other subclasses, but those would not have any differentiating specifications. Contrary to what we discussed, I now think it would be valuable to hint to the user what kind of data the image is. The image file itself would contain much more information about it, e.g., bit depth, gamma correction, dimensions.

This is similar to the question of how to differentiate types of `ElectricalSeries`, e.g., wideband, LFP, AP, non-neural, that all have the same structure. I believe we decided at a recent meeting that it would be better to add an enumerated text field that indicates what type of data this is than to create a subclass for each possible type. I am still torn on that decision, but I think both are fine and we should go based on consensus.

I realize there are some parts of this [proposal](https://github.com/NeurodataWithoutBorders/nwb-schema/issues/603#issuecomment-2616464570) that are not in the schema. I'll create a PR on nwb-schema to address them and we can come back to here.

stephprince:
> This is similar to the question of how to differentiate types of ElectricalSeries, e.g., wideband, LFP, AP, non-neural, that all have the same structure. I believe we decided at a recent meeting that it would be better to add an enumerated text field that indicates what type of data this is than to create a subclass for each possible type. I am still torn on that decision, but I think both are fine and we should go based on consensus.

I am also torn on that decision still... (just linking for future reference the context of the previous discussion was for this issue: https://github.com/NeurodataWithoutBorders/nwb-schema/issues/617). I think in this case it is slightly awkward that RGBImage, GrayscaleImage, etc. are distinct types in one part of the schema but then will be only distinguished by the `image_mode` attribute in other parts of the schema... I am not opposed to the enumerated text field but am just wondering if the circumstances are different here.

Edit: After rereading the discussion in the related nwb-schema PR I see the advantage of an optional `image_mode` attribute to support different file types beyond the current Image data types in the schema.

stephprince (approved):

h-mayorquin:
To add to this discussion. One of the reasons for ExternalImage was that it would be avoid the proliferation of data types and therefore reduce our dev burden.

See simplification of data types section here:
https://github.com/NeurodataWithoutBorders/nwb-schema/issues/603

PR #2088, opened 2025-05-31
Add support for DeviceModel neurodata type

### LINKED ISSUE #2043: [Feature]: Support DeviceModel for NWB Schema 2.9.0

https://github.com/NeurodataWithoutBorders/nwb-schema/pull/608

_No response_

Make changes to `Device` and `DeviceModel` in PyNWB

Yes.

### Code of Conduct

rly:
## Motivation

Fix #2043

## How to test the behavior?
```
```

## Checklist


docs/gallery/domain/ecephys.py
```diff
@@ -80,15 +80,20 @@
 # The electrodes table references a required :py:class:`~pynwb.ecephys.ElectrodeGroup`, which is used to represent a
 # group of electrodes. Before creating an :py:class:`~pynwb.ecephys.ElectrodeGroup`, you must define a
 # :py:class:`~pynwb.device.Device` object using the method :py:meth:`.NWBFile.create_device`. The fields
-# ``description``, ``manufacturer``, ``model_number``, ``model_name``, and ``serial_number`` are optional, but
-# recommended.
+# ``description``, ``serial_number``, and ``model`` are optional, but recommended. The `~pynwb.device.DeviceModel`
```
stephprince:
```suggestion
# ``description``, ``serial_number``, and ``model`` are optional, but recommended. The :py:class:`~pynwb.device.DeviceModel`
```

docs/gallery/domain/ecephys.py
```diff
@@ -80,15 +80,20 @@
 # The electrodes table references a required :py:class:`~pynwb.ecephys.ElectrodeGroup`, which is used to represent a
```
stephprince, line 80: The class diagram image in this tutorial is no longer accurate (doesn't include DeviceModel and has manufacturer as an attribute of Device). Is there an easy way to update this?

docs/gallery/domain/ecephys.py
```diff
@@ -80,15 +80,20 @@
...
-# ``description``, ``manufacturer``, ``model_number``, ``model_name``, and ``serial_number`` are optional, but
-# recommended.
+# ``description``, ``serial_number``, and ``model`` are optional, but recommended. The `~pynwb.device.DeviceModel`
+# object stores information about the device model, which can be useful when searching a set of NWB files or a
+# data archive for all files that use a specific device model (e.g., Neuropixels probe).
+device_model = nwbfile.create_device_model(
+    name="Neurovoxels 0.99",
+    manufacturer="Array Technologies",
+    model_number="PRB_1_4_0480_123",
+    description="A 12-channel array with 4 shanks and 3 channels per shank",
```
stephprince, line 91: is there any distinct information that should be included in the description, or should it be the exact same as device?

stephprince (commented):
Looks good to me! A couple minor questions about documentation

docs/gallery/domain/ecephys.py
rly: In general, it should the DeviceModel description can be the same as the Device description, but Device description might hold more information specific to the instance, e.g., this particular array has a broken contact.

docs/gallery/domain/ecephys.py
rly: Good catch. It's not easy to edit these diagrams (made in Lucidchart) because my free account can no longer add content. What do you think about this one made in Mermaid (using mermaidchart.com)? Unfortunately Mermaid does not support customizing the direction of flow

Editor _ Mermaid Chart-2025-06-09-233037

stephprince: I think the original LucidChart ones were a little easier to read in terms of appearance and how the VectorData type was written, but this also works! 

I realized every image in this tutorial references the Device class in some way.... Maybe would be better to leave for now and make a separate issue to make sure all these diagrams are up to date and use a consistent display format?

docs/gallery/domain/ecephys.py
stephprince: Got it, that makes sense. I think for this tutorial it makes sense to keep as is then or add an example with the broken contact for the device instance description like you mentioned.

docs/gallery/domain/ecephys.py
rly: That's a good idea. I think powerpoint might be the best way to go here for full flexibility in how it is displayed.

stephprince (approved):

PR #2104, opened 2025-06-25
Improve docval for IndexSeries init

rly:
## Motivation

In coordination with https://github.com/NeurodataWithoutBorders/nwb-schema/pull/644

The goal is to not write `resolution`, `conversion`, and `offset` for `IndexSeries`.

This PR also updates `IndexSeries.__init__` so that `unit` cannot be passed in, since the schema requires a fixed value of "N/A" and currently in PyNWB, passing any different value will result in an error. This change improves the usability -- users do not need to provide `unit="N/A" during construction.

WIP. Removing `resolution`, `conversion`, and `offset` for `IndexSeries.__init__` does not work because when `resolution=None` is passed to `TimeSeries.__init__`, then resolution is set to the default value for resolution (-1) specified in `TimeSeries`.

## How to test the behavior?
```
```

## Checklist

- [ ] Is your contribution compliant with our coding style? This can be checked running `ruff check . && codespell` from the source directory.


rly:
We will punt this to the schema 2.10 release.

PR #2106, opened 2025-06-26
HERD inside NWBFile/schema

mavaylon1:
## Motivation

Allow HERD to be stored internally, but still allow external override
- [x] Test roundtrip internal HERD
- [x] Test link_resources
- [x] Test external HERD link through the IO herd path
- [x] Test Add functionality where you provide herd path and it calls link_resources

## How to test the behavior?
```
```

## Checklist

- [ ] Is your contribution compliant with our coding style? This can be checked running `ruff check . && codespell` from the source directory.


mavaylon1:
Future development:
On the surface to the user, HERD is a table and when you query before any I/O, you will see lists and tuples. Under the hood, HERD is written as a structured array and on read is no longer a list/tuple but a structured array. This was a bit surprising and so I spent some time to map over the array to a list on read in the get_item. However, this grew in complexity reaching `__swap_refs` in `AbstractH5TableDataset`. Without a clear easy solution, I am leaving this for future development. It is because of this the tests are structured the way they are in the asserts.

Issue #1984, opened 2024-11-07
[Documentation]: Add HERD Tutorial


stephprince:
### What would you like changed or added to the documentation and why?

With HERD being added to PyNWB it would be great to add an additional tutorial about using it PyNWB. There is an existing [tutorial on the HDMF docs](https://hdmf.readthedocs.io/en/stable/tutorials/plot_external_resources.html#sphx-glr-tutorials-plot-external-resources-py) about how it works, but the PyNWB tutorial would demonstrate how to use HERD within a PyNWB workflow. 

See related discussion here: https://github.com/NeurodataWithoutBorders/pynwb/issues/15#issuecomment-2463446885

@mavaylon1 has begun working on a tutorial here: https://github.com/NeurodataWithoutBorders/pynwb/pull/1781

### Do you have any interest in helping write or edit the documentation?

No.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/.github/CODE_OF_CONDUCT.rst)
- [X] Have you checked the [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [X] Have you ensured this change was not already [requested](https://github.com/NeurodataWithoutBorders/pynwb/issues)?

PR #2111, opened 2025-07-08
HERD Changes internal file

mavaylon1:
Allow HERD to be stored internally, but still allow external override

Future development:
On the surface to the user, HERD is a table and when you query before any I/O, you will see lists and tuples. Under the hood, HERD is written as a structured array and on read is no longer a list/tuple but a structured array. This was a bit surprising and so I spent some time to map over the array to a list on read in the get_item. However, this grew in complexity reaching __swap_refs in AbstractH5TableDataset. Without a clear easy solution, I am leaving this for future development. It is because of this the tests are structured the way they are in the asserts.

## How to test the behavior?
```
```

## Checklist


mavaylon1:
Future Development: 
After reviewing the HERD paper, there needs be work on updating the IO on write. In HDMFIO write, it will go through and create a HERD instance or use a HERD instance to resolve the TermSetWrapper calls that were used in the file. This is currently stored externally. I am keeping HDMF as original as possible, with everything being stored external as we discussed. That being said, there needs to be work updating the write in NWBHDFIO to write it internally.

mavaylon1:
Review Notes: This should be good to go. It needs nwb schema to be merged first.

bendichter:
## Status update on this PR

I've fixed two issues that were blocking this PR:

### 1. Fixed: `docval` crash on import (pynwb)
The `external_resources` parameter in `NWBFile.__init__`'s `@docval` used `'child': True` and `'required_name': 'external_resources'` — keys that belong in `__nwbfields__`, not `@docval`. This caused every CI job to fail with:
```
Exception: docval for external_resources: keys ['child', 'required_name'] are not supported by docval
```
Fixed by moving those keys to `__nwbfields__` and adding a proper `@docval` entry with `type: HERD`.

### 2. Fixed: nwb-schema merge conflict and hdmf-common-schema version (nwb-schema)
- Resolved the merge conflict on the `herd` branch of nwb-schema (PR https://github.com/NeurodataWithoutBorders/nwb-schema/pull/646).
- Updated the `hdmf-common-schema` submodule from 1.8.0 to 1.9.0. This is necessary because `nwb.file.yaml` references `HERD` via `neurodata_type_inc: HERD`, but in hdmf-common-schema 1.8.0, HERD lived in the `hdmf-experimental` namespace. The NWB `core` namespace only depends on `hdmf-common`, so it couldn't find HERD. In hdmf-common-schema 1.9.0 (https://github.com/hdmf-dev/hdmf-common-schema/pull/93), HERD was moved to `hdmf-common`, resolving this.

### Remaining blocker: hdmf release needed
The remaining 8 CI failures (`test_extension.py`) all hit:
```
ValueError: No specification for 'HERD' in namespace 'core'
```
This is because the released hdmf (4.3.1) bundles hdmf-common-schema 1.8.0. The hdmf `dev` branch already includes hdmf-common-schema 1.9.0, and all tests pass locally with it. Once hdmf releases a new version, this PR just needs to pin that version and CI should go green.

bendichter:
depends on https://github.com/NeurodataWithoutBorders/pynwb/pull/2141

src/pynwb/file.py
```diff
@@ -483,9 +487,16 @@ def __init__(self, **kwargs):
...
         args_to_set = popargs_to_dict(keys_to_set, kwargs)
+        external_resources = popargs('external_resources', kwargs)
         kwargs['name'] = 'root'
         super().__init__(**kwargs)
 
+        self._external_herd = None
+        self._internal_herd = None
+
+        if external_resources is not None:
+            self.external_resources = external_resources
```
rly:
```suggestion
            # NOTE: this sets self._internal_herd
            self.external_resources = external_resources
```

rly:
A HERD tutorial will be added as part of https://github.com/NeurodataWithoutBorders/pynwb/issues/1984

src/pynwb/file.py
```diff
@@ -570,6 +581,33 @@ def all_children(self):
...
+        self._external_herd = herd
+
+    @property
+    def external_resources(self):
+        """Return the HERD external resources object for this NWBFile.
+
+        If an external HERD has been linked via ``link_resources``, that object
+        is returned. Otherwise, the internal HERD set via ``__init__`` or the
+        setter is returned.
+        """
```
oruebel: With this setup, it looks like there is no way for a user to access the `self._internal_herd` once the `self._external_herd` is set. In general I think the whole `_interal_herd` and `external_herd` may be confusing. To clarify this a bit I would propose a few things:
1. Rename `self._external_herd` to `self._linked_herd`. This to a) make the terminology consistent with how it used in the user-facing functions, e.g,. `link_resources`, and b) to avoid overloading the term `external` because it used both in `external_resources` to mean linking to things outside of the file and `self._external_herd` to mean data stored outside of the file. 
2. Add a function `get_external_resources(bool linked)` where a user can explictly say if they want the internal HERD or the linked HERD, (or maybe even a join of both, but that's not essential)
3. Add documentation for what `self._external_herd` and `self._internal_herd` are. 
4. To be consistent, we may not want to use the term `herd` but just stick with `self._linked_external_resources` and `self._external_resources`

BTW, why is `_internal_herd` needed? Wouldn't I access that via `general_external_resources`?

oruebel (commented):
Overall this looks good. Just some inconsistency in terminology

src/pynwb/file.py
rly: Those suggestions make sense.

The automatic `__`-joined naming of fields only applies to untyped spec components. Allowing that for typed spec components that are then remapped in the NWBFile ObjectMapper would require a bit of refactoring because there would be multiple mappings for the same spec, and then order matters.

src/pynwb/file.py
```diff
@@ -570,6 +576,29 @@ def all_children(self):
...
+    def link_resources(self, external_resources):
+        """Link an external HERD object as the external resources for this file.
+
+        The linked HERD will not be written on export; the original HERD
+        (if any) is preserved in the exported file. Use
+        ``get_external_resources(linked=True)`` to access the linked HERD.
+        """
+        self._linked_external_resources = external_resources
+
+    def get_external_resources(self, linked=False):
```
oruebel: @rly I'm wondering whether it would make sense to have this functionality to support linked external resources on the `HERDManager` base class instead. It just seems like this would be more generally useful functionality rather than being something that would be specific to `NWBFile`. We can leave this for now as is and make it a separate issue.

oruebel (dismissed):

src/pynwb/file.py
rly: That makes sense. I hadn't thought of that. Let's just handle it now. I'll update the HDMF PR and this one.

rly (approved):

oruebel (approved):

PR #2113, opened 2025-07-14
Skip streaming test if offline

rly:
## Motivation

Tests should pass when I'm on a plane without internet.

## Checklist

### DIFF
--- a/tests/integration/hdf5/test_io.py
+++ b/tests/integration/hdf5/test_io.py
@@ -4,6 +4,7 @@ import numpy as np
 from h5py import File
 from pathlib import Path
 import tempfile
+import urllib.request
 
 from pynwb import NWBFile, TimeSeries, get_manager, NWBHDF5IO, validate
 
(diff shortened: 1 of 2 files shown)


tests/integration/hdf5/test_io.py
```diff
@@ -595,36 +595,38 @@ def test_can_read_file_invalid_hdf5_file(self):
...
-        read_nwbfile = NWBHDF5IO.read_nwb(path=s3_test_path)
-        assert read_nwbfile.identifier == "3f77c586-6139-4777-a05d-f603e90b1330"
-    
-        assert read_nwbfile.subject.subject_id == "1"
\ No newline at end of file
+        try:
+            read_nwbfile = NWBHDF5IO.read_nwb(path=s3_test_path)
+            assert read_nwbfile.identifier == "3f77c586-6139-4777-a05d-f603e90b1330"
+            assert read_nwbfile.subject.subject_id == "1"
+        except FileNotFoundError:
```
stephprince: Could also test the URL to be more specific for connectivity issues like in the ROS3 tests, but I think either works!

https://github.com/NeurodataWithoutBorders/pynwb/blob/f56b6f464fde0ce07716c9627bfaa4b1ede027ff/tests/integration/ros3/test_ros3.py#L23-L26

stephprince (dismissed):

tests/integration/hdf5/test_io.py
rly: Ah, I didn't realize that `urllib.request.urlopen` simply opens the connection but does not download the file. Will update

stephprince (approved):

PR #2124, opened 2025-08-15
Refactor CSS for assistant component: adjust media queries and container widths for better responsiveness

bendichter:
fix #2123 

## Checklist

- [x] Did you update CHANGELOG.md with your changes?
- [ ] Have you checked our [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [ ] Have you ensured the PR clearly describes the problem and the solution?
- [ ] Is your contribution compliant with our coding style? This can be checked running `ruff check . && codespell` from the source directory.
- [ ] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/NeurodataWithoutBorders/pynwb/pulls) for the same change?
- [ ] Have you included the relevant issue number using "Fix #XXX" notation where XXX is the issue number? By including "Fix #XXX" you allow GitHub to close issue #XXX when the PR is merged.

PR #2140, opened 2025-10-24
Update iframe source URL for NWB Assistant chat

magland:
I would like to update the URL for NWB Assistant to point to the new version. Feel free to [try it out](https://nwb-assistant.vercel.app/). The logic / prompts are the same... it's just the latest interface I am using shared [across multiple chat applications](https://github.com/magland/qp).


stephprince:
thanks for the update @magland !

One thing I noticed is that the new chat width does not expand to fill a wider screen display as the previous chat interface did, resulting in the "close assistant" tab dangling. Not sure where the best place to update that is?
Screenshot 2025-10-24 at 10 03 40 AM

magland:
@stephprince What you are showing in the screenshot is not an issue with the current PR. It seems some other changes were made elsewhere to the side panel... what you are showing in that screenshot is not consistent with what is seen in the deployed docs at

https://pynwb.readthedocs.io/en/stable/

Also, I have updated the url in this PR again to the latest location for this: https://nwb-assistant.neurosift.app

magland:
For now, I have redirected the old nwb assistant to the new one... so effectively it is giving https://nwb-assistant.neurosift.app

stephprince:
@magland, I see. I was looking at the readthedocs build from this PR but it looks like the tab issue I noticed was already merged to dev I believe as part of https://github.com/NeurodataWithoutBorders/pynwb/pull/2124. I will look into that as a separate issue.

stephprince (approved):

PR #2141, opened 2025-11-03
deprecate Python 3.9

bendichter:
EOL was Oct 31 2025

## Checklist


requirements-min.txt
```diff
@@ -1,7 +1,7 @@
...
+h5py==3.15.1
 hdmf==4.1.1
-numpy==1.24.0
-pandas==1.2.0
+numpy==2.2.6
+pandas==2.3.3
 python-dateutil==2.8.2
-platformdirs==4.1.0
\ No newline at end of file
+platformdirs==4.1.0
```
stephprince:
```suggestion
h5py==3.6.0
hdmf==4.1.1
numpy==1.24.0
pandas==1.3.4
python-dateutil==2.8.2
platformdirs==4.1.0
```

I think we should keep the minimum requirements as low as possible (and this requirements file should match the pyproject.toml dependency constraints). h5py 3.6.0 and pandas 1.3.4 looks like the first to have pre-built packages for python 3.10. I don't think the others should need to be updated.

.github/workflows/generate_test_files.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       matrix:
         include:
           - { name: pynwb-1.5.1, pynwb-version: "1.5.1", python-version: "3.8"}
-          - { name: pynwb-2.1.0, pynwb-version: "2.1.0", python-version: "3.9"}
+          - { name: pynwb-2.1.0, pynwb-version: "2.1.0", python-version: "3.10"}
```
stephprince:
```suggestion
          - { name: pynwb-2.1.0, pynwb-version: "2.1.0", python-version: "3.9"}
```

This workflow is to demo generating older test files for backwards compatibility. Technically pynwb 2.1.0 did not support python 3.10 so think this should remain 3.9.

stephprince (commented):
One other suggestion otherwise looks good to me! I will let @rly add any comments and can then update the PR required checks to use the new minimum test workflow versions before merging.

rly:
Thanks for this PR @bendichter ! It looks all good to me. This should be in a minor release. We should also explore adding Python 3.14 support in a separate PR.

rly:
We'll make a bugfix release imminently and then do a minor release with this PR deprecating python 3.9 shortly after

h-mayorquin:
Python 3.9 did not allow me to use typing constructions on #2164 

https://github.com/NeurodataWithoutBorders/pynwb/pull/2164/commits/b932db49a6c53f521c1b3e042178935e0667bbf3

bendichter:
@rly this is ready to go

rly:
Can you update the workflows so that all minimum tests on macos use macos-15-intel, and set the minimum h5py version to 3.6.0 (same as hdmf)?

rly (approved):

PR #2146, opened 2025-11-24
Add time duration methods

h-mayorquin:
## Motivation

Previously discussed in slack. One difficulty that users have when building an NWBFile is to understand the overall time organization of the data. There is a push for providing better reporting of this and these methods would be useful in this context.

## Checklist

### DIFF
--- a/src/pynwb/base.py
+++ b/src/pynwb/base.py
@@ -398,6 +398,57 @@ class TimeSeries(NWBDataInterface):
         else:
             return np.arange(len(self.data)) / self.rate + self.starting_time
 
+    def get_starting_time(self):
+        """
+        Get the starting time of this TimeSeries in seconds.
+
+        Returns
+        -------
+        float or None
+            The starting time in seconds, or None if the TimeSeries has no data.
+        """
+        if self.num_samples is None or self.num_samples == 0:
+            return None
+        if self.fields.get('timestamps'):
+            return float(self.timestamps[0])
+        else:
+            return self.starting_time
+
+    def get_duration(self):
+        """
+        Get the duration of this TimeSeries in seconds.
+
+        Returns the time span from the first sample to the last sample.
+        For a single sample, returns 0.
+
+        Returns
+        -------
+        float or None
+            The duration in seconds, or None if the TimeSeries has no data.
+
+        Notes
+        -----
+        For rate-based TimeSeries: duration = (n - 1) / rate
+        For timestamp-based TimeSeries: duration = timestamps[-1] - timestamps[0]
+
+        The duration represents the time span between sample times, not the total
+        recording time. If you need to account for the last sample's duration
+        (e.g., for continuous recordings), add 1/rate manually.
+        """
+        n = self.num_samples
+        if n is None or n == 0:
+            return None
+
+        if n == 1:
+            return 0.0
+
+        if self.fields.get('timestamps'):
+            timestamps = self.timestamps
+            return float(timestamps[-1] - timestamps[0])
+        else:
+            # Rate-based
+            return (n - 1) / self.rate
+
     def get_data_in_units(self):
         """
         Get the data of this TimeSeries in the specified unit of measurement, applying the conversion factor and offset:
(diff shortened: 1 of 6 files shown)


bendichter:
This looks great to me, thanks for working on this, @h-mayorquin ! I'd be interested in @oruebel 's thoughts as well

src/pynwb/base.py
```diff
@@ -398,6 +398,57 @@ def get_timestamps(self):
...
+        """
+        Get the starting time of this TimeSeries in seconds.
+
+        Returns
+        -------
+        float or None
+            The starting time in seconds, or None if the TimeSeries has no data.
+        """
+        if self.num_samples is None or self.num_samples == 0:
+            return None
```
oruebel: If `self.starting_time` is set, why would we return None. If the TimeSeries is empty and has a starting time, then the starting time is still valid and the duration would be 0 I.e., I think the logic should probably be more like this:

```
       if self.starting_time is not None:
             return self.starting_time
       elif self.num_samples > 0:
             return self.timestamps[0]  
       else:
             return None  
```

src/pynwb/base.py
```diff
@@ -398,6 +398,57 @@ def get_timestamps(self):
...
+        Notes
+        -----
+        For rate-based TimeSeries: duration = (n - 1) / rate
+        For timestamp-based TimeSeries: duration = timestamps[-1] - timestamps[0]
+
+        The duration represents the time span between sample times, not the total
+        recording time. If you need to account for the last sample's duration
+        (e.g., for continuous recordings), add 1/rate manually.
+        """
+        n = self.num_samples
```
oruebel: It looks like `n` is not being modified anywhere, but it is  just a copy of `self.num_samples`. In that case I would suggest using self.num_samples instead throughout to avoid possible confusion

src/pynwb/base.py
```diff
@@ -398,6 +398,57 @@ def get_timestamps(self):
...
+        if n is None or n == 0:
+            return None
```
oruebel:
```suggestion
        if n is None or n == 0:
            return 0
```
I think a duration should be 0 if the TimeSeries is empty

src/pynwb/epoch.py
```diff
@@ -81,3 +81,40 @@ def __calculate_idx_count(self, start_time, stop_time, ts_data):
...
+        Get the earliest start time across all intervals in this TimeIntervals table.
+
+        Returns
+        -------
+        float or None
+            The earliest start time in seconds, or None if the table is empty.
+        """
+        if len(self) == 0:
+            return None
+        import numpy as np
```
oruebel: I'm not sure whether this really needs to be a function-level import. I think it would be  Ok to import numpy on the module level here.  @stephprince thoughts?

oruebel:
> I'd be interested in @oruebel 's thoughts as well

This is looks overall good to me. I added a few small suggestions, but overall this seems fine to me. I'll let @stephprince do final review and approval.

src/pynwb/epoch.py
h-mayorquin: I moved this up.

src/pynwb/base.py
h-mayorquin: Sure. This is changed now.

src/pynwb/base.py
h-mayorquin: This makes sense. I changed it.

src/pynwb/epoch.py
```diff
@@ -81,3 +83,39 @@ def __calculate_idx_count(self, start_time, stop_time, ts_data):
...
+            The duration in seconds, or None if the table is empty.
+
+        Notes
+        -----
+        The duration represents the time span from the earliest interval start to the
+        latest interval stop, not the sum of individual interval durations.
+        """
+        if len(self) == 0:
+            return None
+        import numpy as np
```
rly:
```suggestion
```

rly (approved):

rly:
Looks good to me. Thanks @h-mayorquin !

bendichter:
this looks great. Once this is released, let's modify NWB inspector to use these methods.

Let's also port the Units table timing logic that is currently in the NWB Inspector library into PyNWB and do the same there.

Issue #2094, opened 2025-06-10
Update charts in docs to reflect DeviceModel usage


rly:
> I think the original LucidChart ones were a little easier to read in terms of appearance and how the VectorData type was written, but this also works! 
> I realized every image in this tutorial references the Device class in some way.... Maybe would be better to leave for now and make a separate issue to make sure all these diagrams are up to date and use a consistent display format?

_Originally posted by @stephprince in https://github.com/NeurodataWithoutBorders/pynwb/pull/2088#discussion_r2136679045_

subh123321:
Hello, I have set up pynwb locally, and all tests are passing.
I want to work on updating the documentation charts for this issue as my first contribution while preparing for GSoC 2026.
Please let me know if this scope is appropriate or if you would like me to focus on a specific set of diagrams.

rly:
@subh123321 Thanks for taking on this issue. What GSoC 2026 projects are you referring to? As far as I am aware, the NWB team is not participating in GSoC 2026...

PR #2158, opened 2026-01-19
Update ecephys electrodes diagram to reflect DeviceModel usage

subh123321:
This PR updates the UML-like electrodes diagram in the ecephys tutorial to reflect
current DeviceModel usage. Shared device metadata (e.g., manufacturer) is shown on
DeviceModel, with Device referencing the model.

This follows reviewer guidance related to #2094.


rly:
This looks good, but the background should be transparent instead of dark gray. Could you please upgrade the background? The new text should also not have a drop-shadow. Could you please remove that? Thank you, and sorry for the delay in response.

PR #2164, opened 2026-02-04
Add `get_duration` and `get_starting_time` for Units table

h-mayorquin:
A [request](https://github.com/NeurodataWithoutBorders/pynwb/pull/2146#issuecomment-3828517801) from @bendichter on #2146

### DIFF
--- a/tests/unit/test_misc.py
+++ b/tests/unit/test_misc.py
@@ -404,3 +404,59 @@ class UnitsTests(TestCase):
         ut.add_unit(spike_times=[])
         ut.add_unit(spike_times=[])
         self.assertIsNone(ut.get_duration())
+
+    def test_get_starting_time_empty_first_unit(self):
+        """Test get_starting_time when first unit is empty."""
+        ut = Units()
+        ut.add_unit(spike_times=[])  # empty first unit
+        ut.add_unit(spike_times=[5.0, 6.0])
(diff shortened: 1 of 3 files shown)


tests/unit/test_misc.py
```diff
@@ -298,3 +298,95 @@ def test_waveform_attrs(self):
...
+        self.assertEqual(ut.get_starting_time(), 1.0)
+
+    def test_get_duration_units_not_chronological(self):
+        """Test get_duration when earliest/latest spikes are in middle unit."""
+        ut = Units()
+        ut.add_unit(spike_times=[5.0, 6.0, 7.0])
+        ut.add_unit(spike_times=[1.0, 2.0, 20.0])  # has BOTH earliest (1.0) AND latest (20.0)
+        ut.add_unit(spike_times=[10.0, 11.0, 12.0])
+        # Duration from earliest (1.0) to latest (20.0) = 19.0
+        self.assertEqual(ut.get_duration(), 19.0)
```
bendichter, line 392: these are great. Can you also please add tests where the empty unit is the first unit where where the empty unit is the last unit? And also where there are multiple empty units in a row?

h-mayorquin: Added. I also added other tests to test defensive branches that were not included in the coverage.

bendichter (approved):

PR #2174, opened 2026-03-24
Add example of setting Units.resolution in the ecephys tutorial

h-mayorquin:
## Motivation

At the inspector we are working on `check_units_resolution_is_set` check (NeurodataWithoutBorders/nwbinspector#685), which flags `Units` tables that contain `spike_times` but have no `resolution` set. Users running the inspector on files created by following our tutorials would get this warning with no guidance on how to fix it.

I have added a short section to the ecephys tutorial that sets `nwbfile.units.resolution` right after the existing `add_unit` loop, with a brief explanation of what the field means and why it matters. I kept the change minimal to avoid restructuring the existing tutorial flow.

## Checklist

### DIFF
--- a/docs/gallery/domain/ecephys.py
+++ b/docs/gallery/domain/ecephys.py
@@ -348,6 +348,14 @@ for n_units_per_shank in range(n_units):
 # data, typically ``1 / sampling_rate`` of the acquisition system (i.e., the smallest measurable difference
 # between two spike times, in seconds). Setting it helps downstream users judge whether fine-timescale
 # analyses are appropriate for the dataset.
+#
+# When creating the :py:class:`~pynwb.misc.Units` table directly, you can pass it to the constructor:
+#
+# .. code-block:: python
+#
+#     units = Units(name="units", resolution=1 / 30000)  # 30 kHz recording system
+#
+# Since the tutorial uses :py:meth:`.NWBFile.add_unit`, we set it after the fact:
 
 nwbfile.units.resolution = 1 / res  # resolution in seconds (1 / sampling rate)
 
--- a/docs/gallery/domain/ecephys.py
+++ b/docs/gallery/domain/ecephys.py
@@ -343,6 +343,14 @@ for n_units_per_shank in range(n_units):
     spike_times = np.where(np.random.rand((res * duration)) < (firing_rate / res))[0] / res
     nwbfile.add_unit(spike_times=spike_times, quality="good")
 
+####################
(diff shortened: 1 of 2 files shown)


docs/gallery/domain/ecephys.py
```diff
@@ -343,6 +343,22 @@
...
+# between two spike times, in seconds). Setting it helps downstream users judge whether fine-timescale
+# analyses are appropriate for the dataset.
+#
+# When creating the :py:class:`~pynwb.misc.Units` table directly, you can pass it to the constructor:
+#
+# .. code-block:: python
+#
+#     units = Units(name="units", resolution=1 / 30000)  # 30 kHz recording system
+#
+# Since the tutorial uses :py:meth:`.NWBFile.add_unit`, we set it after the fact:
```
rly:
```suggestion
```

I don't think we need to show how to set it on the constructor. That is a rare edge case.

h-mayorquin: how is that an edge case? how are you thinking about it?

rly: I think most users would use `nwbfile.add_unit()` and `nwbfile.add_unit_column()` as we recommend here, rather than setting `nwbfile.units = Units(name="units", ...)` or creating a `Units` object to add elsewhere. What do you think?

h-mayorquin: We use the pattern in neuroconv:
https://github.com/catalystneuro/neuroconv/blob/2ddf1744208050a6229ad67d0e45f634253ae69e/src/neuroconv/tools/spikeinterface/spikeinterface.py#L2041-L2064


But yes, for the units table we don't realy use the Units constructor so it will somehow off. 

I will remove it.

h-mayorquin: As an aside: looking at the tutorials, the constructor is used for pretty much everything (`ElectricalSeries`, `LFP`, `DecompositionSeries`, etc.) except the "canonical" dynamic tables on `NWBFile` (`units`, `trials`, `epochs`, `electrodes`, `invalid_times`), which always go through the `add_X` / `add_X_column` methods. The only place the gallery constructs a `TimeIntervals` directly is for custom (non-canonical) interval tables like `sleep_stages` in `plot_timeintervals.py`.

That convention makes sense, but I don't think it is written down anywhere, which is why "edge case" didn't immediately click for me. I remember being confused myself about how to add a constructed `Units` table to an `NWBFile`.

rly (approved):

rly:
Thanks @h-mayorquin !

PR #2121, opened 2025-08-13
Remove unnecessary calls to deepcopy(get_type_map)

rly:
## Motivation

Fix #1958. Closes #2075. Similar approach as in #2075. Adds a `copy` argument to `get_type_map` that is default True but can be set to False. The False setting is used when creating a setter for each attribute in each NWB container class. Previously, a deepcopy was performed when creating every setter, which significantly hurt performance.

This PR deprecates calling get_type_map with any 'extensions' arguments, a use case I have never seen. IMO it's easy for the rare user to just call load_namespaces on the type map than us maintain all this extra convenience logic that even we don't use. See similar PR for hdmf-common in https://github.com/hdmf-dev/hdmf/pull/1302.

It is possible that deepcopying of typemaps in `get_type_map` can be removed entirely. That will be investigated separately as part of #2120.

## How to test the behavior?
```
Profile performance, run pytest
```

## Checklist

- [x] Did you update CHANGELOG.md with your changes?
- [x] Have you checked our [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [x] Have you ensured the PR clearly describes the problem and the solution?
- [x] Is your contribution compliant with our coding style? This can be checked running `ruff check . && codespell` from the source directory.
- [x] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/NeurodataWithoutBorders/pynwb/pulls) for the same change?
- [x] Have you included the relevant issue number using "Fix #XXX" notation where XXX is the issue number? By including "Fix #XXX" you allow GitHub to close issue #XXX when the PR is merged.

PR #2189, opened 2026-05-13
Add method to find objects of a given neurodata type

### LINKED ISSUE #560: Programmatically find all the time series in any NWB file

h-mayorquin:
This PR tries to complete the work started on PR #1737. It implements a couple of the design points raised in the previous discussion:

* The method is defined on `NWBContainer` rather than on `NWBFile` directly, so it works on any sub-container (`ProcessingModule`, `LFP`, etc.) and is correctly scoped to that container's subtree.
* String-to-class resolution goes through `pynwb.get_type_map(copy=False)`, the live process-wide TypeMap (available since #2121), rather than `self.io.manager.type_map`; the previous approach required the container to have been read or written through an IO, captured a snapshot of the TypeMap at IO-construction time, and so failed on in-memory containers and missed extensions loaded later in the same process.

Fix #560

## Checklist


oruebel:
@h-mayorquin this looks like something we probably would want to add on `Container` in HDMF instead since this is generic functionality that is independent of NWB specifics

h-mayorquin:
@oruebel

The original [PR](https://github.com/NeurodataWithoutBorders/pynwb/pull/1737) supports matching both by class and by string (e.g. `collect_descendants_of_type("ElectricalSeries")`). For matching the strings we need the utility `pynwb.get_type_map()`. Naively moving this to hdmf would create a circular dependency, as `hdmf.common.get_type_map()` does not have all the mappings that `pynwb.get_type_map()` has (e.g. "ElectricalSeries").

We could support the basic matching with class only in hdmf and then a method here can use that functionality and add the string matching, concretely:

```python
# hdmf (no nwb type map knowledge)
def collect_descendants_of_type(self, type_):
    return [obj for obj in self.all_children()
            if obj is not self and isinstance(obj, type_)]
```

```python
# pynwb (add string resolution support)
def collect_descendants_of_type(self, neurodata_type):
    if isinstance(neurodata_type, str):
        neurodata_type = self._get_type_map().get_dt_container_cls(neurodata_type)
    return super().collect_descendants_of_type(neurodata_type)
```

One thing that I don't like is that the hdmf loses support for the pattern that @rly described (filtering objects whose Python class was auto-generated from a cached extension namespace at read time), and we have two slightly different functions in pynwb and hdmf which might be surprising and confusing. This compels me to ask: besides the logical place for a thing to be, what would be the concrete gain of having a reduced version on hdmf?

PR #2190, opened 2026-05-14
Add remote-read support to `read_nwb`

### LINKED ISSUE #2150: [Feature]: Support for remote paths in `pynwb.read_nwb`

Follow up to #1974.

Add support for remote paths in `pynwb.read_nwb` which requires a change to `NWBHDF5IO.can_read`. This should be doable. We may want to change can_read to return the read h5py object so that we are creating two h5py objects - one in `can_read` and one in `read_nwb`.

Trying to simplify how users open NWB files across backends and data sources

See above

Yes.

### Code of Conduct

h-mayorquin:
Closes #2150.

The main contribution of this PR is enabling `pynwb.read_nwb(path)` to read nwbfiles in remote public locations such as public DANDI Archive assets served over HTTPS (e.g. `https://dandiarchive.s3.amazonaws.com/...`) or any publicly-readable S3, GCS, or HTTPS object store.

To move this forward I have decided not to use the `can_read` methods for dispatch, like I did in #1994 for local file support. What we have here is simple URL pattern matching: routing based on conventions like the `.zarr` suffix and DANDI's `/zarr/<uuid>/` URL layout, without opening the file. This avoids the performance cost of reading the file twice and in addition does not overpromise (can_read implies a strong contract that is hard-to-deliver for all the complexity of remote files).

Private files still can be accessed in a lot of cases. For example, a private S3 bucket read with `s3://` when AWS credentials are already configured in the environment (`AWS_PROFILE` or `~/.aws/credentials`), a `gs://` object when `GOOGLE_APPLICATION_CREDENTIALS` is set, or an `abfs://` path via Azure managed identity; fsspec picks up each backend's default credential chain automatically. 

That said, I did not want to make the signature more complex by adding login configuration parameters as that would be against the original spirit of this function. The original `read_nwb` design from #1974 was as simple as possible without config. Power-user scenarios (e.g. forced h5py ROS3 driver or custom S3-compatible endpoints) continue to require dropping to `NWBHDF5IO` or `NWBZarrIO` directly and I think this is where private access should be.

The remote-Zarr test depends on the `resolve_ref` self-reference fix for fsspec stores from [hdmf-dev/hdmf-zarr#348](https://github.com/hdmf-dev/hdmf-zarr/pull/348), which I opened upstream. That fix is now released in hdmf-zarr 0.13.0, so no special pin is needed and the test runs against a normal `hdmf-zarr` install.

I am also fixing a pre-existing scheme bug in `NWBHDF5IO.read_nwb`'s streaming branch: the fsspec filesystem was hard-coded to `"http"` regardless of URL scheme, so `s3://`, `gs://`, and `abfs://` paths silently failed for non-HTTP backends.

## Checklist


tests/integration/io/test_read.py
```diff
@@ -67,11 +74,50 @@ def test_read_invalid_file(self):
...
\ No newline at end of file
+                read_nwb(path=path)
+
+    @unittest.skipIf(not HAVE_FSSPEC, "fsspec not installed")
+    def test_read_nwb_anonymous_remote_hdf5(self):
+        """Test reading an anonymous public HDF5 NWB file over HTTPS through fsspec."""
+        url = (
+            "https://dandiarchive.s3.amazonaws.com/blobs/11e/c89/"
+            "11ec8933-1456-4942-922b-94e5878bb991"
+        )
```
oruebel, line 93: `_REMOTE_SCHEMES =`  lists a broad range of different URL type options. Do we need to test all of those variants or is it sufficient to test just the https version here?

oruebel (dismissed):

tests/integration/io/test_read.py
h-mayorquin: All the schemes in `_REMOTE_SCHEMES` share one code path: `read_nwb` extracts the scheme and hands it to `fsspec.filesystem(scheme)`, which dispatches to the right backend (`s3fs`, `gcsfs`, `adlfs`, etc.). So from pynwb's side there's nothing scheme-specific to test beyond "is it remote, and what's the scheme."

The `https://` test already covers our logic end to end (remote detection, scheme extraction, open through fsspec, read into an `NWBFile`). A test per scheme would just re-test fsspec's backends and would need each backend package installed plus a live public file, making them network-flaky. 

I'd keep the single HTTPS test and let fsspec own the per-backend behavior.

oruebel: Sounds good

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
...
+- Added remote-read support to `pynwb.read_nwb`. The function now accepts remote URLs (`s3://`, `gs://`, `abfs://`, `https://`, etc.) and dispatches to the right backend based on the URL: `.zarr` suffixes (and DANDI Zarr assets under `/zarr/`) are read with `NWBZarrIO`, everything else with `NWBHDF5IO`. Remote files are opened through `fsspec`, which now uses the URL's actual scheme instead of the previous hardcoded `fsspec.filesystem("http")` that mishandled non-HTTP schemes. @h-mayorquin [#2190](https://github.com/NeurodataWithoutBorders/pynwb/pull/2190)
```
oruebel: Please move the changlog entry to the 4.0.1 since 4.0.0 is already released

oruebel:
Other than moving the changelog entry, this looks good to me. I'll let @rly handle merge.

CHANGELOG.md
h-mayorquin: changed

oruebel (dismissed):

rly:
I added a simple regression test to make sure reading an `s3://` path does not use `http://`. The test uses a mocked `fsspec.filesystem` so that it doesn't need the network or add an optional s3fs dependency.

rly:
Looks good to me. Thanks @h-mayorquin !

rly (dismissed):

rly:
I also added a brief note on using `read_nwb` on S3 urls without credentials.

rly (approved):

PR #1781, opened 2023-10-07
Add HERD gallery to pynwb with streaming

mavaylon1:
## Motivation

What was the reasoning behind this change? Please explain the changes briefly.

Now that `HERD` is in pynwb, it should have its own gallery. It should also include an example that handles streaming a dandiset in order to add metadata annotations. 

- [ ] Add small description of the dandiset
- [ ] Replace the `HerdManager` with `NWBFile`
- [ ] Update the examples to use pynwb objects to make it more practical

## How to test the behavior?
```
Show how to reproduce the new behavior (can be a bug fix or a new feature)
```
Run the doc.

## Checklist

- [ ] Did you update CHANGELOG.md with your changes?
- [ ] Have you checked our [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [ ] Have you ensured the PR clearly describes the problem and the solution?
- [ ] Is your contribution compliant with our coding style? This can be checked running `flake8` from the source directory.
- [ ] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/NeurodataWithoutBorders/pynwb/pulls) for the same change?
- [ ] Have you included the relevant issue number using "Fix #XXX" notation where XXX is the issue number? By including "Fix #XXX" you allow GitHub to close issue #XXX when the PR is merged.

PR #2200, opened 2026-06-23
Add HERD tutorial and DANDI streaming example

rly:
## Motivation

HERD (HDMF External Resources Data Structure) is a stable `hdmf-common` type, and NWB schema 2.10.0 added an optional slot for it at `/general/external_resources`. PyNWB had no HERD tutorial, and the only existing guidance (in hdmf) stopped at `HERD.from_zip(...)` without showing how to access the loaded data (hdmf-dev/hdmf#1325).

This adds:

- **`docs/gallery/general/plot_external_resources.py`** (executed): a concise, NWB-focused tutorial that creates a HERD, annotates NWB objects (`Subject.species`, an electrodes-table column against the Allen Mouse Brain CCFv3) with `add_ref`, stores the HERD inside the NWB file, round-trips it, and inspects the loaded data via `to_dataframe()`, `get_object_entities()`, and the individual interlinked tables. It links out to the comprehensive hdmf HERD tutorial rather than duplicating it.
- **`docs/gallery/general/resources_streaming.py`** (rendered but not executed during the offline docs/example build): a companion example that annotates multiple NWB files streamed from a DANDI dandiset with a single shared HERD, then reloads the saved HERD with `from_zip` to add and persist a further annotation.

It also adds:

- **`NWBFile.get_external_resources()`** (`src/pynwb/file.py`): a get-or-create accessor that returns the file's existing HERD or constructs and attaches a new empty pynwb HERD (carrying the pynwb type map) when the file has none, mirroring `get_intracellular_recordings()`. The `external_resources` attribute continues to return the HERD or `None` without creating one. The tutorial uses this accessor so it does not clobber a HERD read from disk. Covered by a new unit test in `tests/unit/test_resources.py`.
- A dedicated **`run-streaming-example`** job in `.github/workflows/run_all_tests.yml` that installs the streaming dependencies and runs `resources_streaming.py` once on the daily/release schedule, since that example is excluded from the offline example tests (it streams from the DANDI Archive). `test.py` gains a `streaming_examples` exclusion list for this and cleans up the tutorial's output file.

It removes the now-dead `"HERD is experimental"` warning filter from `tests/unit/test_resources.py`, since HERD no longer emits that warning.

This supersedes #1781, whose `plot_resources.py` was an unfinished port of the hdmf tutorial that predates HERD becoming non-experimental and storable in an NWB file. The streaming example here salvages and updates that PR's DANDI idea, so its author @mavaylon1 is credited as a co-author.

## How to test the behavior?

```
# Run the executed tutorial directly (cleans up after itself):
python docs/gallery/general/plot_external_resources.py

# Run the streaming example (streams from the DANDI Archive, needs network access):
python docs/gallery/general/resources_streaming.py

# Run the HERD unit tests:
pytest tests/unit/test_resources.py
```

The streaming example reads over the network and is not executed during the offline docs/example build; it runs in the dedicated `run-streaming-example` job in `run_all_tests.yml`.

## Checklist


rly:
@oruebel @bendichter two small parts of these tutorials require changes in HDMF (as mentioned above). I would appreciate your feedback on the tutorials while those issues get resolved.

bendichter:
https://pynwb--2200.org.readthedocs.build/en/2200/tutorials/general/plot_external_resources.html

https://pynwb--2200.org.readthedocs.build/en/2200/tutorials/general/resources_streaming.html

docs/gallery/general/plot_external_resources.py
```diff
@@ -0,0 +1,182 @@
...
+for _ in range(4):
+    nwbfile.add_electrode(location="VISp", group=electrode_group)
+
+nwbfile.external_resources.add_ref(
+    container=nwbfile.electrodes,
+    attribute="location",
+    key="VISp",
+    entity_id="385",
+    entity_uri="https://api.brain-map.org/api/v2/data/Structure/385.json",
+)
```
bendichter: Would we want to use https://purl.brain-bican.org/ontology/mbao/MBA_385 instead?

docs/gallery/general/resources_streaming.py
```diff
@@ -0,0 +1,158 @@
...
+            entity = herd.get_entity(entity_id="NCBI_TAXON:10090")
+            if entity is not None:
+                # the entity is already in the HERD, so reuse it and keep its existing URI
+                entity_uri = None
+            else:
+                # the entity is not yet in the HERD, so provide its URI to create it
+                entity_uri = "https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=10090"
+            herd.add_ref(
+                file=read_nwbfile,
+                container=read_nwbfile.subject,
```
bendichter: I don't understand why it is necessary to provide the file arg here. Can't you resolve the file from the container directly?

docs/gallery/general/resources_streaming.py
```diff
@@ -0,0 +1,158 @@
...
+
+            # reference the subject species
+            species = read_nwbfile.subject.species  # "Mus musculus"
+            entity = herd.get_entity(entity_id="NCBI_TAXON:10090")
+            if entity is not None:
+                # the entity is already in the HERD, so reuse it and keep its existing URI
+                entity_uri = None
+            else:
+                # the entity is not yet in the HERD, so provide its URI to create it
+                entity_uri = "https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=10090"
```
bendichter: What happens if you just provide the entity uri again as the same string? Does it create a duplicate row?

bendichter:
Can we talk about how we are using entity_id?  In some cases, you have a prefix:suffix compact ID. In some cases, this is resolvable in [identifiers.org](http://identifiers.org/), e.g. ROR:013sk6x84: https://identifiers.org/resolve?query=ROR:013sk6x84 resolves to https://ror.org/013sk6x84. My understanding is this is how the prefix:suffix compact identifiers are meant to be used. DANDI works the same way — http://identifiers.org/DANDI:000015 resolves to https://dandiarchive.org/dandiset/000015.

However there are some cases where this isn’t working. For species, we are using NCBI_TAXON:10090, which does not resolve (https://identifiers.org/resolve?query=NCBI_TAXON:10090). However, taxonomy:10090 does resolve: https://identifiers.org/resolve?query=taxonomy:10090 goes to several links that all point to the mouse.

And then there’s the neural data, where the entity_id is simply 385, with no prefix and no clear way to map this to any external resource without the uri. 

It looks like there are 3 separate uses for entity_id here, and only one of them really makes sense to me, where the id is resolvable at identifiers.org

docs/gallery/general/resources_streaming.py
bendichter: we want to populate this automatically and error out when file not present

docs/gallery/general/resources_streaming.py
bendichter: let's check to see if these cases are handled. We want to ensure that the tables are always normalized, so adding an identical entry should not duplicate a row

bendichter:
It looks like https://bioregistry.io/ resolves everything we want so far:
* NCBITaxon: (species)
* ROR: (organizations)
* ORCID: (people)
* MBA: (mouse brain atlast)
* UBERON: (cross-species brain atlas)
* HBA: (human brain atlas)

it also resolves DANDI:

bendichter:
We need a page that is an explanation of best practices for using HERD in the context of NWB files. I'm going to draft it here, as I think this is where it should go, but I would also be OK with putting this in NWB Inspector or nwb.org.

docs/gallery/general/plot_external_resources.py
```diff
@@ -0,0 +1,182 @@
...
+# -------------------------------
+# Use :py:meth:`~hdmf.common.resources.HERD.add_ref` to add a row that links a key on an object to an
+# external entity. Here we link the subject's species to the NCBI Taxonomy entry for *Mus musculus*.
+# Because the subject is already part of the file, the ``file`` argument is resolved automatically
+# from the parent hierarchy and can be omitted.
+
+nwbfile.external_resources.add_ref(
+    container=nwbfile.subject,
+    key=nwbfile.subject.species,
+    entity_id="NCBI_TAXON:10090",
```
bendichter:
```suggestion
    entity_id="NCBITAXON:10090",
```

docs/gallery/general/resources_streaming.py
```diff
@@ -0,0 +1,158 @@
...
+
+herd = HERD()
+for url in tqdm(urls):
+    with fs.open(url, "rb") as f, h5py.File(f) as h5_file:
+        with NWBHDF5IO(file=h5_file) as io:
+            read_nwbfile = io.read()
+
+            # reference the subject species
+            species = read_nwbfile.subject.species  # "Mus musculus"
+            entity = herd.get_entity(entity_id="NCBI_TAXON:10090")
```
bendichter:
```suggestion
            entity = herd.get_entity(entity_id="NCBITAXON:10090")
```

docs/gallery/general/resources_streaming.py
```diff
@@ -0,0 +1,158 @@
...
+                # the entity is already in the HERD, so reuse it and keep its existing URI
+                entity_uri = None
+            else:
+                # the entity is not yet in the HERD, so provide its URI to create it
+                entity_uri = "https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=10090"
+            herd.add_ref(
+                file=read_nwbfile,
+                container=read_nwbfile.subject,
+                key=species,
+                entity_id="NCBI_TAXON:10090",
```
bendichter:
```suggestion
                entity_id="NCBITAXON:10090",
```

docs/gallery/general/plot_external_resources.py
```diff
@@ -0,0 +1,182 @@
...
+)
+
+###############################################################################
+# Create a HERD and attach it to the file
+# ---------------------------------------
+# Create a :py:class:`~pynwb.resources.HERD` and assign it to the ``external_resources`` field of the
+# :py:class:`~pynwb.file.NWBFile`.
+
+nwbfile.external_resources = HERD()
+
```
oruebel, line 59: What happens here if the `nwbfile` already has an existing `HERD` (e.g., if the file was read from disk). 

We already use the pattern, where e.g., `nwbfile.intracellular_recordings` returns the object that exists (and returns None if it is missing) and `nwbfile.get_intracellular_recordings()` either returns the existing object or constructs a new one if it is missing. 

Since there is only one HERD per file, I think we can simply use the same pattern here. I.e., use `nwbfile.external_resources` to access HERD and have `nwbfile.get_external_resources` construct a new HERD if it is missing.

docs/gallery/general/plot_external_resources.py
```diff
@@ -0,0 +1,182 @@
...
+# Alternative: store a HERD outside an NWB file
+# ---------------------------------------------
+# A HERD can also be saved independently of an NWB file as a zip archive of the underlying tables
+# using :py:meth:`~hdmf.common.resources.HERD.to_zip`, and read back with
+# :py:meth:`~hdmf.common.resources.HERD.from_zip`. This is useful when external resources span
+# multiple files; see :ref:`external_resources_streaming` for an example that annotates many NWB
+# files with a single HERD. For the full HERD API, see the
+# `HDMF HERD tutorial <https://hdmf.readthedocs.io/en/stable/tutorials/plot_external_resources.html>`_.
+
+os.remove(filename)
```
oruebel: Is it necessary to remove the file as part of the tutorial or does the build/pytest take care of the clean-up of files?

docs/gallery/general/plot_external_resources.py
```diff
@@ -0,0 +1,182 @@
...
+read_herd.to_dataframe()
+
+###############################################################################
+# View the individual tables:
+
+read_herd.keys.to_dataframe()
+
+###############################################################################
+
+read_herd.entities.to_dataframe()
```
oruebel: I think this could be removed since it is repetitive with the "Inspect the HERD" section. Maybe the "Inspect HERD" section could be moved here to occur after read, which is the more common place where a user would likely need this too.

docs/gallery/general/plot_external_resources.py
```diff
@@ -0,0 +1,182 @@
...
+    name="shank0",
+    description="a shank of the recording probe",
+    location="VISp",
+    device=device,
+)
+for _ in range(4):
+    nwbfile.add_electrode(location="VISp", group=electrode_group)
+
+nwbfile.external_resources.add_ref(
+    container=nwbfile.electrodes,
```
oruebel: How does this look for ragged columns with a `VectorIndex`? Do we in this case annotate the `VectorIndex` or the `VectorData` it points to? I think currently a user would probably try to annotate the `VectorIndex, which I think is fine, but I'm wondering whether that works with HERD (e.g., if it checks the presence of a value and VectorIndex is an int)?

docs/gallery/general/plot_external_resources.py
rly: done

docs/gallery/general/plot_external_resources.py
rly: If a user calls `add_ref(container=table, attribute="col", ...)`, this calls `getattr(table, "col")` which returns `VectorData` for ragged (and non-ragged) columns, so a user trying to annotate a ragged column like the "spike_times" column of the Units table would correctly annotate the `VectorData` not the `VectorIndex`.

If a user tries to annotate the `VectorIndex` directly, like with `add_ref(container=table["col"], attribute=None, ...)`, then the references will annotate the `VectorIndex` and not the `VectorData`, without throwing a warning or error. We could add special handling for `VectorIndex`, but that might be annoying to track and maintain. Instead, I added a note in this section.

`add_ref` does not check the presence of the value. `add_ref_termset` validates a key against a `TermSet`'s permissible values by reading `data_object.data`.

docs/gallery/general/plot_external_resources.py
rly: done

docs/gallery/general/plot_external_resources.py
rly: Good point. Removed

docs/gallery/general/plot_external_resources.py
rly: `get_external_resources` now creates if it doesn't exist already

docs/gallery/general/resources_streaming.py
rly: Fixed in HDMF. Providing the same entity URI will result in a linkage to the previous entity URI, not create a duplicate row. If the same ID is added but points to a new URI, that new URI is ignored with a warning, and the previous entity URI is used. Thanks for adding that into HDMF. https://github.com/hdmf-dev/hdmf/pull/1513

docs/gallery/general/resources_streaming.py
rly: Fixed in HDMF. https://github.com/hdmf-dev/hdmf/pull/1512 . Thanks!

bendichter (approved):

oruebel (approved):

PR #2206, opened 2026-06-23
Add guide on choosing entity_id and entity_uri for HERD references

bendichter:
## Motivation

Follow-up to the HERD tutorial discussion in #2200 and the June 23 CN+LBNL sync. Users adding external resource references with `HERD.add_ref` need clear guidance on what to put in the `entity_id` and `entity_uri` fields, which is currently undocumented and was a recurring point of confusion (e.g. which of the many `NCBITaxon` / `taxonomy` / `NCBI_TAXON` forms to use, and what to do for atlases without per-term URLs).

This adds a narrative reference page (`docs/source/external_resources_entity_guide.rst`, in the **Resources** toctree).

## Guidance

- **`entity_id` should be a CURIE** (`prefix:identifier`) whose prefix is registered with [bioregistry.io](https://bioregistry.io). The Bioregistry maps each CURIE to a canonical, resolvable URL and disambiguates the many overlapping identifier schemes.
- **`entity_uri` should be the URL the CURIE resolves to**, which you can look up via `https://bioregistry.io/<entity_id>`.
- A table of commonly used registries with verified example `entity_id` → `entity_uri` pairs:

  | Prefix | Use for | Example `entity_id` | Example `entity_uri` |
  |---|---|---|---|
  | `NCBITaxon` | Species | `NCBITaxon:10090` | `http://purl.obolibrary.org/obo/NCBITaxon_10090` |
  | `ROR` | Organizations | `ROR:013meh722` | `https://ror.org/013meh722` |
  | `ORCID` | People | `ORCID:0000-0002-1825-0097` | `https://orcid.org/0000-0002-1825-0097` |
  | `UBERON` | Brain regions (cross-species) | `UBERON:0001950` | `http://purl.obolibrary.org/obo/UBERON_0001950` |
  | `MBA` | Allen Mouse Brain Atlas | `MBA:385` | `https://purl.brain-bican.org/ontology/mbao/MBA_385` |
  | `HBA` | Allen Human Brain Atlas | `HBA:4005` | `https://purl.brain-bican.org/ontology/hbao/HBA_4005` |
  | `DANDI` | Dandisets | `DANDI:000015` | `https://dandiarchive.org/dandiset/000015` |

  (Each resolved URL was verified against the Bioregistry API.)

- **Fallback for resources without per-term URLs** (e.g. the macaque **D99** atlas): put the resource's overall URL in `entity_uri` and the term's atlas-specific ID in `entity_id`, so every reference still dereferences to something authoritative.

## Notes

- All example `entity_uri` values are inline literals (not hyperlinks), so the `-W` linkcheck CI does not probe them; only the two stable homepages (`bioregistry.io`, the W3C CURIE spec) are linked. Cross-references to `hdmf.common.resources.HERD` / `add_ref` resolve via the existing hdmf intersphinx mapping.
- I couldn't do a full local docs build to confirm rendering because this checkout's `nwb-schema` submodule is in a modified state (build fails at config with `No specification for 'BaseImage'`), unrelated to this change. RST structure (table columns, title underlines) was validated separately.

### DIFF
--- a/docs/source/external_resources_entity_guide.rst
+++ b/docs/source/external_resources_entity_guide.rst
@@ -67,17 +67,17 @@ the canonical URL the example ``entity_id`` resolves to.
      - ``https://orcid.org/0000-0002-1825-0097``
    * - ``UBERON``
      - Brain regions (cross-species)
-     - ``ElectrodeGroup.location``, ``ImagingPlane.location``, ``electrodes`` ``location`` column
+     - Brain-region location fields [#loc]_
      - ``UBERON:0001950``
      - ``http://purl.obolibrary.org/obo/UBERON_0001950``
    * - ``MBA``
      - Brain regions (Allen Mouse Brain Atlas)
-     - ``ElectrodeGroup.location``, ``ImagingPlane.location``, ``electrodes`` ``location`` column
+     - Brain-region location fields [#loc]_
      - ``MBA:385``
      - ``https://purl.brain-bican.org/ontology/mbao/MBA_385``
    * - ``HBA``
      - Brain regions (Allen Human Brain Atlas)
-     - ``ElectrodeGroup.location``, ``ImagingPlane.location``, ``electrodes`` ``location`` column
+     - Brain-region location fields [#loc]_
      - ``HBA:4005``
      - ``https://purl.brain-bican.org/ontology/hbao/HBA_4005``
    * - ``DANDI``
@@ -85,6 +85,9 @@ the canonical URL the example ``entity_id`` resolves to.
      - (identifies the dataset as a whole)
      - ``DANDI:000015``
      - ``https://dandiarchive.org/dandiset/000015``
+
+.. [#loc] Brain-region annotations commonly apply to ``ElectrodeGroup.location``,
+   ``ImagingPlane.location``, and the ``location`` column of the ``electrodes`` table.
 
 Example
 -------
--- a/docs/source/external_resources_entity_guide.rst
+++ b/docs/source/external_resources_entity_guide.rst
@@ -43,38 +43,46 @@ the canonical URL the example ``entity_id`` resolves to.
 
 .. list-table::
    :header-rows: 1
-   :widths: 12 26 26 36
+   :widths: 10 16 22 20 32
 
    * - Prefix
      - Use for
+     - Common NWB field(s)
      - Example ``entity_id``
      - Example ``entity_uri``
    * - ``NCBITaxon``
      - Species
+     - ``Subject.species``
      - ``NCBITaxon:10090``
      - ``http://purl.obolibrary.org/obo/NCBITaxon_10090``
    * - ``ROR``
      - Organizations / institutions
+     - ``NWBFile.institution``
      - ``ROR:013meh722``
      - ``https://ror.org/013meh722``
    * - ``ORCID``
      - People (researchers)
+     - ``NWBFile.experimenter``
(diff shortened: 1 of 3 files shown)


bendichter:
https://pynwb--2206.org.readthedocs.build/en/2206/external_resources_entity_guide.html

docs/source/external_resources_entity_guide.rst
```diff
@@ -0,0 +1,127 @@
+.. _external_resources_entity_guide:
+
+Choosing ``entity_id`` and ``entity_uri`` for external references
+=================================================================
+
+When you annotate data with an external resource using
+:py:meth:`HERD.add_ref <hdmf.common.resources.HERD.add_ref>`, each reference records two
+fields that identify the external term:
```
oruebel, line 8: Here we should link to the external resources tutorials for PyNWB. Not sure if this can be done in this PR or whether that needs to be done on the other tutorial PRs. Just wanted to flag it.

bendichter: yeah I was going to add cross-linking once the tutorial was merged. Thanks for flagging this.

docs/source/external_resources_entity_guide.rst
```diff
@@ -0,0 +1,127 @@
...
+recognize the registry from the compact ``entity_id`` and dereference the ``entity_uri`` to land
+on an authoritative description of the term.
+
+Commonly used registries
+-------------------------
+
+All of the registries below are registered with the Bioregistry. The ``entity_uri`` column shows
+the canonical URL the example ``entity_id`` resolves to.
+
+.. list-table::
```
oruebel, line 44: @bendichter do you think it would be useful to also provide a column here with "Common NWB Fields" that this apply to, e.g., species applying to `NWBFile.subject.specifies` or for "People" `NWBFile.experimenter` etc.  Just to make it easier for a user to link general concept, e.g,. People, to specific entities in an NWB file. Or maybe that is better served by the tutorials where we can show examples for each of the different fields?

bendichter: oh yeah, good idea!

bendichter: Good idea — added a **Common NWB field(s)** column in 5f202036 mapping each concept to where it appears in an NWB file: `Subject.species`, `NWBFile.institution`, `NWBFile.experimenter`, and `ElectrodeGroup.location`/`ImagingPlane.location`/the `electrodes` `location` column for brain regions (`DANDI` identifies the dataset as a whole, so no single field). I think the table gives the quick concept→field pointer, and the HERD tutorials are still the better place for full worked examples per field.

oruebel:
Thanks @bendichter! This is very helpful! This seems to lay the foundation for a more general guide on external resources. I am wondering whether this may be more appropriate for the https://nwb-overview.readthedocs.io website to provide an entry point from there into the topic of external resources? I think this would probably be it's own top-level page, but I think it would also be useful to add a section to the conversion guide https://nwb-overview.readthedocs.io/en/latest/conversion_tutorial/user_guide.html#  What do you think?

bendichter:
yes, I agree it could go there. That way it can be a reference for MATLAB and Python users.

oruebel:
> yes, I agree it could go there. That way it can be a reference for MATLAB and Python users.

I think this is a great start as is, so I think we can just move it to nwb-overview and then keep adding to it as we go in additional issues/PRs.

bendichter:
closed in favor of nwb-overview

PR #2207, opened 2026-06-25
In test_import_structure, don’t check dunder names

musicinmybrain:
Don’t check for names of the form `__foo__` in the module, except for `__version__` since it’s explicitly assigned.

## Motivation

These are generally subject to change across Python interpreter versions. In particular, `__cached__` is no longer present in Python 3.15; see https://docs.python.org/3.14/reference/datamodel.html#module.__cached__.

Therefore, this PR prevents a test failure in Python 3.15. Dropping `__cached__` from the list would have sufficed for this purpose, but this PR seeks to make the test less unnecessarily brittle in general.

## How to test the behavior?

Just run the tests normally, and observe that they still test. Testing with Python 3.15 in a virtualenv is still a bit tricky since some of the dependencies aren’t ready. We’re able to do it downstream [in Fedora](https://src.fedoraproject.org/rpms/python-pynwb), which is what prompted this PR.

## Checklist

- [ ] Did you update CHANGELOG.md with your changes? **No, and I don’t think they are worth pointing out to users.**
- [ ] Have you included the relevant issue number using "Fix #XXX" notation where XXX is the issue number? By including "Fix #XXX" you allow GitHub to close issue #XXX when the PR is merged. **N/A; I didn’t open an issue.**


rly:
That makes sense. Thanks @musicinmybrain !

rly (approved):

PR #2214, opened 2026-06-30
Fix `mock_electrodes` ignoring `n_electrodes` when sizing the electrodes table

h-mayorquin:
`mock_electrodes` always built a 5-row `ElectrodesTable` while the `DynamicTableRegion` length followed `n_electrodes`, so any value above 5 referenced rows that did not exist. Under HDMF 4.x+, where `DynamicTableRegion.__init__` validates indices at construction, this raises an `IndexError`, which also breaks `mock_ElectricalSeries` for any data with more than 5 channels. This sizes the auto-created table to `n_electrodes` so the region is always in range, leaving the default `n_electrodes=5` behavior unchanged. It was surfaced by https://github.com/catalystneuro/neuroconv/pull/1769, which calls `mock_ElectricalSeries(data=np.ones((10_000, 128)))` and failed at construction.

## Checklist


tests/unit/test_mock.py
```diff
@@ -103,6 +106,20 @@ def test_mock_TimeSeries_w_no_time():
     assert ts.rate == 10.0
 
 
+def test_mock_electrodes_sizes_table_to_n_electrodes():
+    """The auto-created table must have at least n_electrodes rows so the region is in range."""
```
rly:
```suggestion
    """The auto-created table must have exactly n_electrodes rows so the region of n_electrodes is in range."""
```

rly (approved):

rly:
Thanks @h-mayorquin !

PR #2229, opened 2026-07-23
Work around HDF5 ros3 shutdown deadlock on Windows CI

### LINKED ISSUE #2228: Windows ros3 test jobs hang on process exit (since 4.0.0)
The scheduled `Run all tests` workflow has two Windows ros3 jobs that hang and get cancelled at the 6-hour job limit:

- `conda-windows-python3.14-ros3` (`python test.py --ros3`)
- `conda-windows-gallery-python3.14-ros3` (`python test.py --example-ros3`)

The ros3 tests themselves pass. The suite logs its final `Ran N tests - OK` and then the Python process never exits, so the job runs until GitHub cancels it ~6 hours later. The equivalent Linux and macOS ros3 jobs on Python 3.14 pass in ~2 minutes, so this is Windows-specific.

This started with the 4.0.0 release.

Example run: https://github.com/NeurodataWithoutBorders/pynwb/actions/runs/29983760208

rly:
## Summary

Fixes #2228

This is an upstream HDF5 defect, filed at https://github.com/HDFGroup/hdf5/issues/6560. This PR is a CI-side workaround so the jobs go green while this is resolved upstream.

Backing the workaround out once HDF5 ships the fix is tracked in #2231.

## Changes

- `scripts/test_ros3.py`: runs `test.py` as a child process; once the child reports its exit code via `PYNWB_EXITCODE_FILE` but fails to exit, terminates it from the parent (which is not subject to the loader-lock deadlock) and propagates the reported code. A no-op on platforms that exit cleanly.
- `test.py`: writes its computed exit code to `PYNWB_EXITCODE_FILE` when that env var is set, before `sys.exit`.
- `run_all_tests.yml` / `run_tests.yml`: the ros3 jobs run through the wrapper; `run_all_tests.yml` ros3 jobs also get `timeout-minutes: 20` as a backstop.


oruebel (approved):
I would file a follow-up issue to back out these changes once it has been fixed upstream in HDF5, just so we have a reminder.

rly:
Good call. Filed #2231 as the reminder to back this out, with a checklist of exactly what to delete and the condition for acting (an HDF5 release with the fix landing in the conda-forge builds the ros3 jobs install). Also added the upstream link (https://github.com/HDFGroup/hdf5/issues/6560) to the PR description, which had an empty link.

PR #2230, opened 2026-07-24
Read files whose dates carry a sub-minute UTC offset

h-mayorquin:
I ran into this while working with `session1.nwb` from the Karpowicz 2025 cursor-control deposit on Dryad (https://datadryad.org/dataset/doi:10.5061/dryad.q83bk3jtp): `NWBHDF5IO.read()` refused to open it, raising `ParserError: Unknown string format`. The reason is that its `session_start_time` is `1900-10-01T00:00:00-05:50:36`, a sub-minute UTC offset (a placeholder artifact from attaching a `pytz` zone without calling `localize`), and `dateutil` rejects any offset with a seconds component, so a 237 MB file is unreadable over one metadata string. These offsets are outside ISO 8601, so pynwb itself never writes them, but I think we should still be able to read old files that another tool wrote this way. So I made date parsing on read (`session_start_time`, `timestamps_reference_time`, `file_create_date`, and `Subject.date_of_birth`) fall back to parsing the sub-minute offset itself, and raise a clear `ValueError` naming the field when a value is genuinely unparseable. (`datetime.fromisoformat` only accepts these offsets from Python 3.11, so the fallback handles the offset directly to keep working on the 3.10 floor.) What do you think?


src/pynwb/io/file.py
```diff
@@ -10,6 +12,59 @@
...
+        hours=int(match.group("hours")),
+        minutes=int(match.group("minutes")),
+        seconds=int(match.group("seconds")),
+    )
+    if match.group("sign") == "-":
+        offset = -offset
+    return base.replace(tzinfo=datetime.timezone(offset))
+
+
+def _parse_date(datestr, field_name):
```
oruebel: `parse_date` and `parse_subminute_offset_date` look like they are general utility functions that are not specific to `file.py`. As such, I would move these as regular utility functions to [src/pynwb/io/utils.py](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/io/utils.py) instead

oruebel:
> old files that another tool wrote this way

Do you know which tool generated this and whether this is still an issue that needs to be fixed there?

> raise a clear `ValueError` naming the field when a value is genuinely unparseable

Pandas defined the `pd.NaT` as a NaN value for datateime https://pandas.pydata.org/docs/reference/api/pandas.NaT.html Do you think that work as a fallback value when parsing fails? I suspect it probably won't work, because I think `pd.NaT` only behaves like a datetime but is actually not an instance of datetime, but I figured it may be worth a try.

> file is unreadable over one metadata string

The approach here looks fine, because it is a well-defined error case. However, more broadly, fixing read for all possible invalid metadata cases is going to be unmaintainable. Just a crazy idea, but let me see what you think. I suspect that in many cases read of bad metadata ultimately ends of failing in docval not letting None pass. Maybe there could be an `unsafe_read` method that would disable docval validation on read. I.e., when regular read fails, a user could try `unsafe_read` to get to the data by disabling validation. Also, by calling `unsafe_read` explicitly, it would make it clear that all bets are off, and even if the data is being read it is likely invalid and may not work.

t-b:
Would have our validation tools found the error?

src/pynwb/io/file.py
h-mayorquin: It is done.

h-mayorquin:
> Do you know which tool generated this and whether this is still an issue that needs to be fixed there?

I have not been able to pin the writer, the Dryad deposit ships the two NWB files and a README but no conversion code. The mechanism is clear though, and it turns out to be more mundane than I first thought. `-05:50:36` is the unrounded Local Mean Time entry for `America/Chicago` / `US/Central`, and I checked every zone in tzdata, it is the only match. What produces it is the 1900 placeholder date, not any exotic timezone handling: transition tables bottom out at 1901-12-13, the 32-bit `time_t` floor, so any earlier date falls back to the zone's first entry, which is LMT. `dateutil` hands that back unrounded, which is where the seconds come from, while pytz rounds the same entry to `-05:51`. So a plain `datetime(1900, 10, 1)` sentinel plus a real timezone is enough to produce this, no misuse required. The `America/Chicago` part fits, the deposit's authors are partly at Northwestern. So I do not think there is a tool here that needs fixing so much as a conversion script that has already run, which is roughly the case I would expect for old files generally. TimeZone support is probably now better in python so hopefully nothing for us to do other than recommending people to use the native module.

> Pandas defined the `pd.NaT` as a NaN value for datateime https://pandas.pydata.org/docs/reference/api/pandas.NaT.html Do you think that work as a fallback value when parsing fails? I suspect it probably won't work, because I think `pd.NaT` only behaves like a datetime but is actually not an instance of datetime, but I figured it may be worth a try.

I am unsure about this. I can take a look.


> However, more broadly, fixing read for all possible invalid metadata cases is going to be unmaintainable. Just a crazy idea, but let me see what you think. I suspect that in many cases read of bad metadata ultimately ends of failing in docval not letting None pass. Maybe there could be an `unsafe_read` method that would disable docval validation on read. I.e., when regular read fails, a user could try `unsafe_read` to get to the data by disabling validation. Also, by calling `unsafe_read` explicitly, it would make it clear that all bets are off, and even if the data is being read it is likely invalid and may not work.

I agree with this. This would be way easier to mantain and easer the burden on us. In fact, I would be up for deferring this PR to that idea if you think this fix does not accrue enough usability. 


> `parse_date` and `parse_subminute_offset_date` look like they are general utility functions that are not specific to `file.py`. As such, I would move these as regular utility functions to src/pynwb/io/utils.py instead

Agreed, done. Both are now in `src/pynwb/io/utils.py` next to `get_nwb_version`, along with the regex constant they share. I dropped the leading underscores since you said regular utility functions and `get_nwb_version` is unprefixed, happy to put them back if you would rather they stay private.


> Would have our validation tools found the error?

@t-b Yes but `validate` works on builders rather than constructed containers, so it runs happily on a file that `read()` cannot open, and on this one it reports:

```
root/session_start_time (session_start_time): incorrect type - expected 'isodatetime', got 'utf'
```

oruebel:
> I agree with this. This would be way easier to mantain and easer the burden on us. In fact, I would be up for deferring this PR to that idea if you think this fix does not accrue enough usability.

I suspect `unsafe_read` may be a more complex effort, so I think it would be fine to treat it in a separate issue/PR (which could then revert changes here if appropriate). I don't think it is worth holding up this PR for that given that the changes here are fairly self-contained and focused. 

> So I do not think there is a tool here that needs fixing so much as a conversion script that has already run, which is roughly the case I would expect for old files generally.

Thanks for the clarification

> Agreed, done. Both are now in `src/pynwb/io/utils.py`

Thanks, looks good.

oruebel (dismissed):
Looks good to me.

rly:
@h-mayorquin Claude notes and I confirm that Python 3.10 supports sub-minute UTC offsets:
https://docs.python.org/3.10/library/datetime.html#datetime.datetime.fromisoformat

```python
uv run --python 3.10 --no-project python -c "import datetime; print(datetime.datetime.fromisoformat('1900-10-01T00:00:00-05:50:36'))"
```

so I think `parse_subminute_offset_date` doesn't get called. Your change to try `dateutil`, then `fromisoformat` is sufficient.

h-mayorquin:
@rly simplified.

rly:
Thanks @h-mayorquin ! I made some minor adjustments to docstrings, tests, and the changelog entry. Also caught an issue with a bytes array raising an error in `datetime.datetime.fromisoformat`. Looks good to me.

rly (approved):

Issue #727, opened 2018-11-13
NWBData append and extend behavior depends on underlying type


NileGraddis:
Here is a script that replicates this issue (python 3.6, commit 32f3b037d4ce8771e68c0f091731462c65d3ae4c):
```Python
import pynwb
import numpy as np


def main():

    iters = 5
    iter_length = 3

    data = pynwb.core.NWBData(
        name='some_name',
        data=np.array([])
    )
    for ii in range(iters):
        data.append(np.arange(iter_length))
    print(data.data)

    data = pynwb.core.NWBData(
        name='some_name',
        data=[]
    )
    for ii in range(iters):
        data.append(np.arange(iter_length))
    print(data.data)

    data = pynwb.core.NWBData(
        name='some_name',
        data=np.array([])
    )
    for ii in range(iters):
        data.extend(np.arange(iter_length))
    print(data.data)

    data = pynwb.core.NWBData(
        name='some_name',
        data=[]
    )
    for ii in range(iters):
        data.extend(np.arange(iter_length))
    print(data.data)


if __name__ == '__main__':
    main()
```
This produces:
```
[0. 1. 2. 0. 1. 2. 0. 1. 2. 0. 1. 2. 0. 1. 2.]
[array([0, 1, 2]), array([0, 1, 2]), array([0, 1, 2]), array([0, 1, 2]), array([0, 1, 2])]
[0. 1. 2. 0. 1. 2. 0. 1. 2. 0. 1. 2. 0. 1. 2.]
[0, 1, 2, 0, 1, 2, 0, 1, 2, 0, 1, 2, 0, 1, 2]
```

This is odd because the second case here is a conventional representation of a 2D array, while the first is a representation of a 1D array, but the only distinction is whether the NWBData object was constructed with a list or a numpy array as its underlying data store. Similar examples can be constructed for higher-dimensional data (appending a 2D numpy array to a list adds one element, calling np.append flattens both and returns a 1D array).

The core problem here is that instead of ETLing the data into a common internal representation, NWBData maintains these types internally and dispatches further operations by the stored type. This basically creates a case where there are multiple implementations of the same operation with potentially divergent behavior.

[...]

PR #2233, opened 2026-07-29
Fix `NWBData` shadowing the data attribute it inherits from `Data`

h-mayorquin:
## Motivation

I found this while [trying to get neuroconv's default chunking and gzip compression](https://github.com/catalystneuro/neuroconv/pull/1835) to apply to `GrayscaleImage` datasets, and discovering that it silently did nothing. `NWBData` re-declares `self.__data` in its `__init__` and overrides the `data` property to read it, which shadows the storage it inherits from `hdmf.container.Data`. The inherited `set_data_io` writes the parent's attribute while `.data` reads the child's, so the `DataIO` wrapping is applied and then never seen again, and the dataset is written uncompressed with nothing warning you. This affects every `NWBData` subclass, so `ScratchData`, `ImageReferences`, `ExternalImage` and the image types, and the same shadow leaves `transform`, `append`, and `extend` operating on a different attribute than the parent thinks it is holding.

The fix is to let the parent own the storage. The private attribute and the `data` property go, `__len__` was a duplicate of hdmf's so it goes with them, and `append` and `extend` become a type guard that delegates to `super()`, which keeps the existing error message for scalar data.

Two behavior changes worth a reviewer's attention. `Data.__init__` runs `coerce_pandas_data` before storing while `NWBData` stashed the raw argument, so `.data` now returns the coerced value, meaning a `pandas.Series` passed as `data` reads back as the `ndarray` that was always the thing actually written to the file. And because `append` and `extend` now go through hdmf's `append_data` and `extend_data`, appending an array to ndarray-backed data concatenates instead of raising `ValueError: all the input arrays must have same number of dimensions`. That is the reproduction in #727, which this improves but does not close, since the list and ndarray cases still diverge. The existing test suite passes without modification.

## How to test the behavior?

```python
import numpy as np
from hdmf.backends.hdf5 import H5DataIO
from pynwb.image import GrayscaleImage

image = GrayscaleImage(name="image", data=np.zeros((4, 4)))
image.set_data_io(H5DataIO, dict(compression="gzip"))

# before: <class 'numpy.ndarray'>, the H5DataIO is stranded on the parent attribute
# after:  <class 'hdmf.backends.hdf5.h5_utils.H5DataIO'>
print(type(image.data))
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # PyNWB Changelog
 
+## PyNWB 4.1.1 (Unreleased)
+
+### Fixed
+- Fixed `set_data_io` being silently ignored on `NWBData` subclasses (`GrayscaleImage`, `RGBImage`, `RGBAImage`, `ExternalImage`, `ImageReferences`, and `ScratchData`), so requested chunking and compression were dropped without warning and the datasets were written uncompressed. `NWBData` declared its own data storage and `data` property, shadowing the ones inherited from `hdmf.container.Data`, so the inherited `set_data_io` wrapped the parent's attribute while `.data` kept reading the child's. `NWBData` now uses the inherited storage, which also makes `.data` reflect `transform`, `append`, and `extend`, and return the value coerced by `Data.__init__` (a `pandas.Series` passed as `data` now reads back as the `ndarray` that is actually written). @h-mayorquin [#2233](https://github.com/NeurodataWithoutBorders/pynwb/pull/2233)
+
 ## PyNWB 4.1.0 (July 23, 2026)
 
 ### Changed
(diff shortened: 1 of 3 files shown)


rly:
Good catch @h-mayorquin ! I think this is ultimately due to an incomplete refactor when HDMF was extracted out of PyNWB and `NWBData` began inheriting from hdmf's `Data`. Almost all contents of `NWBData` can be removed, except `__init__` to keep the `AllowPositional.WARNING`. Could you remove them all and re-verify?

Could you also please simplify the changelog entry? The first sentence is sufficient.

h-mayorquin:
Yes.

I modified the changelog as requested. It was the product of a verbose model in the top of a verbose person. 

Thanks for the review.

rly:
Thanks @h-mayorquin . What do you think about removing all contents of `NWBData`, except `__init__` to keep the `AllowPositional.WARNING`? I'm happy to do it - let me know.

h-mayorquin:
Should be done.

rly (approved):
Thank you @h-mayorquin !

PR #2237, opened 2026-08-01
Make Units.waveform_unit configurable

### LINKED ISSUE #2162: `Units.waveform_unit` parameter has no effect on serialized `waveform_mean` unit attribute
The `Units` class accepts a `waveform_unit` parameter ([pynwb misc.py L174-175](https://github.com/NeurodataWithoutBorders/pynwb/blob/4b7f5516f0e517a5450e7af0857e2d72f2bb5045/src/pynwb/misc.py#L174-L175)) and pynwb stores it correctly on the container, but the value is never propagated to the HDF5 file. The NWB schema defines the `unit` attribute on `waveform_mean` with a fixed value (`value: volts`) in [nwb-schema nwb.misc.yaml L259-262](https://github.com/NeurodataWithoutBorders/nwb-schema/blob/ade50ef33446beb3c7df4c6f1072ae0e821b5115/core/nwb.misc.yaml#L259-L262), and hdmf enforces this by short-circuiting attribute resolution when `spec.value` is set in [hdmf objectmapper.py L1061-1062](https://github.com/hdmf-dev/hdmf/blob/53570978c6ed0acba66ebb2abab0cbed21f36d2d/src/hdmf/build/objectmapper.py#L1061-L1062). This means the ObjectMapper code in pynwb that maps `waveform_unit` to the `unit` attribute is never reached, making the `waveform_unit` parameter effectively dead code.

My personal view is that the `unit` attribute should either be made configurable by changing `value: volts` to `default: volts` in the schema, or the fixed value should be changed to microvolts which is the natural unit for extracellular electrophysiology recordings.

rly:
Fixes #2162.

Updates the nwb-schema submodule to the latest schema dev (2.10.1-alpha), which includes NeurodataWithoutBorders/nwb-schema#707 and NeurodataWithoutBorders/nwb-schema#709.

### Configurable waveform unit (nwb-schema#707)

That PR changes the `unit` attribute of `Units.waveform_mean`, `waveform_sd`, and `waveforms` from a fixed value of `"volts"` to a default value of `"volts"`.

HDMF short-circuits attribute resolution when a spec has a fixed value, so the existing mapping from `Units.waveform_unit` onto the `unit` attribute was never reached. With the schema change it is reachable, so no write-path change was needed.

Also implements the warning placeholder in `UnitsMap._get_waveform_stat`, since a file can now carry different `unit` values across the three waveform columns while the `Units` container holds only one.

### Waveform peak alignment (nwb-schema#709)

That PR adds an optional `time_before_peak_in_ms` attribute to `Units.waveform_mean`, `waveform_sd`, and `waveforms`. It holds the time, in milliseconds, from the start of each waveform to the spike peak, that is, the alignment point used during spike sorting. The value is the same for every unit and every spike.

PyNWB exposes it as the `waveform_time_before_peak_in_ms` constructor argument and field of `Units`, following the existing `waveform_rate` and `waveform_unit` pattern: one value on the container, written onto all three waveform columns, and read back from the first populated column with a warning when the columns disagree.

### `NO_OVERRIDE` in `UnitsMap`

The constructor argument overrides in `UnitsMap` returned `None` to signal "no override". HDMF 6.2.0 deprecates that in favor of the `ObjectMapper.NO_OVERRIDE` sentinel, and under HDMF 8.0 a `None` return will set the constructor argument to `None`, which would clobber the `"volts"` default of `waveform_unit`. They now return the sentinel via the existing `pynwb.io.utils.NO_OVERRIDE` shim.

### Note

`waveform_unit`, `waveform_rate`, and `waveform_time_before_peak_in_ms` can only be set on the `Units` constructor, so they are not reachable from the `nwbfile.add_unit(...)` path that auto-creates the table. Worth addressing separately.


h-mayorquin (dismissed):
Thanks for this, looks good to me.

h-mayorquin:
If we are improving the metadata provenance for waveforms it would be great to get this one as well (c.c. @alejoe91)

https://github.com/NeurodataWithoutBorders/nwb-schema/pull/709

h-mayorquin (approved):
Thanks for moving quickly with this. LGTM.

Issue #2215, opened 2026-07-01
Clarify intended behavior of ImageSeries.num_samples for internal (non-external) data


ehennestad:
## Clarification request: `ImageSeries.num_samples` is persisted for internal (non-external) data

I'd like to understand the intended behavior of `num_samples` on `ImageSeries`, because the inherited in-memory `TimeSeries.num_samples` property interacts with the schema-defined `num_samples` dataset (added in #2194) in a way that seems inconsistent with the field's documented meaning. I'm raising this as a question rather than a bug report — I may be misreading the intent.

### Background

- `TimeSeries.num_samples` is a computed property that returns `len(data)` (falling back to `len(timestamps)`).
- `ImageSeries` overrides this property, and #2194 also added `num_samples` as an actual schema dataset (`nwb.image.yaml`), documented as:

  > Total number of frames across all external files. This is required when `format='external'` and timing is described using `starting_time` and `rate`, since data is empty and its first dimension cannot be used to determine the number of frames. When `timestamps` is provided, `len(timestamps)` already serves this purpose.

The doc frames `num_samples` as an **external-file** concept: a way to recover a frame count that cannot otherwise be derived. The `_check_num_samples` warning is likewise gated on `external_file is not None and rate is not None and num_samples is None`.

### Observed behavior

For an **internal** `ImageSeries` (data stored in the file, no external file), the user never sets `num_samples`, but:

1. The overridden property falls through to `TimeSeries.num_samples` and returns `len(data)`.
2. There is no custom IO mapper, so the default `ObjectMapper` persists whatever the property returns — so `num_samples` gets written to disk even though `data.shape[0]` already carries that information.

```python
import numpy as np
from pynwb.image import ImageSeries

img = ImageSeries(name="img", data=np.zeros((10, 5, 5)), unit="n.a.", rate=1.0)
print(img._num_samples)  # None  -- user never set it
print(img.num_samples)   # 10    -- inherited TimeSeries property returns len(data)
# on write -> acquisition/img/num_samples == 10 is persisted to the file
```

So the persisted `num_samples` value on an internal `ImageSeries` doesn't correspond to the documented meaning of the field ("total number of frames across all external files") — there are no external files. The value is numerically correct, but it's a redundant write of a field the schema doc describes as external-only.

### Questions

1. Is persisting `num_samples` for internal (non-external) `ImageSeries` intended, or should the dataset only be written in the external-file case it was designed for?
[...]

PR #2239, opened 2026-08-05
Only write ImageSeries.num_samples when it was explicitly provided

### LINKED ISSUE #2227: ImageSeries writes auto-derived num_samples, causing a DtypeConversionWarning
### What happened?

Writing an `ImageSeries`/`OpticalSeries` that was constructed with `data` and `timestamps` (but no explicit `num_samples`) emits an HDMF `DtypeConversionWarning` on write:

```
Spec 'OpticalSeries/num_samples': Value with data type int64 is being converted to data type uint64 (min specification: uint32).
```

This fires 14 times across the integration suite (e.g. `tests/integration/hdf5/test_image.py`).

### Root cause

`ImageSeries.num_samples` auto-derives `len(data)` as a Python `int` (int64) when `num_samples` was not explicitly set. pynwb then writes this derived value against the schema's `uint32` `num_samples` dataset (`core/nwb.image.yaml`), so HDMF widens int64 → uint64 and warns.

The schema doc for `num_samples` states it is only required when `format='external'` and timing uses `starting_time`/`rate`, and that "when timestamps is provided, `len(timestamps)` already serves this." So pynwb likely should not be writing an auto-derived `num_samples` at all when the user did not set one.

### Proposed fix

Only write `num_samples` when it was explicitly provided (i.e. `self._num_samples is not None`), rather than writing the property's auto-derived fallback. Add round-trip tests confirming: (a) no dtype-conversion warning for the derived case, (b) an explicitly set `num_samples` still round-trips as `uint32`.

### Notes

Deferred out of the 4.1.0 release prep (#2226) because it is a write-behavior change that warrants its own review and round-trip tests.

adityasingh2400:
## Motivation

Fix #2227.

`ImageSeries.num_samples` is a property that falls back to the inherited `TimeSeries.num_samples`, i.e. `len(data)`, when the user did not supply a value. There is no custom mapping for that dataset, so the default `ObjectMapper` read the property and persisted the derived value. Every `ImageSeries` (and `OpticalSeries`, `OnePhotonSeries`, `TwoPhotonSeries`) built from internal data therefore wrote a `num_samples` dataset the user never set, and because the derived value is a Python `int` and the schema dataset is `uint32`, HDMF widened it to `uint64` and emitted `DtypeConversionWarning: Spec 'ImageSeries/num_samples': Value with data type int64 is being converted to data type uint64 (min specification: uint32)`. That warning fires across the image and ophys integration tests today.

The schema documents `num_samples` as the frame count that cannot otherwise be recovered, which is the `format='external'` plus `starting_time`/`rate` case, so writing a value derived from `len(data)` is redundant. This matches the conclusion in #2215 that persisting it for internal data was an oversight.

## Fix

`ImageSeriesMap` now maps the write path of the `num_samples` spec to the private `_num_samples`, which holds a value only when `num_samples` was explicitly passed to the constructor. When it was not set, nothing is written. The read path is untouched, so a file that does contain `num_samples` still maps it back to the `num_samples` constructor argument and round-trips unchanged. The in-memory `ImageSeries.num_samples` property is also unchanged and still returns the frame count derived from the data.

An `object_attr` override additionally casts an explicitly provided, in-range value to `uint32` so it is written with the schema dtype rather than being widened to `uint64`. Out-of-range values are passed through untouched so HDMF still reports the mismatch.

## How to test the behavior?

```python
import numpy as np, warnings, h5py
from datetime import datetime
from dateutil.tz import tzlocal
from pynwb.image import ImageSeries

nwbfile = NWBFile("s", "i", datetime.now(tzlocal()))
nwbfile.add_acquisition(ImageSeries(name="img", data=np.zeros((10, 5, 5), dtype=np.uint8), unit="n.a.", rate=1.0))
with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
    with NWBHDF5IO("test.nwb", "w") as io:
        io.write(nwbfile)
print([str(x.message) for x in w])   # before: one DtypeConversionWarning, after: []
with h5py.File("test.nwb") as f:
    print(list(f["acquisition/img"].keys()))   # before: [..., 'num_samples', ...], after: no num_samples
```

Four new tests in `tests/integration/hdf5/test_image.py` cover this. They fail on `dev` and pass with the change:

```
FAILED tests/integration/hdf5/test_image.py::TestImageSeriesWithNumSamplesIO::test_num_samples_written_as_uint32
FAILED tests/integration/hdf5/test_image.py::TestImageSeriesDerivedNumSamplesIO::test_derived_num_samples_not_written
FAILED tests/integration/hdf5/test_image.py::TestImageSeriesNumSamplesWriteWarnings::test_no_warning_for_derived_num_samples
FAILED tests/integration/hdf5/test_image.py::TestImageSeriesNumSamplesWriteWarnings::test_no_warning_for_explicit_num_samples
4 failed, 4 passed, 9 deselected
```

After the change, `pytest tests/integration/hdf5/test_image.py tests/integration/hdf5/test_ophys.py` passes with zero `num_samples` dtype-conversion warnings, down from the previous count.

I deliberately left out the consistency check you suggested in #2215, where an explicitly provided `num_samples` on an internal `ImageSeries` should be validated against the number of frames in the data. Happy to add it here or in a follow-up if you would like it.

## Checklist


rly:
Thanks @adityasingh2400 . I made a couple minor changes. This is good to go.

rly (approved):

PR #2243, opened 2026-08-16
Fix `NWBFile.objects` going stale after the file is modified

### LINKED ISSUE #2242: [Bug]: NWBFile.objects is cached on first access and never invalidated
Adding new objects to an NWB file causes `nwb_file.objects` to become stale.

<details>
<summary>Details (AI-generated)</summary>

### What happened?

`NWBFile.objects` is populated once, lazily, and nothing ever invalidates it
([`src/pynwb/file.py:587-591`](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/file.py#L587-L591)):

```python
@property
def objects(self):
    if self.__obj is None:
        self.all_children()
    return self.__obj
```

`all_children()` rebuilds `self.__obj` and returns a fresh walk; `objects` returns whatever the
first walk happened to see. So the first read of `.objects` decides its contents for the life of the
`NWBFile` — containers added afterwards are absent from it, and containers removed afterwards are
still in it.

I expected `.objects` to reflect the file as it currently stands, the way `all_children()` does. At
minimum I did not expect merely *reading* the property to change what a later read returns.

### Where I hit it

I'm adding NWB export to GuPPy ([LernerLab/GuPPy#357](https://github.com/LernerLab/GuPPy/pull/357)),
which reads an existing NWB file, adds the fiber-photometry derived data to it via a NeuroConv
interface, and writes the result:

```python
nwbfile = read_nwb(path=source_nwbfile_path)
interface.add_to_nwbfile(nwbfile=nwbfile, metadata=interface.get_metadata())
configure_and_write_nwbfile(nwbfile=nwbfile, nwbfile_path=nwbfile_path, backend="hdf5")
```

The output was valid and the data was correct, but every dataset the interface added — the derived
traces, the transients tables, the events table — came out with `compression=None` and
`chunks=None`, while the datasets that came from the source file were `gzip`.

The cause is the cache. Early in `add_to_nwbfile`, the interface resolves GuPPy's store ids against
the file's existing `FiberPhotometryResponseSeries` by reading `nwbfile.objects` — a read-only
inspection, with no intent to mutate anything. That froze the cache at 42 objects. The finished file
had 85. NeuroConv's backend configuration then walks `.objects` to decide which datasets to wrap in
an `H5DataIO`, saw the 42-object snapshot, and wrote the other 43 containers with backend defaults.

Nothing raised, so this is only visible by inspecting the written file's filters. On a real
multi-hour session those uncompressed traces are among the largest objects in the file. The
NeuroConv-side report is [catalystneuro/neuroconv#1909](https://github.com/catalystneuro/neuroconv/issues/1909);
we'll switch our call sites to `all_children()`, but the trap is easy to fall into from anywhere:
whether an inspection of an `NWBFile` corrupts a later inspection depends on which of the two
accessors each unrelated piece of code happened to pick.

### Steps to Reproduce

```python
"""`NWBFile.objects` is computed once and never invalidated."""

import numpy as np

nwbfile = mock_NWBFile()

nwbfile.objects  # anything that inspects the file freezes the cache here

nwbfile.add_acquisition(mock_TimeSeries(name="AddedAfter", data=np.arange(10)))
print("in objects:      ", any(o.name == "AddedAfter" for o in nwbfile.objects.values()))
print("in all_children: ", any(c.name == "AddedAfter" for c in nwbfile.all_children()))
print("in objects after all_children:", any(o.name == "AddedAfter" for o in nwbfile.objects.values()))

# Nested containers behave the same way.
module = nwbfile.create_processing_module(name="behavior", description="")
nwbfile.objects
module.add(mock_TimeSeries(name="Nested", data=np.arange(10)))
print("nested in objects:", any(o.name == "Nested" for o in nwbfile.objects.values()))

# And removals never leave the cache.
timeseries = mock_TimeSeries(name="Removed", data=np.arange(10))
nwbfile.add_acquisition(timeseries)
nwbfile.all_children()
del nwbfile.acquisition["Removed"]
timeseries.reset_parent()
print("removed still in objects:", any(o.name == "Removed" for o in nwbfile.objects.values()))
```

```
in objects:       False
in all_children:  True
in objects after all_children: True
nested in objects: False
removed still in objects: True
```

Note the third line: `all_children()` repopulates `__obj` as a side effect, so an unrelated call
elsewhere can also silently *un*-stale the cache. Whether `.objects` is correct depends on the
interleaving of reads and writes across otherwise independent code.

### Environment

- pynwb 4.0.0 (`objects` is unchanged on `dev` as of this writing), hdmf 6.1.0
- Python 3.13.13, conda, macOS

</details>

HugoFara:
## Motivation

`NWBFile.objects` populates its dict once, lazily, and nothing invalidates it, so the first read decides its contents for the life of the file: containers added afterwards are missing from it, containers removed are still in it. `all_children()` repopulates the same cache as a side effect, so whether `objects` is correct depends on how reads and writes in unrelated code interleave.

Nothing raises when it goes wrong. In #2242 a read-only inspection froze the dict at 42 objects in a file that ended up with 85; NeuroConv's backend configuration then walked `.objects` and wrote the other 43 containers with backend defaults, i.e. uncompressed — visible only by reopening the file and checking its filters.

`objects` is now rebuilt on every access, and `all_children()` no longer writes to cached state. I considered keeping the cache and invalidating it on mutation, but HDMF exposes no hook for "a descendant was added or removed", so invalidation could only be best-effort. Rebuilding costs one walk per access, so reading `.objects` inside a loop over containers is now O(n²) where it was O(n); happy to trade the other way if you prefer.

## How to test the behavior?

The snippet in #2242 prints `False / True / True / False / True` on `dev` and `True / True / True / True / False` with this change. End to end through NeuroConv 0.9.4, the three `TimeSeries` added after the `.objects` read are written `compression=None, chunks=None` on `dev` and `gzip, chunks=(1000,)` here.

Three regression tests in `tests/unit/test_file.py` — `test_objects_includes_container_added_after_first_access`, `test_objects_includes_nested_container_added_after_first_access`, `test_objects_excludes_removed_container` — fail on `dev` and pass with the change. `pytest tests/unit`: 575 passed, 3 skipped. `pytest tests/integration/hdf5`: 223 passed, 3 skipped. `ruff check .` and `codespell` clean. Python 3.13, hdmf 6.1.0, h5py 3.16.0.

Fix #2242

## Checklist

---
*Pull request created by Claude Opus 5 using the principles of [knock-first](https://github.com/HugoFara/knock-first).*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Fixed
 - Fixed `Units.waveform_unit` having no effect on the written file. The `waveform_unit` passed to `Units` is now written to the `waveform_mean`, `waveform_sd`, and `waveforms` columns, and `"volts"` remains the default. PyNWB now also warns when the waveform columns of a file being read carry different `unit` or `sampling_rate` attributes, since only one value per attribute is kept on the `Units` container. @rly [#2162](https://github.com/NeurodataWithoutBorders/pynwb/issues/2162)
+- Fixed `NWBFile.objects` being cached on first access and never invalidated, so containers added to (or removed from) the file after the property was first read were missing from (or still in) it. It is now rebuilt on every access and always reflects the current contents of the file. @HugoFara [#2242](https://github.com/NeurodataWithoutBorders/pynwb/issues/2242)
 - Fixed `mock_DeviceModel` defaulting `manufacturer` to `None`. The mock now defaults it to `"manufacturer"`. @HugoFara [#2232](https://github.com/NeurodataWithoutBorders/pynwb/pull/2232)
 - Fixed reading a file whose dates carry a sub-minute UTC offset (e.g. `1900-10-01T00:00:00-05:50:36`). @h-mayorquin [#2230](https://github.com/NeurodataWithoutBorders/pynwb/pull/2230)
 - Fixed wide pandas DataFrames in the tutorials spilling out of the content column and into the right margin. @bendichter [#2236](https://github.com/NeurodataWithoutBorders/pynwb/pull/2236)
--- a/src/pynwb/file.py
+++ b/src/pynwb/file.py
@@ -566,18 +566,14 @@ class NWBFile(MultiContainerInterface, HERDManager):
         for key, val in args_to_set.items():
             setattr(self, key, val)
 
-        self.__obj = None
-
     def all_children(self):
+        """Get a list of this file and all of its children, recursively."""
         stack = [self]
         ret = list()
-        self.__obj = LabelledDict(label='all_objects', key_attr='object_id')
         while len(stack):
             n = stack.pop()
             ret.append(n)
-            if n.object_id is not None:
-                self.__obj[n.object_id] = n
-            else:
+            if n.object_id is None:
                 warn('%s "%s" does not have an object_id' % (n.neurodata_type, n.name))
             if hasattr(n, 'children'):
                 for c in n.children:
@@ -586,9 +582,15 @@ class NWBFile(MultiContainerInterface, HERDManager):
 
     @property
     def objects(self):
-        if self.__obj is None:
-            self.all_children()
-        return self.__obj
+        """A dict of this file and all of its children, recursively, keyed by object ID.
+
+        The dict is rebuilt on every access, so it always reflects the current contents of the file.
+        """
+        ret = LabelledDict(label='all_objects', key_attr='object_id')
+        for n in self.all_children():
+            if n.object_id is not None:
+                ret[n.object_id] = n
+        return ret
 
     @property
     def epoch_tags(self):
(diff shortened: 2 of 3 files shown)


HugoFara:
CI is not passing, I can create a second PR to fix it but I'd rather ask maintainers before stacking up work.

Two solutions from what I see:

1. Change the nwbinspector test to `test_sweep_table.parent = nwbfile` (and `location=None`, since a real child has no in-memory path). I verified that passes both with and without this PR, so it can merge on its own schedule and this gate goes green with no further commit here. Happy to open it.
2. Keep write-through in pynwb: hold one persistent dict, refresh real children in place, leave externally-inserted keys alone. Roughly ten lines, but it makes writing into a derived index a supported thing, and injected entries would then never expire.

1 is cleaner, but 2 is easy if you would rather not break that usage. The codecov patch percentage is the two branch partials in the new loop; I can add a case for a child without an `object_id` if you want it green.

PR #2245, opened 2026-08-25
Fix `TimeSeriesReference.timestamps` for series that store `starting_time` and `rate`

h-mayorquin:
## Motivation

I was reading `TimeSeriesReference` and noticed that its `timestamps` property multiplies the sample index by `rate` where it should divide by it, so every timestamp reconstructed from `starting_time` and `rate` comes back off by a factor of `rate` squared. 

`TimeSeries.get_timestamps` does the right thing:
https://github.com/NeurodataWithoutBorders/pynwb/blob/b3d8d293/src/pynwb/base.py#L373
but they don't match.

I also changed a test that was freezing the old incorrect behavior.

## How to test the behavior?
```python
import numpy as np
from pynwb.base import TimeSeries, TimeSeriesReference

ts = TimeSeries(name="ts", description="d", data=np.arange(1000), unit="V", starting_time=0.0, rate=100.0)
tsr = TimeSeriesReference(100, 4, ts)

print(tsr.timestamps)
print(ts.get_timestamps()[100:104])
```
On `dev` the two lines disagree:
```
[10000. 10100. 10200. 10300.]
[1.   1.01 1.02 1.03]
```
With this PR both print `[1.   1.01 1.02 1.03]`.

## Checklist

### DIFF
--- a/src/pynwb/base.py
+++ b/src/pynwb/base.py
@@ -676,8 +676,8 @@ class TimeSeriesReference(NamedTuple):
             return self.timeseries.timestamps[self.idx_start: (self.idx_start + self.count)]
         # construct the timestamps from the starting_time and rate
         else:
-            start_time = self.timeseries.rate * self.idx_start + self.timeseries.starting_time
-            return np.arange(0, self.count) * self.timeseries.rate + start_time
+            start_time = self.idx_start / self.timeseries.rate + self.timeseries.starting_time
+            return np.arange(0, self.count) / self.timeseries.rate + start_time
 
     @property
     def data(self):
--- a/src/pynwb/base.py
+++ b/src/pynwb/base.py
@@ -676,8 +676,8 @@ class TimeSeriesReference(NamedTuple):
             return self.timeseries.timestamps[self.idx_start: (self.idx_start + self.count)]
         # construct the timestamps from the starting_time and rate
         else:
-            start_time = self.idx_start / self.timeseries.rate + self.timeseries.starting_time
-            return np.arange(0, self.count) / self.timeseries.rate + start_time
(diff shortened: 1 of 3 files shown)


src/pynwb/base.py
```diff
@@ -676,8 +676,8 @@ def timestamps(self):
             return self.timeseries.timestamps[self.idx_start: (self.idx_start + self.count)]
         # construct the timestamps from the starting_time and rate
         else:
-            start_time = self.timeseries.rate * self.idx_start + self.timeseries.starting_time
-            return np.arange(0, self.count) * self.timeseries.rate + start_time
+            start_time = self.idx_start / self.timeseries.rate + self.timeseries.starting_time
+            return np.arange(0, self.count) / self.timeseries.rate + start_time
```
rly: This is ever so slightly different from `TimeSeries.get_timestamps()` such that this and that would differ in float rounding. `return (self.idx_start + np.arange(self.count)) / self.timeseries.rate + self.timeseries.starting_time` would match. I'll make a quick change.

h-mayorquin: Thanks!

tests/unit/test_base.py
```diff
@@ -867,7 +867,7 @@ def _create_time_series_with_rate(self):
             data=np.arange(10),
             unit="unit",
             starting_time=5.0,
-            rate=0.1,
+            rate=100.0,
```
rly, line 870: I also made this `rate` a little more realistic so that weird test outputs are more realistic and we're more likely to spot odd behavior.

rly:
Great catch on a long-standing bug. Thanks @h-mayorquin !

rly (approved):

PR #2251, opened 2026-08-27
Illustrate usage of add_ref_container in HERD tutorial

oruebel:
## Motivation

The PyNWB tutorials currently don't show usage of `add_ref_container` for populating `HERD` when a term set configuration is used. 

## Checklist

### DIFF
--- a/docs/gallery/general/plot_configurator.py
+++ b/docs/gallery/general/plot_configurator.py
@@ -112,6 +112,47 @@ nwbfile.subject = subject
 # current configuration.
 config = get_loaded_type_config()
 
+####################################
+# How to populate HERD from TermSet-wrapped fields
+# -------------------------------------------------
+# Since ``experimenter`` and ``species`` are validated against a
+# :py:class:`~hdmf.term_set.TermSet`, PyNWB can automatically populate a
+# :py:class:`~pynwb.resources.HERD` (HDMF External Resources Data Structure) with references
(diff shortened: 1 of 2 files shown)


docs/gallery/general/plot_configurator.py
```diff
@@ -112,6 +112,47 @@
...
+# the whole file, use :py:meth:`~hdmf.common.resources.HERD.add_ref_termset` directly on that
+# field, e.g.:
+#
+# .. code-block:: python
+#
+#     herd.add_ref_termset(
+#         container=subject,
+#         attribute="species",
+#         termset=subject.fields["species"].termset,
+#         key=subject.fields["species"].value,
```
rly, line 153: These are odd to me. We shouldn't expose `fields` unless we need to, and the termset and key should be derivable from container + attribute. Would be nice if something like this:
```
herd.add_ref_termset(container=subject, attribute="species")
```
would work by default.

These docs are fine, but let's update them after we clean some of this up in HDMF.

rly (dismissed):

docs/gallery/general/plot_configurator.py
rly: See https://github.com/hdmf-dev/hdmf/issues/1568 and https://github.com/hdmf-dev/hdmf/issues/1569

rly (approved):

PR #2253, opened 2026-08-29
Store FeatureExtraction.times and Clustering.peak_over_rms as given, instead of copying them into a list

cboulay:
## Motivation

`FeatureExtraction.__init__` and `Clustering.__init__` copy one of their fields into a Python list:

```python
self.times = list(times)                                        # ecephys.py:463
args_to_set['peak_over_rms'] = list(args_to_set['peak_over_rms'])  # ecephys.py:317
```

On read, that field is the backend's lazily-read dataset, and `list()` walks it one element at a time. On a real 30 kHz recording whose `FeatureExtraction` holds 23,630 feature times, that is 23,630 separate `h5py.Dataset.__getitem__` calls during `io.read()`:

| | |
|---|---|
| `list(times)` — element-wise | 24.2 ms |
| `np.asarray(times)` — one read | 0.29 ms |

I found this while profiling `NWBHDF5IO.read()` on a 5.7 GB file: those 23,630 calls were 23,630 of the 23,677 h5py reads the whole open performed, and roughly a fifth of its wall time. The cost scales with the number of feature times, so longer sessions pay proportionally more.

It also works against [#2235](https://github.com/NeurodataWithoutBorders/pynwb/pull/2235), which retyped these very fields from `Iterable` to `('array_data', 'data')` so they would accept a lazy zarr array, and replaced `len()` calls with `get_data_shape` to avoid touching the data. A `list()` two lines later materializes what that work took care not to. Element-wise access is worse on zarr than on h5py, since each index can pull a chunk.

## Changes

Both fields are now stored as given.

This matches the fields beside them: `FeatureExtraction` already keeps `features` and `description` as passed, and `Clustering` keeps `num` and `times`. `peak_over_rms` and `times` were the only two array-valued fields in `src/pynwb` copied this way.

## Behavior changes

`FeatureExtraction.times` and `Clustering.peak_over_rms` now return whatever was passed in, rather than always a `list` — an `h5py.Dataset` or `zarr.Array` when read from a file, an `ndarray` when constructed with one, and still a `list` when constructed with one. Both remain sequences supporting `len()`, indexing, iteration, and `np.asarray()`.

The memory saving is secondary and small: a list of Python floats is about five times an equivalent ndarray, 0.95 MB against 0.19 MB for the recording above, and it is confined to that one vector.

`Clustering` is deprecated, but its constructor still runs when reading legacy files, so it gets the same fix.

## How tested

Full unit and integration suite passes (854 passed). The 14 `tests/validation/test_validate.py::TestValidateCLI` failures in my environment are pre-existing — they reproduce on a clean `dev` checkout.

The existing assertions pass unchanged, because they compare against the plain lists the tests pass in and an uncopied list compares equal to itself. The two new tests pass an `ndarray` and assert identity, which is what pins the absence of a copy rather than merely the equality of one.

## Checklist

- [x] Did you add/update tests?


rly (approved):
LGTM! I have no idea why there was a `list(...)` there to begin with. Strange.

rly:
Thanks @cboulay !

PR #2255, opened 2026-08-29
Fix NaN handling for TimeIntervals bounds

### LINKED ISSUE #2212: TimeIntervals.get_duration / get_starting_time return NaN with NaN start/stop times
### What happened?

`TimeIntervals.get_duration` (added in #2146) uses `np.max` on `stop_time` data. Stop times are allowed to be NaN (e.g. for ongoing/unbounded intervals), and `np.max` propagates NaN, so `get_duration` returns `nan` whenever any stop time is NaN. The same issue applies to `get_starting_time`, which uses `np.min` on `start_time`.

### Expected behavior

- `get_starting_time` should ignore NaNs (use `np.nanmin`), and return `None` if the table is empty or all start times are NaN.
- `get_duration` should ignore NaNs (use `np.nanmin`/`np.nanmax`), with these edge cases:
  - All start times **and** all stop times are NaN → return `NaN`.
  - All stop times are NaN but valid start times exist → return the span of the start times (latest start − earliest start).
  - Empty table → return `None`.

### Steps to reproduce

```python
import numpy as np
from pynwb.epoch import TimeIntervals

ti = TimeIntervals(name="intervals")
ti.add_interval(start_time=0.0, stop_time=1.0)
ti.add_interval(start_time=2.0, stop_time=np.nan)  # ongoing interval

print(ti.get_duration())  # -> nan, expected 2.0 (latest valid stop is 1.0; here all-NaN-stop fallback not triggered)
```

### Operating system

N/A

### Python version

N/A

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/.github/CODE_OF_CONDUCT.md)
- [x] Have you ensured this change was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues)?

AtomicGlance:
Fixes #2212

## What changed

- Ignore NaN start times in TimeIntervals.get_starting_time.
- Calculate duration from the earliest valid start to the latest known start or stop boundary, so an ongoing interval with a missing stop does not shorten the observed span.
- Return None for an empty table and NaN when no valid start time exists.
- Add regression coverage for mixed and all-NaN inputs and document the behavior in the upcoming changelog.

## Testing

- `python -m pytest tests/unit/test_epoch.py -q` — 26 passed
- `ruff check src/pynwb/epoch.py tests/unit/test_epoch.py` — passed
- The broader unit suite reports 594 passed and 2 skipped; its only failures are the existing typemap-cache tests, which require the missing NWB schema submodule in this partial checkout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@
 - Added `model` and `serial_number` parameters to `mock_Device`. Passing a `DeviceModel` as `model` together with an `nwbfile` also places that `DeviceModel` in the `NWBFile`, so the link resolves when the file is written. @rly [#2238](https://github.com/NeurodataWithoutBorders/pynwb/pull/2238)
 
 ### Fixed
-- Fixed `FeatureExtraction.times` and `Clustering.peak_over_rms` being copied into a Python list on construction, which was slow for data read from a file. @cboulay [#2253](https://github.com/NeurodataWithoutBorders/pynwb/pull/2253)
+- Fixed `TimeIntervals.get_starting_time` and `get_duration` returning NaN whenever a time column contained NaN values. NaN values are now ignored when calculating the observed interval span. @AtomicGlance [#2212](https://github.com/NeurodataWithoutBorders/pynwb/issues/2212)
+- Fixed `FeatureExtraction.times` and `Clustering.peak_over_rms` being copied into a Python list on construction, which read the backing dataset one element at a time. @cboulay [#2253](https://github.com/NeurodataWithoutBorders/pynwb/pull/2253)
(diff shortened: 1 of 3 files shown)


AtomicGlance:
I updated this branch onto the current PyNWB dev branch and resolved the release changelog conflict. All 26 tests in 	ests/unit/test_epoch.py pass locally, including NaN starts/stops and ongoing-interval cases. The diff remains limited to TimeIntervals behavior, tests, and the changelog. It is ready for review when convenient.

AtomicGlance:
I rechecked the refreshed branch against the local source tree: the NaN/ongoing interval cases pass (5 passed), including all-NaN and later-start boundaries. No new code changes were needed; the branch is current and ready for review.

PR #2256, opened 2026-09-01
DO NOT MERGE: test HDF5 PR 6654 against the Windows ros3 deadlock

rly:
**Do not merge.** This is a throwaway branch for running one experiment: does [HDFGroup/hdf5#6654](https://github.com/HDFGroup/hdf5/pull/6654) fix the Windows ROS3 shutdown deadlock from #2228 / [HDFGroup/hdf5#6560](https://github.com/HDFGroup/hdf5/issues/6560)?

## Result

Yes. Run: https://github.com/NeurodataWithoutBorders/pynwb/actions/runs/33473222181

| | `stock` (conda-forge) | `unpatched` (built here) | `patched` (built here + PR) |
|---|---|---|---|
| process-exit check | `HUNG` at 180s | `HUNG` at 180s | `EXITED` after 1.1s |
| `scripts/test_ros3.py` watchdog | fired | fired | did not fire |
| pynwb ros3 tests | Ran 4 tests, OK | Ran 4 tests, OK | Ran 4 tests, OK |

All three legs report `h5py 3.16.0 / HDF5 2.1.0` with `ros3` registered.

## How it tests the PR

`.github/workflows/test_ros3_windows_hdf5_pr6654.yml` adds a `windows-latest` job with three matrix legs:

- `stock`: the unmodified conda-forge HDF5 that `environment-ros3.yml` pulls in.
- `unpatched`: HDF5 built from the `hdf5_2.1.0` tag, with the resulting `hdf5.dll` copied over the conda-forge one.
- `patched`: the same build with the PR's `src/` and `config/` changes applied.

`unpatched` and `patched` differ only in the PR, which isolates the PR as the cause of the difference between them. `stock` matching `unpatched` additionally shows that the build settings used here reproduce the deadlock that users hit with released builds, so the ways this build differs from conda-forge's (szip, libcurl) are not affecting the outcome.

The PR's diff applies cleanly to the `hdf5_2.1.0` tag (only small line offsets in `src/H5FDros3_s3comms.c`), and conda-forge `h5py 3.16.0` on win-64 depends on `hdf5 >=2.1.0,<3.0a0`. Building the same tag means the ABI matches the h5py binary exactly, so h5py does not need to be rebuilt from source. The PR's `release_docs/CHANGELOG.md` hunk is dropped because it does not apply to the release tag.

Each leg runs two checks:

1. `scripts/check_ros3_exit.py` spawns a child that opens `ros3test.nwb` over the ROS3 driver, prints `ROS3_READ_OK`, and returns from `main`. The parent reports whether the child exits within 180s. This isolates process teardown from everything else.
2. `python scripts/test_ros3.py --ros3`, the real suite, checking whether the existing shutdown watchdog fires.

The `patched` leg fails if either check still shows the hang. The `stock` and `unpatched` legs only report.

One wrinkle worth recording: Git Bash cannot pass through the `ProgramFiles(x86)` environment variable, so CMake's Visual Studio generator cannot find vswhere and reports that no Visual Studio instance exists. The job uses `ilammy/msvc-dev-cmd` plus the Ninja generator instead.

## Does the PR address the concerns in HDFGroup/hdf5#6560?

Yes for the deadlock, by design as a stopgap; the maintainer says so explicitly in the changelog entry.

On Windows the PR stops registering `H5FD__s3comms_term_func` with `atexit` and instead calls it from `H5FD__s3comms_term`, which the ROS3 VFD's `terminate` callback (`H5FD__ros3_term`) invokes. Inside that function, `aws_s3_library_clean_up()` is guarded by `RtlDllShutdownInProgress()`. So:

- Process exits without `H5close()`: HDF5's own `atexit` handler runs during `DLL_PROCESS_DETACH`, `RtlDllShutdownInProgress()` returns true, and the `aws_thread_join_all_managed()` call that blocks on the loader lock is skipped. No deadlock. Skipping is safe here because `ExitProcess` has already terminated the AWS worker threads, which is why the join never returns in the first place.
- Application calls `H5close()` first: the guard returns false and the aws-c-s3 cleanup runs normally.

Two caveats worth naming:

- Without an explicit `H5close()`, aws-c-s3 resources are simply leaked at exit. The changelog says so. For pynwb that is fine, but it means the "correct" path requires downstream callers (h5py, ultimately) to call `H5close()`, which nothing does today.
- If `CHECK_LIBRARY_EXISTS_CONCAT` fails to find `RtlDllShutdownInProgress` in `ntdll`, the `#ifndef H5_HAVE_WINDOWS` / `#elif defined(H5_HAVE_RTLDLLSHUTDOWNINPROGRESS)` chain in `H5FD__s3comms_term_func` leaves no cleanup call at all on Windows. Still no deadlock, just an unconditional leak.

🤖 Generated with [Claude Code](https://claude.com/claude-code) and reviewed and modified by me.


rly:
## Three-leg run, with a self-built unpatched control

Run: https://github.com/NeurodataWithoutBorders/pynwb/actions/runs/33473222181

`unpatched` builds the same `hdf5_2.1.0` tag with the same CMake settings and toolchain as `patched`, without the PR. The PR is the only variable between those two legs.

| | `stock` (conda-forge) | `unpatched` (built here) | `patched` (built here + PR) |
|---|---|---|---|
| process-exit check | `HUNG` at 180s | `HUNG` at 180s | `EXITED` after 1.1s |
| `scripts/test_ros3.py` watchdog | fired | fired | did not fire |
| pynwb ros3 tests | Ran 4 tests, OK | Ran 4 tests, OK | Ran 4 tests, OK |

All three report `h5py 3.16.0 / HDF5 2.1.0` with `ros3` registered.

`stock` and `unpatched` agreeing also confirms that the build settings used here reproduce the deadlock users see in released builds, so the szip and libcurl differences between this build and conda-forge's are not doing anything.

Earlier two-leg run, for reference: https://github.com/NeurodataWithoutBorders/pynwb/actions/runs/33466245039

